PR #43, opened 2024-04-13
Update readme.md

pedrotski:
CSGO doesn't exist any more so updated it to say Counter Strike 2. Just a slight change.


QuintenQVD0:
but the egg is still the csgo server? cs2 is not here as a stock egg.

lancepioch:
Thanks but we're gonna revamp the readme!

PR #96, opened 2024-04-17
First removal of unused variables

notAreYouScared:
Working on Issue #2, Removing unused variables.

### DIFF
--- a/app/Http/Controllers/Admin/Nodes/NodeController.php
+++ b/app/Http/Controllers/Admin/Nodes/NodeController.php
@@ -13,7 +13,7 @@ class NodeController extends Controller
     /**
      * Returns a listing of nodes on the system.
      */
-    public function index(Request $request): View
+    public function index(): View
     {
         $nodes = QueryBuilder::for(
             Node::query()->withCount('servers')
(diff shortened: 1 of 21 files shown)


Boy132:
Aren't some of the request variables you removed needed for things like permission checks? Like `GetEggRequest` for example?

notAreYouScared:
> Aren't some of the request variables you removed needed for things like permission checks? Like `GetEggRequest` for example?

image


Where is the variable `$request` used in these functions...
```php
    public function index(GetEggsRequest $request): array
    {
        return $this->fractal->collection(Egg::all())
            ->transformWith($this->getTransformer(EggTransformer::class))
            ->toArray();
    }

    /**
     * Return a single egg that exists
     */
    public function view(GetEggRequest $request, Egg $egg): array
    {
        return $this->fractal->item($egg)
            ->transformWith($this->getTransformer(EggTransformer::class))
            ->toArray();
    }
}
```

Boy132:
It's not used directly. But it's used internally to check if the key has the required permissions?
https://github.com/pelican-dev/panel/blob/3.x/app/Http/Requests/Api/Application/Eggs/GetEggRequest.php#L10-L12
https://github.com/pelican-dev/panel/blob/3.x/app/Http/Requests/Api/Application/ApplicationApiRequest.php#L34

PR #101, opened 2024-04-18
url function around static resource links

Miniontoby:
Wrapped the static resource links in the url function to make it sync to APP_URL

This would allow later on to support running in an subfolder

### DIFF
--- a/app/Extensions/Themes/Theme.php
+++ b/app/Extensions/Themes/Theme.php
@@ -16,6 +16,6 @@ class Theme
 
     protected function getUrl($path): string
     {
-        return '/themes/panel/' . ltrim($path, '/');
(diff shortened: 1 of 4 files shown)


Miniontoby:

iamkubi:
Hmm, looking at the CLA failure.  It doesn't look like it added your signature or rechecked after your comment, so something probably broke.

Miniontoby:
recheck

Miniontoby:
Maybe it will trigger now?

iamkubi:
I think I likely broke the CLA workflow for this PR by messing with the signatures in the middle.  Safe to ignore if it continues to fail for this PR.

Miniontoby:
The check just needs to be retriggered/retried/reexecuted. I know that that is a thing with github actions

PR #102, opened 2024-04-18
url function around static resource links (FILAMENT)

Miniontoby:
Wrapped the static resource links in the url function to make it sync to APP_URL

This would allow later on to support running in an subfolder

*Filament version, with fix for livewire*

### DIFF
--- a/app/Extensions/Themes/Theme.php
+++ b/app/Extensions/Themes/Theme.php
@@ -16,6 +16,6 @@ class Theme
 
     protected function getUrl($path): string
     {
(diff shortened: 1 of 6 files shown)


Miniontoby:
Oke, yeah someone needs to lint the branch... but I don't want to do it in here...

lancepioch:
Thank you for your contribution. I understand where you're coming from, but in short there's a more than this in the list of reasons why this app simply won't work in a sub directory. We won't be able to support that in the mean time.

PR #108, opened 2024-04-19
Fix missed adjustments on http parameters

NeonSpectrum:
Update some daemon http outdated parameters

### DIFF
--- a/app/Repositories/Daemon/DaemonFileRepository.php
+++ b/app/Repositories/Daemon/DaemonFileRepository.php
@@ -56,9 +56,7 @@ class DaemonFileRepository extends DaemonRepository
             return $this->getHttpClient()
                 ->withQueryParameters(['file' => $path])
                 ->withBody($content)
-                ->post(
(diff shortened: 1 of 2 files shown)


app/Repositories/Daemon/DaemonFileRepository.php
```diff
@@ -55,13 +53,12 @@ public function putContent(string $path, string $content)
...
-                    'query' => ['file' => $path],
-                    'body' => $content,
-                ]
-            );
+            return $this->getHttpClient()
+                ->withQueryParameters(['file' => $path])
+                ->withBody($content)
+                ->post(
+                    sprintf('/api/servers/%s/files/write', $this->server->uuid)
+                );
```
lancepioch, line 61:
```suggestion
                ->post(sprintf('/api/servers/%s/files/write', $this->server->uuid));
```

iamkubi:
Seems like this probably broke the file browser.

PR #139, opened 2024-04-22
Support SQLite

### LINKED ISSUE #134: Support SQLite

Boy132:
Closes #134


Boy132:
Tests for sqlite fail because of the new `tags` column for eggs. The seeding fails because tags haven't been implemented yet.
(It doesn't fail for mysql because the mysql dump doesn't contain the tags column)

PR #147, opened 2024-04-23
Added Install command to make installing the panel easier

Poseidon281:
This command is a merge of the 3 environment commands with additional features to make setting up the panel a lot easier. 

#71


lancepioch:
I am certainly not opposed to this concept. However I will point out that I want to add a visual GUI based setup flow soon.

Poseidon281:
> I am certainly not opposed to this concept. However I will point out that I want to add a visual GUI based setup flow soon.

Oke i will put this on hold then and make a command when that is finished based that GUI

lancepioch:
Okay thank you. Let's have a call about this in the future when you're ready!

Issue #70, opened 2024-04-14
User Suspension


lancepioch:
### Discussed in https://github.com/pelican-dev/panel/discussions/12

<div type='discussions-op-text'>

<sup>Originally posted by **martinbndr** April 12, 2024</sup>
Being able to suspend users in pelican panel.
Users suspended should be completly logged out and won´t be able to login anymore. Specific error message will show that.
Please also add api support for that.</div>

Googletod12355:
I can agree, right now there is no way to stop certain users from remaking their account

Poseidon281:
Working on it.

PR #150, opened 2024-04-24
User Suspension

Poseidon281:
#12 
#70 

Added User-Suspension in Filament Admin.

### DIFF
--- a/app/Filament/Resources/UserResource/Pages/EditUser.php
+++ b/app/Filament/Resources/UserResource/Pages/EditUser.php
@@ -71,7 +71,6 @@ class EditUser extends EditRecord
 
             Actions\Action::make('toggleSuspend')
                 ->hidden(fn (User $user) => $user->is_suspended || $user->root_admin)
-                ->tooltip('testt')
                 ->label('Disable User')
                 ->color('warning')
                 ->action(function (User $user) {
--- a/app/Filament/Resources/UserResource/Pages/EditUser.php
+++ b/app/Filament/Resources/UserResource/Pages/EditUser.php
@@ -3,6 +3,7 @@
 namespace App\Filament\Resources\UserResource\Pages;
 
 use App\Filament\Resources\UserResource;
+use App\Services\Servers\SuspensionService;
 use Filament\Actions;
 use Filament\Resources\Pages\EditRecord;
 use App\Models\User;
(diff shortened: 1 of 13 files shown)


Poseidon281:
> Not sure I would've used the word banned, I think `is_suspended`, `suspended`, `suspended_message` would be a better fit
> 
> This should also suspend servers to prevent people from making a sub-user to access the serves, if any, allowing them to bypass the suspend.

I can change the ban to suspended, but i dont have any free servers to run a node on at the moment so i wont be able to test the server suspend. Someone else will have to add that feature.

Poseidon281:
@notAreYouScared 

Like i said i cannot test the server suspend, but this code should just work if i am right. If you or someone could test it that would be great.

notAreYouScared:
@Poseidon281 

So there is a built-in suspend function to suspend the servers, [SuspensionService.php](https://github.com/pelican-dev/panel/blob/main/app/Services/Servers/SuspensionService.php)

It should trigger and be supplied the associated server id's to properly suspend them in both the panel and wings. Setting the column value does nothing until wings pulls the new server status.

notAreYouScared:
Should be all set.

app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -62,18 +54,45 @@ public function form(Form $form): Form
...
             ]);
     }
     protected function getHeaderActions(): array
     {
         return [
             Actions\DeleteAction::make(),
+
+            Actions\Action::make('toggleSuspend')
+                ->hidden(fn (User $user) => $user->is_suspended || $user->root_admin)
+                ->tooltip('testt')
```
Boy132, line 74: Leftover from testing?
```suggestion
```

app/Http/Middleware/CheckBanned.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Filament\Notifications\Notification;
+
+class CheckBanned
```
Boy132, line 8: This is still named "banned"

```suggestion
class CheckUserSuspension
```

notAreYouScared: We've been talking about not suspending the user account, and kicking them out of the panel, but just suspending the servers. So it will probably change

Poseidon281: Do you want to finish it Charles or is there something you would like me to do for this PR?

notAreYouScared: Let it be for now. We're going to have more discussion about it, I'll finish it from here.

lancepioch:
We've added the mass suspend servers button to the User resource.

As for actually suspending users, you can currently change their password or delete them.

PR #153, opened 2024-04-25
Add api for server transfers

### LINKED ISSUE #129: Add server transfer API

Boy132:
Closes #129

- POST /api/application/{server}/transfer _(starts a server transfer)_
- POST /api/application/{server}/transfer/cancel _(cancels a server transfer)_


iamkubi:
Can we add a GET endpoint to report the status of the transfer?

Boy132:
> Can we add a GET endpoint to report the status of the transfer?

If there is a wings endpoint for that, sure.
But I believe the current status is only available via websocket? (haven't double checked yet)

iamkubi:
There's no wings endpoint for it, the Panel ultimately handles the status for the transfer since it spans nodes.  The application API's server status variable will report if the state is transferring.  If the state is `none` I suppose we'd have to see if we can derive a status from the database?  I haven't looked at how transfers are recorded and purged from the database so I don't really know what's possible here.

PR #178, opened 2024-05-06
Auto pint

Poseidon281:
I made changes to lint.yml so on each commit it now automatically runs pint and makes a commit named Pint Fixes if needed


Poseidon281:
recheck

Poseidon281:
My changes do work, only there is one little thing wrong. The job fails because he can't find the branch in this repository and I can't manage to get him to search for the fork repository, but that does not affect the job. As you can see from my tests, he does run pint and then makes a commit. With my last commit I adjusted it so that it is no longer executed when a Pull Request is made, but only when a commit is made, because as I explained at Pull Requests it fails because it can't find the branch.

lancepioch:
You can use a pre commit git hook to run vendor/bin/pint

PR #189, opened 2024-05-08
Dashboard button change & composer update

Poseidon281:
This changed the overview in the Dashboard to buttons with redirects:

image


lang/en/commands.php
notAreYouScared, line 1: Please use single quotes to match the other /lang/en files.

Poseidon281: This is already fixed by pint in a other commit and i have not changed anything in this file in this PR

composer.lock
Poseidon281, line 1: I updated the composer. I have tested it and i dont think this caused any errors. I dont know why it should but it does not

notAreYouScared: I have done this locally, and all seems well.... So I'm sure its fine <3

resources/views/filament/pages/dashboard.blade.php
notAreYouScared, line 1: Makes no sense to have buttons at the top and have the nav bar on the left. 
They should just be, placeholders? or labels.

lancepioch:
We already have the same links on the side, these are just supposed to be a high level overview. Perhaps we could show the badges on the navigation sidebar?

PR #166, opened 2024-04-30
feat: Replace Yarn with Pnpm

alexevladgabriel:

PR #190, opened 2024-05-08
feat: Roles

Poseidon281:
With the help of [Filament Shield](https://filamentphp.com/plugins/bezhansalleh-shield) i am making a Roles System for the panel. Suggestions are always welcome 

#37 
#157


KITOXIS:
Will the roles be creatable or will we have to use what you made?

Poseidon281:
> Will the roles be creatable or will we have to use what you made? And also, if we will be able to create the roles. Will we be able to set the color and the "badge" icon? (Like font awesome)

I still have to discuss this with the developers so I am not sure if this is going to be added, but if this is going to be added then yes the roles will be creatable. To answer your other question Currently the plugin I am using to create the roles does not support that, but I will see what I can do to add that myself after I have discussed it.

yarn.lock
Poseidon281, line 1: I dont know why but i was getting an yarn error so i did yarn upgrade and it resolved itself that is why i did yarn upgrade despite it being unnecessary because of #166

lancepioch:
Roles and permissions are going to be revamped separately!

Issue #197, opened 2024-05-12
Admin View a Server


antho-444:
### Current Behavior

The current behavior that I'm experiencing is whenever you have a server created, you are an Administrator, and click onto the button on the right which is next to Activity, it send you to a 404 | Not Found page.

### Expected Behavior

The expected behavior is the button next to activity was supposed to send the user to the Admin Server Editing Page.

### Steps to Reproduce

To reproduce this bug, You'll need to be a Administrator and have any servers created. if you click onto the button on the right of "Activity", it will give a 404 | Not Found Page. 

### Panel Version

1.0.0-beta1

### Wings Version

1.0.0-beta1

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

notAreYouScared:
This is due to the new admin side using the uuid in the url instead of the servers id, like the old admin side did.

Julian-Louis:
I created a PR to fix this issue: https://github.com/pelican-dev/panel/pull/198

PR #198, opened 2024-05-12
Fix server external link to admin

Julian-Louis:
Hello,
This PR fix the following issue: https://github.com/pelican-dev/panel/issues/197.


Julian-Louis:

notAreYouScared:
Thanks for the fix, but we should change it on the new admin panel side, as editing node/eggs uses the ID and not the UUID.
makes no sense to have one use UUID and the rest use ID

lancepioch:
Thanks but we'll be fixing the url itself 😅

PR #232, opened 2024-05-15
Update: Navigation sidebar

Poseidon281:
I made changes to the Navigation sidebar as asked [here](https://github.com/pelican-dev/panel/pull/189#issuecomment-2109288814). Would you like me to remove the overview on the Dashboard?

If you want me to alter/add/undo anything just let me know

### DIFF
--- a/app/Providers/Filament/AdminPanelProvider.php
+++ b/app/Providers/Filament/AdminPanelProvider.php
@@ -36,6 +36,7 @@ class AdminPanelProvider extends PanelProvider
             ->default()
             ->id('admin')
             ->path('admin')
+            ->topNavigation(true)
             ->login()
             ->homeUrl('/')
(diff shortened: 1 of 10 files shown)


app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -36,6 +36,7 @@ public function panel(Panel $panel): Panel
             ->default()
             ->id('admin')
             ->path('admin')
+            ->topNavigation(true)
```
Poseidon281, line 39: Because of the badges not all of the labels of the navigation sidebar where visible anymore. So i added this line which changes the sidebar to a topbar: 
image

Should you not want this simply set it to false and then it is disables again. This can be added to a settings page which i can make if you guys want to.

PR #239, opened 2024-05-16
Add CPU limit to node

### LINKED ISSUE #233: Add Node CPU resource allocations
Nodes should be able to define CPU resource allocations that are respected by the deployment service.  This would work the same as disk/memory where you can set a maximum amount of CPU that should be assignable to the node.  After this automated deployments using the deployment service would consider the node to be full if adding the new server would drive its allocated CPU time above the node's defined limit.

As best I can tell no work is required on the wings side for this.  Node models would need to be updated, total assigned CPU allocations per node would need to be calculated, and that allocated usage would be used by the deployment service to determine if a node can deploy a server.

Boy132:
Closes #233


app/Http/Requests/Api/Application/Nodes/GetDeployableNodesRequest.php
```diff
@@ -10,6 +10,7 @@ public function rules(): array
             'page' => 'integer',
             'memory' => 'required|integer|min:0',
             'disk' => 'required|integer|min:0',
+            'cpu' => 'required|integer|min:0',
```
lancepioch, line 13: Probably should be optional so that all previous api requests don't break.

lancepioch (changes requested):
Awesome thanks, super close, let's test this and also make sure that the new param is optional (defaulted) in the api endpoints.

PR #244, opened 2024-05-16
fix: Creating User

notAreYouScared:
Removes the required password field for making new users, triggers email send for user to make their own password.


notAreYouScared:
Im sure there is a better way to do this as it bypasses the validation on the fields... It just sends the data to the user creation service, resulting in a 500 if the email isn't and email and if the username is blank

PR #245, opened 2024-05-17
Small start to the client side

Poseidon281:
Because of the high demand on the Discord i started working on the client side of the panel with Filament. Any suggestions are always welcome

If you want me to alter/add/undo anything just let me know.


Boy132:
I would say let's focus on the admin area first. There is no reason to start on the client area while the admin area isn't even finished.

Poseidon281:
> I would say let's focus on the admin area first. There is no reason to start on the client area while the admin area isn't even finished.

I am just doing working on the start further then that i am not going for now. Until the admin side is complete

Poseidon281:
> I would say let's focus on the admin area first. There is no reason to start on the client area while the admin area isn't even finished.

Multiple people can do multiple things so while i am waiting on for an answer on some questions for some idea's i had for the admin side. I made this.

crocofied:
Very nice, thank you! Hopefully it will look as nice as the admin panel

lancepioch:
Don't create a PR for this yet, thanks.

PR #246, opened 2024-05-17
Added footer & small improvments 

Poseidon281:


resources/views/filament/Footer.blade.php
```diff
@@ -0,0 +1,8 @@
+<footer class="fixed bottom-0 left-0 z-20 w-full p-4 bg-white border-t border-gray-200 shadow md:flex md:items-center md:justify-between md:p-6 dark:bg-gray-800 dark:border-gray-600">
+    <div class="flex items-center">
+        <img src="{{ asset('pelican.ico') }}" alt="Pelican Logo" class="w-6 h-6" style="margin-right: 8px;">
+        <span class="text-sm text-gray-500 sm:text-center dark:text-gray-400">© 2024 - <?php echo date('Y'); ?>
+            <a href="https://github.com/pelican-dev/panel" class="hover:underline">Pelican</a>
```
lancepioch, line 5:
```suggestion
            <a href="https://pelican.dev" class="hover:underline">Pelican</a>
```

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -51,12 +51,21 @@ public function panel(Panel $panel): Panel
...
+                    ->icon('tabler-arrow-back')
+                    ->sort(12)
+                    ->visible(fn (): bool => auth()->user()->root_admin),
+            ])
             ->discoverResources(in: app_path('Filament/Resources'), for: 'App\\Filament\\Resources')
             ->discoverPages(in: app_path('Filament/Pages'), for: 'App\\Filament\\Pages')
             ->discoverClusters(in: app_path('Filament/Clusters'), for: 'App\\Filament\\Clusters')
-            ->pages([
-                // Pages\Dashboard::class,
-            ])
```
lancepioch, line 59: Keep

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -51,12 +51,21 @@ public function panel(Panel $panel): Panel
...
+                'panels::sidebar.footer',
+                fn () => view('filament.Footer'),
+            )
+            ->navigationItems([
+                NavigationItem::make('client')
+                    ->label('Return to client')
+                    ->url('/') // TODO once the Filament client side is done this can be changed to /client
+                    ->icon('tabler-arrow-back')
+                    ->sort(12)
+                    ->visible(fn (): bool => auth()->user()->root_admin),
```
lancepioch, line 64:
```suggestion
```

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -51,12 +51,21 @@ public function panel(Panel $panel): Panel
...
                 'blurple' => Color::hex('#5865F2'),
             ])
+            ->renderHook(
+                'panels::sidebar.footer',
+                fn () => view('filament.Footer'),
+            )
+            ->navigationItems([
+                NavigationItem::make('client')
+                    ->label('Return to client')
+                    ->url('/') // TODO once the Filament client side is done this can be changed to /client
```
lancepioch, line 61:
```suggestion
                    ->url(route('route.name.here'))
```

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -51,12 +51,21 @@ public function panel(Panel $panel): Panel
                 'warning' => Color::Amber,
                 'blurple' => Color::hex('#5865F2'),
             ])
+            ->renderHook(
+                'panels::sidebar.footer',
+                fn () => view('filament.Footer'),
+            )
```
lancepioch, line 57:
```suggestion
            ->renderHook('panels::sidebar.footer', fn () => view('filament.Footer'))
```

app/Filament/Resources/ServerResource.php
```diff
@@ -10,7 +10,7 @@ class ServerResource extends Resource
 {
     protected static ?string $model = Server::class;
 
-    protected static ?string $navigationIcon = 'tabler-brand-docker';
+    protected static ?string $navigationIcon = 'tabler-server-2';
```
lancepioch, line 13:
```suggestion
    protected static ?string $navigationIcon = 'tabler-brand-docker';
```

app/Filament/Resources/NodeResource.php
```diff
@@ -11,7 +11,7 @@ class NodeResource extends Resource
 {
     protected static ?string $model = Node::class;
 
-    protected static ?string $navigationIcon = 'tabler-server-2';
+    protected static ?string $navigationIcon = 'tabler-server';
```
lancepioch, line 14:
```suggestion
    protected static ?string $navigationIcon = 'tabler-server-2';
```

resources/views/filament/Footer.blade.php
```diff
@@ -0,0 +1,8 @@
+<footer class="fixed bottom-0 left-0 z-20 w-full p-4 bg-white border-t border-gray-200 shadow md:flex md:items-center md:justify-between md:p-6 dark:bg-gray-800 dark:border-gray-600">
+    <div class="flex items-center">
+        <img src="{{ asset('pelican.ico') }}" alt="Pelican Logo" class="w-6 h-6" style="margin-right: 8px;">
+        <span class="text-sm text-gray-500 sm:text-center dark:text-gray-400">© 2024 - <?php echo date('Y'); ?>
```
lancepioch, line 4:
```suggestion
        <span class="text-sm text-gray-500 sm:text-center dark:text-gray-400">© 2024
```

lancepioch (changes requested):
Awesome, thanks. I'll take a look later!

app/Providers/Filament/AdminPanelProvider.php
Poseidon281: I tried this at first and it did not work for some reason i get this error:

image

So i did this. It works just as well

Poseidon281:
> Awesome, thanks. I'll take a look later!
@lancepioch 

Right now i have the Return to Client on the nav bar like this:

image

If you like that more i can also add it to the user menu like this:

image

Which ever one you like more

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -51,17 +51,25 @@ public function panel(Panel $panel): Panel
                 'warning' => Color::Amber,
                 'blurple' => Color::hex('#5865F2'),
             ])
+            ->renderHook('panels::sidebar.footer', fn () => view('filament.Footer'))
+            ->navigationItems([
+                NavigationItem::make('client')
+                    ->label('Return to client')
```
notAreYouScared, line 57: Can we name this `Exit Admin`
```suggestion
                    ->label('Exit Admin')
```

resources/views/filament/Footer.blade.php
```diff
@@ -0,0 +1,8 @@
+<footer class="fixed bottom-0 left-0 z-20 w-full p-4 bg-white border-t border-gray-200 shadow md:flex md:items-center md:justify-between md:p-6 dark:bg-gray-800 dark:border-gray-600">
```
notAreYouScared, line 1: Rename the file to `footer.blade.php`. No need for it to be capital

gOOvER:
2nd one is nice. Return to Client on top

lancepioch:
So the footer looks nice but also awkward because it's abruptly cut off.
Perhaps we can get it showing across the whole bottom?
Also the footer doesn't show up at all if the top navigation is enabled.

Poseidon281:
> So the footer looks nice but also awkward because it's abruptly cut off. Perhaps we can get it showing across the whole bottom? Also the footer doesn't show up at all if the top navigation is enabled.

Yes that is because I set it to the footer of het navbar, but with my last commit I changed it to look like this:

image

notAreYouScared:
> image


We can use this one.

lancepioch:
Very last screenshot looks good, think we can get that one in

Poseidon281:
> Very last screenshot looks good, think we can get that one in

Oké well that is the one I have for this PR right now so I think the PR can be merged then.

resources/views/filament/footer.blade.php
```diff
@@ -0,0 +1,32 @@
...
+
+footer img {
+    margin-right: 8px;
+}
+
+footer span {
+    padding: 0;
+    margin: 0;
+}
+</style>
```
lancepioch, line 32: These should just be tailwind classes above

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -76,5 +84,11 @@ public function panel(Panel $panel): Panel
             ->authMiddleware([
                 Authenticate::class,
             ]);
+
+        if (config('panel.filament.top-navigation') === false) {
+            $panel->renderHook('panels::sidebar.footer', fn () => view('filament.footer'));
+        } else {
+            //$panel->renderHook('panels::topbar.start', fn () => view('filament.footer')); // TODO find a fix for the footer if topbar is enabled
+        }
```
lancepioch, line 92: I don't actually see the footer before or after my changes.

lancepioch (changes requested):
Thanks for the work, two last things and we're gtg

app/Providers/Filament/AdminPanelProvider.php
Poseidon281: Thats weird. I do see it as you can see on the screenshot? Do you have topbar on? Because then it does not work only with sidebar for now

resources/views/filament/footer.blade.php
Poseidon281: The styles defined in the <style> block should remain because changing them to Tailwind classes alone does not work then the footer is over Pelican for some reason. I have not got it working in any other way except this one.

app/Providers/Filament/AdminPanelProvider.php
Poseidon281: Fixed it. For some reason this bit of code did not work anymore, but it did work when i last tested it and send my screenshots so no idea why it does not now.

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -50,6 +51,14 @@ public function panel(Panel $panel): Panel
                 'warning' => Color::Amber,
                 'blurple' => Color::hex('#5865F2'),
             ])
+            ->renderHook('panels::sidebar.footer', fn () => view('filament.footer')) // TODO make the footer also work with topNavigation
```
lancepioch, line 54:
```suggestion
            ->renderHook(PanelsRenderHook::FOOTER, fn () => view('filament.footer')) // TODO make the footer also work with topNavigation
```

Poseidon281: When you do this with only footer and not sidebar footer then you have this:

image

I have tested almost all of the renderhooks and this is the only one that comes out nice:

image

lancepioch:
For better or worse I think it'll be best for us to absolutely stick with TailwindCSS instead of using raw CSS in here. Let's try to tackle that.

Poseidon281:
> For better or worse I think it'll be best for us to absolutely stick with TailwindCSS instead of using raw CSS in here. Let's try to tackle that.

Oke i cant get it to work with Tailwind so if one of the team members can try that. That would be great

lancepioch:
Let's table this for now until we can figure out Tailwind. It will probably be a lot easier after #353

PR #256, opened 2024-05-18
Feature: egg meta.update_url field implementation

a-sync:
A frontend/browser only proof of concept use case for the meta.update_url egg property.
This is how it works:  
 1. if the update_url property is available, a "Download from update URL" button appears next to the "Update egg" button
 1. activating the "Download from update URL" button downloads the referenced egg JSON in the browser context, and pre-sets it in the file upload form with a name originated from the exported_at property
 1. activating the "Update egg" button uploads the pre-set file (unless something else was selected)

The only catch is that the host serving the egg JSON must have the proper CORS headers configured for the clients to be able to download the file.  
Tested it with success with an egg hosted in a github repo served through raw.githubusercontent.com, jsdelivr.com and statically.io.  
update_url poc

### DIFF
--- a/resources/views/admin/eggs/view.blade.php
+++ b/resources/views/admin/eggs/view.blade.php
@@ -43,6 +43,9 @@
                         <div class="col-xs-4">
                             {!! csrf_field() !!}
                             <button type="submit" name="_method" value="PUT" class="btn btn-sm btn-danger pull-right">Update Egg</button>
+                            @if($egg->update_url)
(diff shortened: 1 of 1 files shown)


a-sync:

Poseidon281:
The Pterodactyl admin side is no longer used in Pelican the Filament admin side is what the team is currently working on and in time I think this side will be removed and fully replaced by the Filament admin side.

PR #264, opened 2024-05-22
Fixed mumble host replace value to support both IPv4 and IPv6 allocations.

JordanPlayz158:
Commit message is descriptive but with last egg, it was hard coded to `0.0.0.0` that only seems to work with IPv4 allocations but when you leave it blank, it supports both IPv4 and IPv6 allocations.

Edit: Should be a way to squash on merge


JordanPlayz158:

Software-Noob (approved):

PR #276, opened 2024-05-26
feat: Sorting and grouping

### LINKED ISSUE #249: Add sorting and/or grouping to the front end
### Discussed in https://github.com/pelican-dev/panel/discussions/26

<sup>Originally posted by **CodeByCam** April 12, 2024</sup>
G'day, 

Just quickly want to say as a Pterodactyl user for quite some time, I'm very excited about Pelican to see more frequent development! 
I'd like to request something that was long sought-after in Pterodactyl, which is easily sorting & grouping servers on the main page. 

Sorting of servers could be as simple as drag and drop and this alone would tremendously improve the user experience when one has many servers. It would also be great to have grouping too, for example you could have two groups: "Production" & "Development" which would house different servers inside yet again improving the user experience. Grouping servers could be similar to sorting servers, but you drag and drop a server onto another to create the group (think creating groups for apps on a smartphone). This is just a huge quality of life feature that I feel has been missing from Pterodactyl from the get-go.

Here is a quick video by @smashyalts who created something similar in Pterodactyl as an open source addon.

https://github.com/pelican-dev/panel/assets/41496198/82608c4f-e501-477c-a6fb-76208d7dde0a

Cheers for taking the time to read this, and I hope it will be taken under consideration! </div>

Poseidon281:
This PR closes #249 

This PR uses code from [here](https://github.com/smashyalts/PterodactylModifications/blob/main/DashboardContainer.tsx)

### TODO List 
| Task| Status | Number |
|--------|--------|--------|
| Change default group name to All Servers | Done | 1 |


bootstrap/app.php
```diff
@@ -9,7 +9,7 @@
         \Prologue\Alerts\AlertsServiceProvider::class,
     ])
     ->withRouting(
-        web: __DIR__.'/../routes/web.php',
+        web: __DIR__.'/../routes/base.php',
```
Poseidon281, line 12: This was necessary because otherwise my panel would not load. web.php does not exist then I get as error

Poseidon281:
End result:

image

resources/scripts/components/dashboard/DashboardContainer.tsx
```diff
@@ -51,36 +64,187 @@ export default () => {
         if (!error) clearFlashes('dashboard');
     }, [error]);
 
+    useEffect(() => {
+        const savedGroups = localStorage.getItem(`groups:${uuid}`);
+        if (savedGroups) {
+            setGroups(JSON.parse(savedGroups));
+        }
+    }, [uuid]);
```
DjordyKoert, line 72: This will cause unnecessary re-renders. https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

Poseidon281: I am not really good at Typescript. 

Did you mean [this](https://github.com/pelican-dev/panel/pull/276/commits/d946b207eb214d3ab484e99446bb81689eadc270)? If not could you make a suggestion.

lancepioch:
Thanks, we're gonna table this because of #353 - we'll just have Filament grouping and sorting :D

PR #284, opened 2024-05-28
Replace encrypt/ decrypt with `encrypted` casting 

### LINKED ISSUE #4: Replace encrypt and decrypt calls with encrypted casting
We have calls to `encrypt` and `decrypt` all over the place. We should be able to replace almost all of them with an encrypted cast: https://laravel.com/docs/11.x/eloquent-mutators#encrypted-casting

Boy132:
Closes #4


Boy132:
I don't know why but for old data it will return the decrypted value of an `encrypted` attribute in a weird format:
grafik

For new data this is not the case:
grafik

notAreYouScared:
Beta 3 will already have a breaking change, We could tell users to rotate the daemon key with the reset button. That should fix it? Or does this happen for everything? like passwords?

Boy132:
I believe it happens for everything but haven't double checked yet.

lancepioch (approved):

PR #295, opened 2024-05-29
Make deploy.locations optional [backwards compat] as it has been removed

ItzExotical:
The locations functionality was removed from the panel, and will be replaced with tags shortly.
The API has to be changed so that it does not force validation of the location data.


ItzExotical:

PR #313, opened 2024-06-03
Docker

lancepioch:


.github/docker/entrypoint.sh
```diff
@@ -1,77 +1,22 @@
...
-     echo -e "APP_KEY=$APP_KEY" > /app/var/.env
-  else
-    echo -e "APP_KEY exists in environment, using that."
-    echo -e "APP_KEY=$APP_KEY" > /app/var/.env
-  fi
+cd /var/www/html
 
-  ln -s /app/var/.env /app/
-fi
+#chmod -R 775 storage/* bootstrap/cache/
```
parkervcp, line 7: remove this line

.github/docker/entrypoint.sh
```diff
@@ -1,77 +1,22 @@
...
+#chown -R caddy:caddy .
```
parkervcp, line 8: move to the `exec "$@"` and change `caddy:caddy` to `www-data:`

parkervcp (changes requested):
Need to update the `entrypoint.sh`. 

We should also have a `compose.yml` for easy automation

engels74:
I'm not maintainer at all, but I still have a question :)

Previously, Pterodactyl Panel had a [quite long list](https://github.com/pterodactyl/panel/blob/develop/.env.example) of variables in its `.env.example` file. In this PR, the `.env.docker` file only has [a few variables](https://github.com/pelican-dev/panel/blob/577282094271076ac0977d8cc0fbcaaae46ba559/.env.docker). 

Is this intentional?

parkervcp:
> I'm not maintainer at all, but I still have a question :)
> 
> Previously, Pterodactyl Panel had a [quite long list](https://github.com/pterodactyl/panel/blob/develop/.env.example) of variables in its `.env.example` file. In this PR, the `.env.docker` file only has [a few variables](https://github.com/pelican-dev/panel/blob/577282094271076ac0977d8cc0fbcaaae46ba559/.env.docker).
> 
> Is this intentional?

From what I saw while testing, this runs a basic panel setup that uses an sqlite db. There are no variables needed for that.

parkervcp (approved):
Looks good to me. We should probably fire this up where it's actually on the internet to see how it goes before we fully release it.

We can add more compose files later. Some with all the things needed for redis/valkey and mariadb support.

compose.yml
```diff
@@ -0,0 +1,54 @@
...
+    volumes:
+      - pelican-data:/pelican-data
+    environment:
+      <<: [*panel-environment, *mail-environment]
+#      APP_ENV: "production"
+#      APP_DEBUG: "false"
+#      APP_ENVIRONMENT_ONLY: "false"
+#      SESSION_DRIVER: "file"
+
+volumes:
```
parkervcp, line 47: We need a volume for the certs that caddy generates.

parkervcp (changes requested):

compose.yml
parkervcp: We can use the `XDG_DATA_HOME` variable and set it to the same path as the pelican-data directory as well.

parkervcp: Should probably have a volume for logs as well.

luthes:
I've been digging into this a bit, both testing the Compose as well as the container builds, and tinkering with Helm. I run entirely in k8s any more, and having everything so tightly coupled in the current Dockerfile causes issues there.

That being said, I was able to split out Caddy and the Panel with Docker Compose (and Helm as well), by simply overriding commands, and editing and mounting the config file. But I do believe the base Dockerfile for the Panel should not bundle Caddy, each component (Caddy, Panel, Redis, etc.) should be it's own service.

From what I've been looking into this evening, what's needed to make this more micro-service friendly is:

     * Use volumes to provide access to `/var/www/html` files in Compose, rather than `COPY . .` in the Docker build.
     * Existing Panel Dockerfile should not bundle anything that's not necessary to running php-fpm. 
     * Caddyfile needs updated to use Docker overlay network, rather than localhost
     
Just a note that I'm willing to work on the above if you'd accept a PR. Let me know what you think of the above or if you're looking to go a different direction, happy to help.

lancepioch:
@luthes Can you please DM on Discord, I think you've got the right idea.

parkervcp:

parkervcp:
@LumiKattlin did you have this issue while building the container? I just tried out a build and it worked.

Have you tried this recently?

LumiKattlin:
Yeah, dw, it ended up being a mistake on my part

ended up working :+1:

VozDeOuro:
@parkervcp sorry for ping, do you have any ETA on when this will be merged ?

parkervcp:
Well that was ugly

compose.yml
```diff
@@ -0,0 +1,58 @@
+x-common:
+  panel:
+    &panel-environment
+    APP_URL: "https://localhost" # can be set to 'http://localhost' to un on port 80 only
```
parkervcp, line 4: this should be defaulting to `http://` not `https://` it throws errors about certificates.

parkervcp (changes requested):

composer.json
```diff
@@ -7,7 +7,6 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-pdo": "*",
-        "ext-pdo_mysql": "*",
```
parkervcp, line 10: We need to support sql from the container still.

parkervcp (changes requested):

parkervcp (approved):
LGTM

PR #316, opened 2024-06-03
Fix Node Updating

lancepioch:


notAreYouScared:
When the daemon key is rotated, the new `token_id` & `token` are not set to wings.

Boy132:
That's why the cloning is needed, to make sure it updates properly.
Just removing the `->save()` on the cloned node should fix the problem.

notAreYouScared (approved):

PR #327, opened 2024-06-04
Fix default sqlite database path in setup command

Boy132:
Since https://github.com/pelican-dev/panel/commit/a9e58bb493b9e087e54fbe5c0c567aa3904a6127 the database path is no longer absolute but relative.

### DIFF
--- a/app/Console/Commands/Environment/DatabaseSettingsCommand.php
+++ b/app/Console/Commands/Environment/DatabaseSettingsCommand.php
@@ -98,7 +98,7 @@ class DatabaseSettingsCommand extends Command
         } elseif ($this->variables['DB_CONNECTION'] === 'sqlite') {
             $this->variables['DB_DATABASE'] = $this->option('database') ?? $this->ask(
                 'Database Path',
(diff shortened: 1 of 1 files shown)


app/Console/Commands/Environment/DatabaseSettingsCommand.php
```diff
@@ -98,7 +98,7 @@ public function handle(): int
         } elseif ($this->variables['DB_CONNECTION'] === 'sqlite') {
             $this->variables['DB_DATABASE'] = $this->option('database') ?? $this->ask(
                 'Database Path',
-                config('database.connections.sqlite.database', database_path('database.sqlite'))
+                config('database.connections.sqlite.database', 'database.sqlite')
```
RMartinOscar, line 101: still has database_path() in config/database.php

Boy132: > still has database_path() in config/database.php

Yes, that's why it needs to be removed here.

lancepioch (approved):

Issue #317, opened 2024-06-03
Minecraft eggs problem.


JoanFont1212:
### Current Behavior

The current behavior is that when you start any minecraft service instead of picking the primary allocation picks the default allocation (25565) for all the servers, so if you have 3 servers all will start with 25565 every time, even if you change that in the server.properties, goes again to 25565.

### Expected Behavior

The behavior expected is to mantain the primary port instead of every time you start the server puts 25565 instead of your primary port.

### Steps to Reproduce

Update to the last version.
Have a minecraft server and start it.

### Panel Version

1.0.0-beta3

### Wings Version

1.0.0-beta3

### Games and/or Eggs Affected

All the eggs from Minecraft

### Docker Image

All the versions of the docker image.

### Error Logs

```bash
There isn't a "error log" for that.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

notAreYouScared:
Can you post a screenshot of all three servers allocation tables? 

The table is at the bottom of the edit server page

JoanFont1212:
The other two servers aren't accesible rn because aren't mine, i will show the server that is mine.
![image](https://github.com/pelican-dev/panel/assets/161775222/9bf02503-09fa-4044-be44-acaada71a40a)
As you can see, i have this ports and the 25516 is the primary.
Every time i open the server starts in 25565 because changes it in server.properties.

JoanFont1212:
(You cannot see, but 25516 is the primary, sorry for that)

JoanFont1212:
I have a better example, in the pterodactyl demo, you can see it in this screenshot.
![image](https://github.com/pelican-dev/panel/assets/161775222/1fa25b7e-d6bd-4a5f-bc23-bb2e9f7478cc)

Software-Noob:
Can you share the egg that you are using? The egg parser should modify the properties file to utilize the primary port. Either the egg is modified or there's an issue with the egg configuration service

notAreYouScared:
This is an issue with the panel not passing the right replace config to wings.

JoanFont1212:
I'm using the default eggs for Minecraft.

notAreYouScared:
We know about the issue, the fix is as follows. 

You can "fix" it yourself.

[...]

PR #331, opened 2024-06-04
Fix the egg variable disaster...

notAreYouScared:
This PR does the following

Updates stock eggs, with new values for the config->file updates
Updates all eggs currently in the database, with migrations
As eggs are imported they're updated, then saved.

If we update them as they're saved, any time the egg is exported it should have the correct variables. 

Closes https://github.com/pelican-dev/panel/issues/317


Boy132 (commented):
https://github.com/pelican-dev/panel/blob/charles/fixing-eggs/app/Services/Eggs/EggConfigurationService.php#L167 needs to be updated as well so `env.VARNAME` is replaced with `server.build.environment.VARNAME` instead of `server.build.env.VARNAME`.

Otherwise looks good.

app/Services/Eggs/Sharing/EggUpdateImporterService.php
```diff
@@ -27,6 +28,18 @@ public function handle(Egg $egg, UploadedFile $file): Egg
...
+            'server.build.env.SERVER_IP' => 'server.allocations.default.ip',
+            'server.build.default.ip' => 'server.allocations.default.ip',
+            'server.build.env.SERVER_PORT' => 'server.allocations.default.port',
+            'server.build.default.port' => 'server.allocations.default.port',
+            'server.build.env.SERVER_MEMORY' => 'server.build.memory_limit',
+            'server.build.memory' => 'server.build.memory_limit',
+            'server.build.env' => 'server.build.environment',
+        ];
+
+        $parsed = $this->recursiveStringReplace($parsed, $replacements);
```
lancepioch, line 41:
```suggestion
        $parsed = collect($replacements)->reduce(fn ($result, $value, $key) => str_replace($key, $value, $result), $parsed);
```

app/Services/Eggs/Sharing/EggUpdateImporterService.php
```diff
@@ -47,4 +60,16 @@ public function handle(Egg $egg, UploadedFile $file): Egg
...
+        return array_map(function ($value) use ($replacements) {
+            if (is_array($value)) {
+                return $this->recursiveStringReplace($value, $replacements);
+            } elseif (is_string($value)) {
+                return Str::replace(array_keys($replacements), array_values($replacements), $value);
+            }
+
+            return $value;
+        }, $data);
+    }
```
lancepioch, line 74:
```suggestion
```

app/Services/Eggs/Sharing/EggImporterService.php
```diff
@@ -45,4 +58,17 @@ public function handle(UploadedFile $file): Egg
...
+        return array_map(function ($value) use ($replacements) {
+            if (is_array($value)) {
+                return $this->recursiveStringReplace($value, $replacements);
+            } elseif (is_string($value)) {
+                return Str::replace(array_keys($replacements), array_values($replacements), $value);
+            }
+
+            return $value;
+        }, $data);
+    }
```
lancepioch, line 73:
```suggestion
```

app/Services/Eggs/Sharing/EggImporterService.php
```diff
@@ -25,6 +26,18 @@ public function handle(UploadedFile $file): Egg
...
+            'server.build.env.SERVER_IP' => 'server.allocations.default.ip',
+            'server.build.default.ip' => 'server.allocations.default.ip',
+            'server.build.env.SERVER_PORT' => 'server.allocations.default.port',
+            'server.build.default.port' => 'server.allocations.default.port',
+            'server.build.env.SERVER_MEMORY' => 'server.build.memory_limit',
+            'server.build.memory' => 'server.build.memory_limit',
+            'server.build.env' => 'server.build.environment',
+        ];
+
+        $parsed = $this->recursiveStringReplace($parsed, $replacements);
```
lancepioch, line 39:
```suggestion
        $parsed = collect($replacements)->reduce(fn ($result, $value, $key) => str_replace($key, $value, $result), $parsed);```

**Review by lancepioch — CHANGES_REQUESTED** [review]
I've got a couple small changes that I suggested, also can you please combine the 3 migrations into 1, thanks!

File: app/Services/Eggs/Sharing/EggImporterService.php [hunk 5]
```diff
@@ -25,6 +26,18 @@ public function handle(UploadedFile $file): Egg
     {
         $parsed = $this->parser->handle($file);
 
+        $replacements = [
```
DjordyKoert, line 29: This should be placed inside a `const`, `app/Services/Eggs/Sharing/EggUpdateImporterService.php` uses the same exact same array

app/Services/Eggs/Sharing/EggImporterService.php
```diff
@@ -45,4 +58,17 @@ public function handle(UploadedFile $file): Egg
             return $egg;
         });
     }
+
+    protected function recursiveStringReplace(array $data, array $replacements): array
```
DjordyKoert, line 62: Wouldn't it make sense to place this logic in a trait (or a different class / service)?

notAreYouScared: This was removed, and replaced with lances suggestion.

Boy132 (approved):
LGTM

app/Services/Eggs/Sharing/EggImporterService.php
notAreYouScared: I mean... It works <3

lancepioch: It should be only in the code once

notAreYouScared:
Looks good, tested this morning.

PR #332, opened 2024-06-05
Allow user to switch time zones

### LINKED ISSUE #192: Timezone selector
Users should be able to change the time zone that datetimes are shown in. This is separate from the app's timezone.

lancepioch:
Fixes #192


app/Filament/Resources/UserResource/Pages/EditProfile.php
```diff
@@ -82,6 +83,11 @@ protected function getForms(): array
                                             ->visible(fn (Get $get): bool => filled($get('password')))
                                             ->dehydrated(false),
 
+                                        Select::make('timezone')
+                                            ->prefixIcon('tabler-clock-pin')
```
Boy132, line 87: Make timezone required so it doesn't error when you try to save without a timezone selected.

```suggestion
                                        Select::make('timezone')
                                            ->required()
                                            ->prefixIcon('tabler-clock-pin')
```

Boy132 (dismissed):
`timezone` should be added to requests too, e.g. `StoreUserRequest`.


Also, do we want to change the cron input for schedules? Either add a note that it is always UTC or convert it from the users timezone.
Example: if I enter `0 10 * * *` it's actually 10 AM UTC and not my local timezone. So "Next run" would be displayed as 12 AM (because I use `Europe/Berlin`) which could be confusing.

notAreYouScared (approved):

PR #334, opened 2024-06-06
Better exception handling

### LINKED ISSUE #297: [Panel] - Proper exception reporting to the end user.
Many places in the panel will throw a 500 if a value is required but missing, or it won't report anything at all...

Ex: Eco egg requires a web server port, if left blank the create button appears to do nothing, and nothing tells the end user that the value is required.

lancepioch:
This doesn't handle very case, but it lays the ground work to be able to do that. Resolves #297 

Simply add the magic `exception` method to a Livewire component and optionally add custom members to the Exception!

### DIFF
--- a/app/Exceptions/Service/User/TwoFactorAuthenticationTokenInvalid.php
+++ b/app/Exceptions/Service/User/TwoFactorAuthenticationTokenInvalid.php
@@ -6,9 +6,9 @@ use App\Exceptions\DisplayException;
 
 class TwoFactorAuthenticationTokenInvalid extends DisplayException
 {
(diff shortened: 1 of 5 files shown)


notAreYouScared:
Thanks for adding this, Ideally the main area of concern are egg 
variable rule validation exceptions not getting passed to the end user on server create and edit.. resulting in a create button that appears to do nothing, or shows a 500, and the user has no idea why.

notAreYouScared (approved):
Tested, and works well! No more "ghost errors" when creating server with empty "required" values

PR #336, opened 2024-06-06
Use env value instead of config value for database path

Boy132:
This (finally) fixes the wrong default path for sqlite.


notAreYouScared (approved):
LGTM

Issue #193, opened 2024-05-10
Revamp Roles and Permissions


lancepioch:

Poseidon281:
Do you mean [#190](https://github.com/pelican-dev/panel/issues/190) by this or does something else need to be done (aswell)

timoschirmer:
Are there any plans to support SSO options via SAML/OIDC or any other way in this process?

KeySC07:
> Are there any plans to support SSO options via SAML/OIDC or any other way in this process?

Might be best to ask for this in #385 as it is more relevant. 

I did ask for the option to create user's upon first sign-in, which could be used to build on what you are asking for.

MCSHUBpanels:
**I would like to say a suggestion that I want some custom permission/ access to the panel**
_Unlike pterodactyl, when we make a person admin he / she got all the perms but I want him to have a specific access like my support staff , he / she gets access to see all the server only and system administrator gets access to eggs , nodes section and owner got all the access_

PR #361, opened 2024-06-10
Refactor api key permissions

Boy132:
Removes all `r_` columns for api key permissions and replaces these with a single array `permissions`.
Needed for plugins later.

This does _not_ handle #193 as it's only api key permissions.


database/migrations/2024_07_12_095213_fix_missing_sqlite_foreign_keys.php
```diff
@@ -32,18 +32,11 @@ public function up(): void
...
-                "r_users" integer not null default \'0\', 
-                "r_eggs" integer not null default \'0\', 
-                "r_database_hosts" integer not null default \'0\', 
-                "r_server_databases" integer not null default \'0\', 
                 "identifier" varchar, 
                 "key_type" integer not null default \'0\', 
                 "last_used_at" datetime, 
                 "expires_at" datetime,
-                "r_mounts" integer not null default \'0\',  
+                "permissions" text,
```
lancepioch, line 39: This migration already exists, so this should be put in a new migration, right?

Boy132: Yeah, that's because this PR is so old.

lancepioch (approved):

RMartinOscar (approved):

PR #319, opened 2024-06-03
Update AllocationsRelationManager to allow big endian

RMartinOscar:
Allow big endian
Adding 1025-1028 ports worked but 1028-1025 didn't before this

PR #364, opened 2024-06-11
Update CreateServer to allow big endian

RMartinOscar:
Following https://github.com/pelican-dev/panel/pull/319 so it also work on CreateServer page


Boy132 (approved):
LGTM

PR #366, opened 2024-06-11
Cleanup unused vars

RMartinOscar:
Removed unused $state

Unsure if `->map(fn ($exception) => $e->getTrace())` should be changed to
`->map(fn ($exception) => $exception->getTrace())` or
`->map(fn () => $e->getTrace())`

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -215,7 +215,7 @@ class Handler extends ExceptionHandler
                         ->map(fn ($trace) => Arr::except($trace, ['args']))
                         ->all(),
                     'previous' => Collection::make($this->extractPrevious($e))
-                        ->map(fn ($exception) => $e->getTrace())
+                        ->map(fn () => $e->getTrace())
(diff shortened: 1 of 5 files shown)


app/Exceptions/Handler.php
```diff
@@ -215,7 +215,7 @@ protected function convertExceptionToArray(\Throwable $e, array $override = []):
                         ->map(fn ($trace) => Arr::except($trace, ['args']))
                         ->all(),
                     'previous' => Collection::make($this->extractPrevious($e))
-                        ->map(fn ($exception) => $e->getTrace())
+                        ->map(fn () => $e->getTrace())
```
Boy132, line 218:
```suggestion
                        ->map(fn ($exception) => $exception->getTrace())
```

RMartinOscar: yeah that's what i thought but i wasn't sure

RMartinOscar:
Should be good to go

lancepioch (approved):

PR #367, opened 2024-06-11
Add default for Automatic Allocation in CreateNode

RMartinOscar:
Add ->default matching env PANEL_CLIENT_ALLOCATIONS_ENABLED to avoid 500 when creating a node and Automatic Allocation is not set

### DIFF
--- a/app/Filament/Resources/NodeResource/Pages/CreateNode.php
+++ b/app/Filament/Resources/NodeResource/Pages/CreateNode.php
@@ -212,7 +212,7 @@ class CreateNode extends CreateRecord
                                     false => 'success',
                                 ]),
                             Forms\Components\ToggleButtons::make('public')
-                                ->default(config('panel.client_features.allocations.enabled', false))
+                                ->default(true)
                                 ->columnSpan(1)
                                 ->label('Automatic Allocation')->inline()
                                 ->options([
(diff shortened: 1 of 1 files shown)


app/Filament/Resources/NodeResource/Pages/CreateNode.php
```diff
@@ -212,6 +212,7 @@ public function form(Forms\Form $form): Forms\Form
                                     false => 'success',
                                 ]),
                             Forms\Components\ToggleButtons::make('public')
+                                ->default(config('panel.client_features.allocations.enabled', false))
```
Boy132, line 215:
```suggestion
                                ->default(true)
```

This has nothing to do with "automatic allocations". `public` only controls if a node can be used for deployment.
"automatic allocations" is the feature that let users create a random allocation in the client area. (yes, the naming is a bit confusing)

Boy132 (changes requested):

app/Filament/Resources/NodeResource/Pages/CreateNode.php
Boy132: If you want you can change the label to something like `Useable for deployment`  to make it clearer and add a helper text that explains the field.

RMartinOscar:
Mybad i think it's ready to merge now

Boy132 (approved):
LGTM

PR #369, opened 2024-06-11
Update AllocationsRelationManager.php

RMartinOscar:
Comment allocation deletion on the server page until we make it dissociate before deleting to avoid 500
image


app/Filament/Resources/ServerResource/RelationManagers/AllocationsRelationManager.php
```diff
@@ -68,7 +68,7 @@ public function table(Table $table): Table
             ->bulkActions([
                 Tables\Actions\BulkActionGroup::make([
                     Tables\Actions\DissociateBulkAction::make(),
-                    Tables\Actions\DeleteBulkAction::make(),
+                    // Tables\Actions\DeleteBulkAction::make(),
```
lancepioch, line 71:
```suggestion
```

PR #376, opened 2024-06-11
Fix varchar(191) by replacing with 255 #135

### LINKED ISSUE #135: Fix varchar(191) by replacing with 255
https://www.grouparoo.com/blog/varchar-191

https://stackoverflow.com/questions/43832166/laravel-5-4-specified-key-was-too-long-why-the-number-191

RMartinOscar:
Edited Filament, Models, mysql-schema and made new migration to fix https://github.com/pelican-dev/panel/issues/135

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -15,7 +15,6 @@ use Illuminate\Pagination\Paginator;
 use Illuminate\Support\Facades\Broadcast;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Http;
-use Illuminate\Support\Facades\Schema;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Facades\View;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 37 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -29,7 +29,7 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot(): void
     {
-        Schema::defaultStringLength(191);
+        Schema::defaultStringLength(255);
```
lancepioch, line 32: Don't need this anymore!

database/migrations/2024_06_11_220722_update_field_length.php
```diff
@@ -0,0 +1,288 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('activity_log_subjects', function (Blueprint $table) {
+            $table->string('subject_type', 255)->change();
```
lancepioch, line 15: Omit 255 as the parameter for each of them, it's the default!

lancepioch (changes requested):

database/migrations/2024_06_11_220722_update_field_length.php
RMartinOscar: I tried without specifying it won't overwrite 191.

app/Providers/AppServiceProvider.php
RMartinOscar: Removed aswell as the useless import

Boy132 (approved):
LGTM

lancepioch (approved):

PR #378, opened 2024-06-12
GH-362: Hide container status & server state if nothing is displayed

### LINKED ISSUE #362: Implement new single Server State/Status display
We currently have 2 separate enums for server state and server status.
Server state refers to our own state such as reinstall, transfer, etc.
Server status refers to the docker container's status such as running, stopped, etc.

If the current server state exists, then show it. If it's empty, then we should show the current container status.

This also includes the labelling in the Edit Server page.

ItsNeil17:
Implement new single Server State/Status

> We currently have 2 separate enums for server state and server status.
> Server state refers to our own state such as reinstall, transfer, etc.
> Server status refers to the docker container's status such as running, stopped, etc.
> 
> If the current server state exists, then show it. If it's empty, then we should show the current container status.

> This also includes the labelling in the Edit Server page.

Fixes #362

### DIFF
--- a/app/Filament/Resources/ServerResource/Pages/EditServer.php
+++ b/app/Filament/Resources/ServerResource/Pages/EditServer.php
@@ -51,7 +51,7 @@ class EditServer extends EditRecord
                         $details = $service->setServer($server)->getDetails();
 
                         return $details['state'] ?? 'unknown';
-		    })
+                    })
                     ->hidden(fn (string $state) => $state === null)
                     ->options(fn ($state) => collect(ContainerStatus::cases())->filter(fn ($containerStatus) => $containerStatus->value === $state)->mapWithKeys(
                         fn (ContainerStatus $state) => [$state->value => str($state->value)->replace('_', ' ')->ucwords()]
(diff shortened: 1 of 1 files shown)


ItsNeil17:

app/Filament/Resources/ServerResource/Pages/EditServer.php
```diff
@@ -142,6 +144,7 @@ public function form(Form $form): Form
                                     ->columnSpanFull(),
 
                                 Forms\Components\TextInput::make('uuid')
+                                    ->label("UUID")
```
RMartinOscar, line 147:
```suggestion
                                    ->label("UUID")
```
```suggestion
                                    ->label('UUID')
```

app/Filament/Resources/ServerResource/Pages/EditServer.php
```diff
@@ -51,7 +51,8 @@ public function form(Form $form): Form
                         $details = $service->setServer($server)->getDetails();
 
                         return $details['state'] ?? 'unknown';
-                    })
+		    })
+                    ->hidden(fn (string $state) => $state === null)
```
RMartinOscar, line 55:
```suggestion
                    ->hidden(fn (string $state) => $state === null)
```
```suggestion
                    ->hidden(fn ($state) => $state === null)
```

RMartinOscar (commented):
Little type adjustment + Pint

ItsNeil17:
@RMartinOscar Did the changes you wanted, annotation is removed

lancepioch:
The issue isn’t just about hiding. The important part is the state needs to be single across the app.

ItsNeil17:
> The issue isn’t just about hiding. The important part is the state needs to be single across the app.

@iamkubi says it should not be one state:

> Although the way the API represents the state could be improved, it is important for performance reasons not to combine these 2 fields.

I can still combine these two if necessary, and I believe they would be stored inside "status" column of the servers table?

ItsNeil17:
Should I merge them then? @lancepioch

Issue #385, opened 2024-06-13
Oauth (via socialite)


Boy132:
https://laravel.com/docs/11.x/socialite
https://socialiteproviders.com/

- [x] Backend (#386)
- [x] Frontend (#718 & #416)
- [x] Add more providers (e.g. authentik, google, microsoft, steam) (#718)
- [x] Make it easier to add additional providers (without code edit) (#718)
- [x] Make OAuth settings configurable via panel UI (#839)

KeySC07:
Would be nice to see an option for JIT(Just-In-Time) provisioning added to create users upon sign-in via SSO.

timoschirmer:
Would love to see support for authentik or some kind of generic oauth2 provider in the future.

KeySC07:
> Would love to see support for authentik or some kind of generic oauth2 provider in the future.

Authentik is already supported see here https://socialiteproviders.com/Authentik/#installation-basic-usage
just gotta make some changes to get it working.

andyondrya:
Weird migration error,
```bash

   INFO  Running migrations.  

  2024_06_13_120409_add_oauth_column_to_users .............................................................................. 1.30ms FAIL

In Connection.php line 813:
                                                                                                                                         
  SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL (Connection: sqlite, SQL: alter table "users"   
  add column "oauth" text not null)                                                                                                      
                                                                                                                                         

In Connection.php line 571:
                                                                                          
  SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL

RMartinOscar:
> Weird migration error,
> 
> ```shell
>    INFO  Running migrations.  
> 
>   2024_06_13_120409_add_oauth_column_to_users .............................................................................. 1.30ms FAIL
> 
> In Connection.php line 813:
>                                                                                                                                          
>   SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL (Connection: sqlite, SQL: alter table "users"   
>   add column "oauth" text not null)                                                                                                      
>                                                                                                                                          
> 
> In Connection.php line 571:
>                                                                                           
>   SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL
> ```

Fixed in [PR #409](https://github.com/pelican-dev/panel/pull/409)

KeySC07:
[...]

PR #386, opened 2024-06-13
Add OAuth backend

Boy132:
Backend for #385.
Doing the frontend only makes sense after we converted the client area to filament too.

For now I only added configs for `github` and `discord` for testing. More providers should be added when the frontend is been worked on.
I tested `github` by manually calling the endpoints, `discord` is untested but since the underlying system is the same it should also work.


app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Callback from OAuth provider.
+     */
+    protected function callback(Request $request, string $driver): RedirectResponse
+    {
+        $oauthUser = Socialite::driver($driver)->user();
+
+        // User is already logged in and wants to link a new OAuth Provider
+        if ($request->user() != null) {
```
lancepioch, line 42:
```suggestion
        if ($request->user()) {
```

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -0,0 +1,62 @@
...
+            $this->updateService->handle($request->user(), ['oauth' => json_encode($oauth)]);
+
+            return redirect()->route('account');
+        }
+
+        try {
+            $user = User::query()->whereJsonContains('oauth->'. $driver, $oauthUser->getId())->firstOrFail();
+
+            $this->auth->guard()->login($user, true);
+        } catch (\Exception $e) {
```
lancepioch, line 55:
```suggestion
        } catch (Exception $e) {
```

Import please

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -0,0 +1,62 @@
...
+            // No user found - redirect to normal login
+            return redirect()->route('auth.login');
+        }
+
+        return redirect('/');
```
lancepioch, line 60: is there a home route?

app/Http/Controllers/Base/OauthController.php
```diff
@@ -0,0 +1,44 @@
...
+        return Socialite::with($driver)->redirect();
+    }
+
+    /**
+     * Remove a OAuth link
+     */
+    protected function unlink(Request $request): Response
+    {
+        $oauth = $request->user()->oauth;
+        unset($oauth[$request->get('driver')]);
```
lancepioch, line 38: Should we use ->only in combination with this?

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -0,0 +1,62 @@
...
+            $oauth = $request->user()->oauth;
+            $oauth[$driver] = $oauthUser->getId();
+
+            $this->updateService->handle($request->user(), ['oauth' => json_encode($oauth)]);
+
+            return redirect()->route('account');
+        }
+
+        try {
+            $user = User::query()->whereJsonContains('oauth->'. $driver, $oauthUser->getId())->firstOrFail();
```
lancepioch, line 52: Nice!

config/services.php
```diff
@@ -9,4 +9,16 @@
         'scheme' => 'https',
     ],
 
+    'github' => [
+        'client_id' => env('OAUTH_GITHUB_CLIENT_ID'),
+        'client_secret' => env('OAUTH_GITHUB_CLIENT_SECRET'),
+        'redirect' => '/auth/oauth/callback/github',
+    ],
+
```
lancepioch, line 17:
```suggestion
```

Don't think we need this as a default, plugin probably.

routes/auth.php
```diff
@@ -18,6 +18,10 @@
 Route::get('/password', [Auth\LoginController::class, 'index'])->name('auth.forgot-password');
 Route::get('/password/reset/{token}', [Auth\LoginController::class, 'index'])->name('auth.reset');
 
+// Endpoints for OAuth
+Route::get('/oauth/redirect', [Auth\OAuthController::class, 'redirect'])->name('auth.oauth.redirect');
```
lancepioch, line 22:
```suggestion
Route::get('/oauth/redirect/{driver}', [Auth\OAuthController::class, 'redirect'])->name('auth.oauth.redirect');
```

Looks nicer in the url

lancepioch (changes requested):

config/services.php
Boy132: It's included by default in socialite and nice for testing. That's why I added it.

app/Http/Controllers/Auth/OAuthController.php
Boy132: Don't think so?

app/Http/Controllers/Base/OauthController.php
Boy132: Not sure what you mean by this?

lancepioch: $request->only(['necessaryForHandleVars', 'driver']),

In case somebody adds more

config/services.php
lancepioch: Ah okay, sounds good then!

PR #387, opened 2024-06-13
Created a new upgrade command 

Poseidon281:
This command introduces new features to the upgrade process in with just 2 clicks you can upgrade the panel to the latest release

Any suggestions are always welcome

#262

### DIFF
--- a/app/Console/Commands/UpgradeCommand.php
+++ b/app/Console/Commands/UpgradeCommand.php
@@ -3,9 +3,10 @@
 namespace App\Console\Commands;
 
 use Illuminate\Console\Command;
-use App\Console\Kernel;
 use Symfony\Component\Process\Process;
 use Symfony\Component\Console\Helper\ProgressBar;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Config;
 
 class UpgradeCommand extends Command
 {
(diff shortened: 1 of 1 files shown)


Boy132:
There is a reason why the update command is no longer used (or recommended).
The main problem is that the update command updates the code that it needs to run itself. This can cause problems.

Poseidon281:
> There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.


There is propably a way to exclude the upgrade command from being updated by this command and once the upgrade command is merged when everyone has looked over it. It does not need to be changed again probably.

Poseidon281:
> There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.

But i have tested it of course like 10 times and i have not had any problems even tho the command changed back to before i changed it when i ran it of course

Boy132:
> > There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.
> 
> There is propably a way to exclude the upgrade command from being updated by this command and once the upgrade command is merged when everyone has looked over it. It does not need to be changed again probably.

That's not what I meant. While the artisan command is running it will change panel files (and dependencies). This can cause problems depending on the update.

Poseidon281:
> > > There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.
> > 
> > 
> > There is propably a way to exclude the upgrade command from being updated by this command and once the upgrade command is merged when everyone has looked over it. It does not need to be changed again probably.
> 
> That's not what I meant. While the artisan command is running it will change panel files (and dependencies). This can cause problems depending on the update.

Sorry, could you explain. What problems could occur? Maybe if dependencies from the command are removed, but why would that happen.

Boy132:
> > > > There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.
> > > 
> > > 
> > > There is propably a way to exclude the upgrade command from being updated by this command and once the upgrade command is merged when everyone has looked over it. It does not need to be changed again probably.
> > 
> > 
> > That's not what I meant. While the artisan command is running it will change panel files (and dependencies). This can cause problems depending on the update.
> 
> Sorry, could you explain so i maybe can change the command to fix that. What problems could occur? Maybe if dependencies from the command are removed, but why would that happen.

A simple example would be a Laravel update. Maybe it changes the method parameters.
Now the command starts on Version A, downloads the new files/ updates dependencies and suddenly you are on Version B and have a syntax error.
There _could_ be many other problems. Updating something while it's running is generally not stable. That's why self updaters normally run on a separate/ external process.

Poseidon281:
> > > > > There is a reason why the update command is no longer used (or recommended). The main problem is that the update command updates the code that it needs to run itself. This can cause problems.
> > > > 
> > > > 
> > > > There is propably a way to exclude the upgrade command from being updated by this command and once the upgrade command is merged when everyone has looked over it. It does not need to be changed again probably.
> > > 
> > > 
> > > That's not what I meant. While the artisan command is running it will change panel files (and dependencies). This can cause problems depending on the update.
> > 
> > 
> > Sorry, could you explain so i maybe can change the command to fix that. What problems could occur? Maybe if dependencies from the command are removed, but why would that happen.
> 
> A simple example would be a Laravel update. Maybe it changes the method parameters. Now the command starts on Version A, downloads the new files/ updates dependencies and suddenly you are on Version B and have a syntax error. There _could_ be many other problems. Updating something while it's running is generally not stable. That's why self updaters normally run on a separate/ external process.

So if not a artisan command. What do you suggest for "auto" updating the Panel. A .sh script?

Boy132:
> So if not a artisan command. What do you suggest for "auto" updating the Panel. A .sh script?

Nothing. Manually updating is not that hard.
And once the dockerized install becomes the preferred method of install updating will be even simpler.

Poseidon281:
> > So if not a artisan command. What do you suggest for "auto" updating the Panel. A .sh script?
> 
> Nothing. Manually updating is not that hard. And once the dockerized install becomes the preferred method of install updating will be even simpler.

Oke

PR #388, opened 2024-06-13
Rename OauthController.php to OAuthController.php

RMartinOscar:
Fixes `Class App\Http\Controllers\Base\OAuthController located in ./app/Http/Controllers/Base/OauthController.php does not comply with psr-4 autoloading standard (rule: App\ => ./app). Skipping.`


RMartinOscar:
Also don't know how did it passed [tests](https://github.com/pelican-dev/panel/blob/main/.github/workflows/ci.yaml#L70) without warning

lancepioch (approved):

Issue #353, opened 2024-06-09
Main Application translated to Filament


lancepioch:

Poseidon281:
Do you mean the client side by this or?

Boy132:
> Do you mean the client side by this or?

Yes, the client side.

PR #403, opened 2024-06-16
Add DatabaseHost alias

RMartinOscar:
Add alias to DatabaseHosts and display that instead of the ip in the endpoint.
Useful when you create pelicanuser@127.0.0.1 but still want to show either a domain or for example 172.18.0.1
List
Edit
Create
Client side eye
Client side list


lancepioch:
Sorry I don't get what this solves. You should be able to use the domain in the host column and the name is the "alias"

Boy132:
> Sorry I don't get what this solves. You should be able to use the domain in the host column and the name is the "alias"

It's for users that use a local ip or `localhost` for the host but want to show a different value.

notAreYouScared:
> > Sorry I don't get what this solves. You should be able to use the domain in the host column and the name is the "alias"
> 
> It's for users that use a local ip or `localhost` for the host but want to show a different value.

We have a 'Display Name' section. The admin could just put `db.pelican.dev` and that would be displayed to the end-user on the connection details. Unless that's broken?

RMartinOscar:
> We have a 'Display Name' section. The admin could just put `db.pelican.dev` and that would be displayed to the end-user on the connection details. Unless that's broken?

The Display name is only for the admin side its not passed to react and never was

Boy132 (changes requested):
Like Charles said the `Display Name` is supposed to be used as an alias.
Could change the PR to fix that? Thanks.

RMartinOscar:
> Like Charles said the `Display Name` is supposed to be used as an alias. Could change the PR to fix that? Thanks.

That means i'd have to edit the old client part is it worth doing ?

Boy132:
> > Like Charles said the `Display Name` is supposed to be used as an alias. Could change the PR to fix that? Thanks.
> 
> I thought it was just a name on the admin side kinda like for Nodes That means i'd have to edit the old client part is it worth doing or should i wait and reimplement it after [#353](https://github.com/pelican-dev/panel/issues/353) ?

You can do it now, it's just a small edit. #353 will take a while.

RMartinOscar:
I was looking for issues to solve on Pterodactyl's git and i stumbled upon [this one](https://github.com/pterodactyl/panel/issues/5096) that made me unsure about the solution, should i follow what was previously said or implement it differently to comply with that issue aswell ?

lancepioch:
After reading the original issue, I think it has some merit, however I think we should think more carefully about the execution. Let's continue this in a repo discussion!

PR #407, opened 2024-06-16
Panel: Fix wings stoplogic

QuintenQVD0:
# Changes

This is a PR that needs to be combined with the wings PR I am making in about 5 minutes. This changes that the stop `^C` get transferd in to a generic stop but a full signal that the wings can understand.

Needs: https://github.com/pelican-dev/wings/pull/31


lancepioch (approved):

QuintenQVD0:
For anyone merging this. The wings change must be made else this will just kill all containers and say unknown stop signal terminat

PR #409, opened 2024-06-17
Update OAuth migration

RMartinOscar:
Make oauth column nullable (JSON fields cannot have default value)
```
INFO  Running migrations.  

  2024_06_13_120409_add_oauth_column_to_users .............................................................................. 1.30ms FAIL

In Connection.php line 813:

  SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL (Connection: sqlite, SQL: alter table "users"   
  add column "oauth" text not null)


In Connection.php line 571:

  SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL
```


Boy132 (approved):
LGTM

PR #395, opened 2024-06-15
Replace Allocations with Endpoints

lancepioch:
Resolves #68

PR #411, opened 2024-06-17
Add exception handling to CreateNode/ EditNode & CreateServer/ EditServer

Boy132:
We probably want to add some more fine tuned exception handling at some point but for now this stops all 500 errors or silent errors and replaces them with a notification.


lancepioch (approved):
Looks good, I wonder if we should use traits for now?

notAreYouScared:
I attempted to test this with editing a server but it caused the panel to lock up...

Edit Server -> Remove assigned user, click save.

Boy132:
> I attempted to test this with editing a server but it caused the panel to lock up... image
> 
> Edit Server -> Remove Owner, click save.
> 
> This seems to break the stock error handling as the owner is a required field.

Yeah, seems like Filament doesn't like the livewire `exception` hook. I used good old try-catchs now.

notAreYouScared:
Anyway we can catch the 500 on the create allocation, on the create server page? 

image
image
image


Unless we just need to rewrite the logic, the port already exists... we should just skip over it, and present an error saying something along the lines of

Failed to Create Allocation
Failed to Create: <IP>:<Port>, <error>

Boy132:
> Anyway we can catch the 500 on the create allocation, on the create server page?
> 
> image image image
> 
> Unless we just need to rewrite the logic, the port already exists... we should just skip over it, and present an error saying something along the lines of
> 
> Failed to Create Allocation Failed to Create: :,

Wouldn't #395 fix/ change that anyways?

notAreYouScared:
It would, but who knows when that would be done.

notAreYouScared:
With latest changes 500's are still thrown when it skips the ports and does not return anything

image

lancepioch (changes requested):
Memory leak

notAreYouScared:
^^  Memory Leak when creating ports that **already** exist. 
image

Boy132:
Closing this for now. Will look into that later, probably.

PR #415, opened 2024-06-17
Add required to editing server variables and fix #413

### LINKED ISSUE #413: [Edit Server] Allows saving of non-nullable | required egg variables.
When editing a server 
image

The rules are `required|string|max:64`

But still allows it to be empty on save and saves a null value

lancepioch:


notAreYouScared (approved):

notAreYouScared (changes requested):
This results in a 500 error when loading the edit server page. 
image

notAreYouScared (approved):

PR #417, opened 2024-06-18
Combine Server states and statuses and resolve #362

### LINKED ISSUE #362: Implement new single Server State/Status display

lancepioch:


Boy132 (approved):
LGTM

notAreYouScared (approved):

Issue #428, opened 2024-06-25
Allocations can not be created with an ipv6 address


theoparis:
### Current Behavior

![image](https://github.com/pelican-dev/panel/assets/11761863/f36f727d-f9bd-4121-997a-50d7760f9f3b)

The panel complains that it has to be an ipv4 address 😬 

### Expected Behavior

.

### Steps to Reproduce

Enter a ipv6 address in a node allocation 

### Panel Version

1.0.0-beta5

### Wings Version

v1.0.0-beta3

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

RMartinOscar:
Hey will be fixed with PR #429

lancepioch:
This is correct behavior. We do not support ipv6.

RMartinOscar:
Should be _Not planed_ instead of _Completed_

PR #429, opened 2024-06-25
Update Allocation fields to allow ipv4 and ipv6 #428

RMartinOscar:


notAreYouScared:
@RMartinOscar Have you tested this with wings? Do IPv6 Allocations correctly translate into wings and get assigned to servers? I don't do much of anything with IPv6 so my knowledge is limited... 

@iamkubi any thing on the wings side that this might break?

lancepioch:
We do not support ipv6 at all or at any level afaik. This is something at the bottom of the backlog.

PR #432, opened 2024-06-26
Add MariaDB Changes, Split out Caddy to it's own container

luthes:
Lance and I worked on this a bit this evening. There may be some work left to do here, but this essentially splits Caddy, the Panel, and adds in the MariaDB database. A `docker compose up` in the main directory will get it running. I'll work on adding Redis next. 

I think splitting this out to a full cluster setup, and something a bit more like what Lance had in the original PR with a single container/sqlite is a good idea. Maybe something like compose-cluster.yml, compose[-sqlite].yml, and flagging compose-cluster.yml as a more advanced setup option might be a good idea.

Apologies for the second PR. I based this branch off of Lance's issue/311 PR, but it needs a sync to main, and this can be pointed back over there without being too messy.


.env.docker
```diff
@@ -0,0 +1,3 @@
+APP_KEY=
+
+DB_DATABASE=docker/database.sqlite
```
luthes, line 3: If using  MariaDB, maybe even with sqlite, this can be removed as it's set in the compose.yml.

lancepioch:
Thank you Steven, please check out the other original PR. Parker and I are keeping Caddy in the single image, however we added a flag to just skip running Caddy, so you should be able to run the Panel still with K8S.

PR #434, opened 2024-06-26
Separate these out in the navigation

lancepioch:
After: 
image

Before:
image


notAreYouScared:
I see no reason why this couldn't be in Beta6 if its done <3

notAreYouScared (approved):

PR #259, opened 2024-05-19
Settings page & few small improvements

Poseidon281:
I made a settings page in the Filament admin side.

### TODO List 
✅ = Done
❔= Need help with/dont know how
❌ = Needs to be done/not yet started
| Task | Status | Number |
|--------|--------|--------|
| Only update settings when appy settings is pressed | ✅ | 1 |
| Add descriptions to the settings| ✅ | 2 |
| Add more settings | Working on it | 3 |
| Apply Settings needs to be triggerd without the page expiring | ❔/ No priority | 4 |
| Add [tabs](https://filamentphp.com/docs/3.x/forms/layout/tabs) to the Settingspage | ✅ | 5 |
| Add last updated column | ✅ | 7 |
| Only trigger setConfigFromDatabase when Apply Settings is pressed and not on its own | ✅ | 8 |
| Implement this [Plugin](https://filamentphp.com/plugins/tallyschmeits-character-counter) for the settings | Removed | 9
| Remove the database and make it a livewire page | ✅ | 10 |
| Click anywhere it edits the settings | ✅ | 10 |
| Make sure that the options and the limit works | ✅ | 11 |
| Make that saving and appyling the settings works | ✅ | 12 |
| Updating or removing the old API for the settings | ❔/ No priority | 13 |
| Creating [Import](https://filamentphp.com/docs/3.x/actions/prebuilt-actions/import) and [Export](https://filamentphp.com/docs/3.x/actions/prebuilt-actions/export#automatically-generating-exporter-columns) | Removed | 14 |
| Fix the saving bug |✅| 15 |
| Removing the settings table in the DB and the SettingsServiceProvider | Will be done later by Lance | 16 |

Reminder 2: Talk to Lance about a paginated setting for the tables

PR #435, opened 2024-06-27
Finish OAuth on the backend

Poseidon281:
I also have an idea for "(Optional) Make it easier to add additional providers (without code edit)", but for that the [Settings Page](https://github.com/pelican-dev/panel/pull/259) needs to be merged in first because i want to make it in the new model

This PR works on #385


Boy132:
Thanks but we'll add more providers once the oauth frontend is done. It will also be less hardcoded.

PR #450, opened 2024-06-30
Added description and notes text fields to Nodes

Poseidon281:
Suggested: [Here](https://github.com/pelican-dev/panel/discussions/448)


RMartinOscar:
What's the difference between description and notes ?

NerdsCorp:
I don't think there would be a major difference. Is there already an option for a description on a node?

RMartinOscar:
> I don't think there would be a major difference. Is there already an option for a description on a node?

The field is in place sinc pterodactyl but currently never used

NerdsCorp:
@RMartinOscar the difference in my thought is I'd like to add notes, not just a description. Like if it's in the cloud or in my home lab at home. Also, last time I serviced that particular node and when it was put into service. I think a description would work just as well as long as it allowed you to add a few lines, not just a single line. I think the one pterodactyl had was great!

RMartinOscar:
> @RMartinOscar the difference in my thought is I'd like to add notes, not just a description. Like if it's in the cloud or in my home lab at home. Also, last time I serviced that particular node and when it was put into service. I think a description would work just as well as long as it allowed you to add a few lines, not just a single line. I think the one pterodactyl had was great!

The fact that its a text field means its limited to 65 535 chars imo that's plenty enough

NerdsCorp:
@RMartinOscar i did not know that! I think the one from pterodactyl would be just fine! 😁

app/Filament/Resources/NodeResource/Pages/CreateNode.php
```diff
@@ -191,6 +192,12 @@ public function form(Forms\Form $form): Forms\Form
                                     'https' => 'tabler-lock',
                                 ])
                                 ->default(fn () => request()->isSecure() ? 'https' : 'http'),
+
+                            Textarea::make('notes')
+                                ->label('Notes')
```
RMartinOscar, line 197:
```suggestion
                            Textarea::make('description')
                                ->label('Description')
```

app/Filament/Resources/NodeResource/Pages/EditNode.php
```diff
@@ -193,7 +193,15 @@ public function form(Forms\Form $form): Forms\Form
                                     'http' => 'tabler-lock-open-off',
                                     'https' => 'tabler-lock',
                                 ])
-                                ->default(fn () => request()->isSecure() ? 'https' : 'http'), ]),
+                                ->default(fn () => request()->isSecure() ? 'https' : 'http'),
+
+                            Textarea::make('notes')
+                                ->label('Notes')
```
RMartinOscar, line 199:
```suggestion
                            Textarea::make('description')
                                ->label('Description')
```

app/Models/Node.php
```diff
@@ -20,6 +20,7 @@
  * @property string|null $description
  * @property string $fqdn
  * @property string $scheme
+ * @property string|null $notes
```
RMartinOscar, line 23:
```suggestion
```

database/migrations/2024_06_30_123658_add_notes_to_nodes.php
RMartinOscar, line 1: Remove

app/Models/Node.php
RMartinOscar, line 1: Keep current file

RMartinOscar (commented):
Using description like Pterodactyl instead of Notes

Poseidon281:
> Using description like Pterodactyl instead of Notes

If the Pelican Team wants me to I will switch the field to description, but I think there is a real difference between notes and description because I would have used description if there was not. So for now I am going with notes.

lancepioch:
Pull requests must be associated with an existing issue.

PR #451, opened 2024-07-01
Added subheadings to list pages & sidebar collapsible config

Poseidon281:
The subheadings are the same as on Pterodactyl at the moment


lancepioch:
Pull requests must be associated with an existing issue.

PR #459, opened 2024-07-04
Add Node CPU/Memory Graphs

notAreYouScared:
Adds an overview tab to EditNode, Shows basic system information, and current host node cpu/memory usage, updated every 5 seconds 

image

### DIFF
--- a/app/Filament/Resources/NodeResource/Widgets/NodeCpuChart.php
+++ b/app/Filament/Resources/NodeResource/Widgets/NodeCpuChart.php
@@ -11,7 +11,7 @@ class NodeCpuChart extends ChartWidget
 {
     protected static ?string $heading = 'CPU';
     protected static ?string $pollingInterval = '5s';
-
+    protected static ?string $maxHeight = '300px';
     public ?Model $record = null;
 
     protected static ?array $options = [
@@ -79,7 +79,7 @@ class NodeCpuChart extends ChartWidget
                     ],
                     'tension' => '0.3',
                     'fill' => true,
-                    'label' => 'CPU Average 1',
+                    'label' => 'Load Average - 1m',
                 ],
                 [
                     'data' => array_column($cpu5, 'cpu5'),
@@ -88,7 +88,7 @@ class NodeCpuChart extends ChartWidget
                     ],
                     'tension' => '0.3',
                     'fill' => true,
-                    'label' => 'CPU Average 5',
+                    'label' => 'Load Average - 5m',
                 ],
                 [
                     'data' => array_column($cpu15, 'cpu15'),
@@ -97,7 +97,7 @@ class NodeCpuChart extends ChartWidget
                     ],
                     'tension' => '0.3',
                     'fill' => true,
-                    'label' => 'CPU Average 5',
+                    'label' => 'Load Average - 15m',
                 ],
             ],
             'labels' => array_column($cpu, 'timestamp'),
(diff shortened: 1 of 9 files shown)


notAreYouScared:
Newest Commit 

image

app/Filament/Resources/NodeResource/Widgets/NodeCpuChart.php
```diff
@@ -0,0 +1,80 @@
...
+    protected function getData(): array
+    {
+        /** @var Node $node */
+        $node = $this->record;
+
+        $cpu = collect(cache()->get("nodes.$node->id.cpu_percent"))
+            ->slice(-10)
+            ->map(fn ($value, $key) => [
+                'cpu' => $value,
+                'timestamp' => Carbon::createFromTimestamp($key)->format('H:i:s'),
```
Boy132, line 27: Is the time displayed in the user timezone? Or app timezone (UTC)?
This looks like app timezone.

app/Filament/Resources/NodeResource/Widgets/NodeMemoryChart.php
```diff
@@ -3,66 +3,78 @@
...
     {
         /** @var Node $node */
         $node = $this->record;
 
-        $total = ($node->statistics()['memory_total'] ?? 0) / 1024 / 1024 / 1024;
-        $used = ($node->statistics()['memory_used'] ?? 0) / 1024 / 1024 / 1024;
-        $unused = $total - $used;
+        $memUsed = collect(cache()->get("nodes.$node->id.memory_used"))->slice(-10)
+            ->map(fn ($value, $key) => [
+                'memory' => $value / 1024 / 1024 / 1024,
```
Boy132, line 25: All values should be converted to/ from GB/ GiB based on the config value for the binary prefix. (`config('panel.use_binary_prefix')`)

notAreYouScared: Why even give the option... Seems soo extra.

Boy132: Because most people want to use MB/ GB. They don't want MiB/ GiB.

notAreYouScared: Isn't MegaByte, MB. 1024? That's what I'm using? I think our plan was to always use MB/GB

Boy132: MegaByte = MB = 1000 KB (decimal prefix)
MebiByte = MiB = 1024 KiB (binary prefix)

This is why there is a config option. It's very easy to mess these up.

notAreYouScared: The latest commit should correctly use the config variable.

app/Filament/Resources/NodeResource/Widgets/NodeCpuChart.php
notAreYouScared: Set to use the users timezone, if that for some reason does not exist it falls back to UTC

Boy132 (commented):
Timezone and binary prefix stuff looks good.

But should we add a little note that in the cpu graph 100% doesn't mean 1 thread? Otherwise it might be confusing/ inconsistent.
Or we divide it by the threads count?

notAreYouScared:
> Timezone and binary prefix stuff looks good.
> 
> But should we add a little note that in the cpu graph 100% doesn't mean 1 thread? Otherwise it might be confusing/ inconsistent. Or we divide it by the threads count?

It's returned from wings as 100% is the cpu as a whole. 

@iamkubi could we change it ?

Boy132 (approved):
LGTM

PR #463, opened 2024-07-07
Use database for notifications

notAreYouScared:
Adds a notification bell on top bar and  saves notification history. 

For future use in the front end, ex. Server crashes.

### DIFF
--- a/app/Exceptions/DisplayException.php
+++ b/app/Exceptions/DisplayException.php
@@ -53,7 +53,8 @@ class DisplayException extends PanelException implements HttpExceptionInterface
                 ->title(static::class)
                 ->body($this->getMessage())
                 ->danger()
-                ->send();
+                ->send()
+                ->sendToDatabase(auth()->user());
 
             return;
         }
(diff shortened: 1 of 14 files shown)


RMartinOscar:
Hey i believe we can use `$panel->databaseNotifications()` in _AdminPanelProvider.php_ with [Filament 3](https://filamentphp.com/docs/3.x/panels/notifications) to avoid repetitiveness

notAreYouScared:
> Hey i believe we can use `$panel->databaseNotifications()` in _AdminPanelProvider.php_ with [Filament 3](https://filamentphp.com/docs/3.x/panels/notifications) to avoid repetitiveness

We already use that. That's how you get the notification panel to appear at the top

lancepioch (changes requested):
This basically duplicates the activity log. It should be consolidated.

Boy132 (commented):
I agree with Lance. We should only send database notifications for "important" stuff. Sending both notifications for a simply success or error notifications seems unnecessary.

notAreYouScared:
> This basically duplicates the activity log. It should be consolidated.

Are you referring to the client area activity log? This would rmsotly replace that when the new client area comes.

I still plan to keep activity logs per server.

Also there is no loy for the admin side.

PR #339, opened 2024-06-06
PHPStan workflow + PHPStan fixes

DjordyKoert:
I noticed that PHPStan was available but not actually being used. 

This PR contains a few changes & fixes, namely:
- PHPStan is now ran in CI
- Added PHPStan script (running PHPStan locally)
- A bunch of fixes that I encountered while running PHPStan locally
	- I was not able to fix some though because I don't have the required knowledge of these parts (or maybe some of these are actually bugged currently) . These should be added to baseline or fixed if someone knows how to.

```txt
 ------ -------------------------------------------------------------------------------------- 
  Line   Filament/Resources/UserResource/Pages/EditProfile.php                                 
 ------ -------------------------------------------------------------------------------------- 
  142    Access to an undefined property chillerlan\QRCode\QROptions::$svgLogo.                
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
  143    Access to an undefined property chillerlan\QRCode\QROptions::$svgLogoScale.           
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
  147    Access to protected property chillerlan\QRCode\QROptions::$version.                   
  149    Access to protected property chillerlan\QRCode\QROptions::$outputBase64.              
  150    Access to protected property chillerlan\QRCode\QROptions::$eccLevel.                  
  151    Access to protected property chillerlan\QRCode\QROptions::$addQuietzone.              
  153    Access to protected property chillerlan\QRCode\QROptions::$connectPaths.              
  154    Access to protected property chillerlan\QRCode\QROptions::$drawCircularModules.       
  157    Access to protected property chillerlan\QRCode\QROptions::$svgDefs.                   
 ------ -------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------- 
  Line   Models/Node.php                                                                       
 ------ -------------------------------------------------------------------------------------- 
  242    Access to an undefined property App\Models\Node::$servers_sum_memory.                 
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
  249    Access to an undefined property App\Models\Node::$servers_sum_disk.                   
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
  256    Access to an undefined property App\Models\Node::$servers_sum_cpu.                    
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
[...]

PR #466, opened 2024-07-08
Fix client Activity tab issues; fixes #465

### LINKED ISSUE #465: Client Activity tab does not work properly
### Current Behavior

If a login is failed with a valid username/email, the target user will have their Activity tab on their profile fail to render.
Second issue is that auth logs are duplicated for any auth event.

### Expected Behavior

The Activity tab should still render properly even when an auth attempt has failed.
Auth logs should not be logged twice.

### Steps to Reproduce

1. Attempt to log in with a valid username/email but an incorrect password.
2. Check the Activity tab in the user profile. It will fail to render.

### Panel Version

1.0.0-beta6

### Wings Version

1.0.0-beta3

N/A

### Docker Image

N/A

### Error Logs

_No response_

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

ItzExotical:
This fixes both issues with the Activity tab, both duplicate entries as well as the page breaking after a failed auth attempt.
Tested and confirmed working on 2 different panel instances.

The second issue mentioned above was originally fixed by [this](https://github.com/DjordyKoert/panel/commit/065f3f246803fbfa8517559f82c35d43cb426679) commit by @lancepioch but was later overwritten (presumably by accident) on @DjordyKoert's commit [here](https://github.com/pelican-dev/panel/pull/339/commits/4090c66af07442be3c5c37049a55972e3991f100#diff-e3f0808490450e9dd8161858e6d57f67e622bc1a1e32323c96be445dac96ad95R58).

Huge thanks to @RMartinOscar for discovering why the log function was being called several times (because of entries in the providers file).


app/Transformers/Api/Client/ActivityLogTransformer.php
```diff
@@ -55,7 +55,7 @@ protected function properties(ActivityLog $model): object
 
         $properties = $model->properties
             ->mapWithKeys(function ($value, $key) use ($model) {
-                if ($key === 'ip' && !$model->actor->is($this->request->user())) {
+                if ($key === 'ip' && $model->actor && !$model->actor->is($this->request->user())) {
```
RMartinOscar, line 58:
```suggestion
                if ($key === 'ip' && $model->actor instanceof User && !$model->actor->is($this->request->user())) {
```
or
```suggestion
                if ($key === 'ip' && isset($model->actor) && !$model->actor->is($this->request->user())) {
```
or
```suggestion
                if ($key === 'ip' && !optional($model->actor)->is($this->request->user())) {
```

RMartinOscar (commented):
To make PHPStan happy cause we've clearly shown that `$model->actor` can indeed be null

bootstrap/providers.php
```diff
@@ -4,10 +4,8 @@
     App\Providers\ActivityLogServiceProvider::class,
     App\Providers\AppServiceProvider::class,
     App\Providers\BackupsServiceProvider::class,
-    App\Providers\EventServiceProvider::class,
```
Boy132, line 7: I haven't double checked yet but removing the `EventServiceProvider` looks like a bad idea.
This will most likely break stuff like subuser email sending.

ItzExotical: Server subuser emailing seems to work just fine with it removed.
image

ItzExotical: Server subuser emailing seems to work just fine even with it removed.
image

Boy132: Is that a fresh install or did you upgrade from previous beta versions?

ItzExotical: Upgraded installation

Boy132: > Server subuser emailing seems to work just fine even with it removed. image

Okay, figured it out. This is the "Account created" email but you didn't get the "Added to server" email, right?
That's the email that is missing.

Boy132 (changes requested):
Proper fix is to remove the `subscribe` for the `AuthenticationListener`, [here](https://github.com/pelican-dev/panel/blob/main/app/Providers/EventServiceProvider.php#L27-L29) and [here](https://github.com/pelican-dev/panel/blob/main/app/Listeners/Auth/AuthenticationListener.php#L32-L36).

Could you make these changes and revert the removal of `EventServiceProvider`?

ItzExotical:
Removing unused imports.

Boy132 (approved):
LGTM

RMartinOscar (commented):
LGTM

lancepioch (approved):

PR #468, opened 2024-07-08
Fix default timestamp for activity logs

Boy132:
The `timestamp` column for activity logs [uses `useCurrent`](https://github.com/pelican-dev/panel/blob/main/database/migrations/2022_05_28_135717_create_activity_logs_table.php#L22) which sets the default column value to `CURRENT_TIMESTAMP`. But this uses the database timezone.
This causes the timezone to be "added twice" on the frontend.

Example with `Europe/Berlin`:
grafik
grafik

This PR removes the default and sets the timestamp to `Carbon::now()` when the model is created.

### DIFF
--- a/app/Models/ActivityLog.php
+++ b/app/Models/ActivityLog.php
@@ -141,7 +141,7 @@ class ActivityLog extends Model
         parent::boot();
 
         static::creating(function (self $model) {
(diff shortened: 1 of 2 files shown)


ItzExotical (commented):
Tested. Code works without issues.
image

lancepioch (approved):

Issue #473, opened 2024-07-08
Panel does not correctly handle X-Forwarded-For


ItzExotical:
### Current Behavior

The panel does not care about the X-Forwarded-For header provided by Caddy among other web servers, resulting in the panel getting the user's IP wrong (if a service like Cloudflare is used).

### Expected Behavior

The panel should get the user's actual IP and store that in logs too.

### Steps to Reproduce

Sorry, but I think this one is quite straightforward.

### Panel Version

1.0.0-beta6

### Wings Version

1.0.0-beta3

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

Boy132:
You need to set the `TRUSTED_PROXIES` in your .env file. Then it'll show the real ip.
See https://pelican.dev/docs/panel/config#reverse-proxy-setup

ItzExotical:
Oh well, I totally missed that. What about wings? Still shows proxy IP for server events, such as opening and writing to files.

Boy132:
Wings also has a config option for that. Don't remember the name tho but it's somewhere in the `config.yml`.

Boy132:
I'm going to close this. If you still have problems with it showing the wrong ip you should visit the Discord and open a support thread. But setting the trusted proxies config vars should work.

ItzExotical:
I set `TRUSTED_PROXIES` to `127.0.0.1` in my `.env` and it still showed the Cloudflare IP and not my IP. I even tried setting it to wildcard but same thing there. X-Forwarded-For exists and is properly set to the real connecting IP of the client, but Pelican is not handling it correctly it looks like.

Boy132:
Because `127.0.0.1` is wrong, you need to set it to the cloudflare ips as described in the docs. At this point please visit the Discord for support.

PR #474, opened 2024-07-08
Correctly handle X-Forwarded-For and retrieve the real IP (#473)

ItzExotical:
This will correctly show the user's IP if the web server is configured to do so properly and is behind a service such as Cloudflare. Tested with Caddy built with [this](https://caddyserver.com/docs/modules/http.ip_sources.cloudflare) module. Works on my instance, but would love further testing by someone. The same should be done for wings, but I am not good at Go so someone else gets that task.

### DIFF
--- a/app/Http/Middleware/ProxyHandler.php
+++ b/app/Http/Middleware/ProxyHandler.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Http\Middleware;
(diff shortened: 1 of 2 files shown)


Boy132:
See https://github.com/pelican-dev/panel/issues/473#issuecomment-2216692933

PR #483, opened 2024-07-15
Add timeouts

### LINKED ISSUE #475: Panel Dashboard doesn't load without internet access because of the Github repository version checks

notAreYouScared:
Add Timeouts to github call.

Should close #475

### DIFF
--- a/app/Services/Helpers/SoftwareVersionService.php
+++ b/app/Services/Helpers/SoftwareVersionService.php
@@ -89,13 +89,23 @@ class SoftwareVersionService
             $versionData = [];
 
             try {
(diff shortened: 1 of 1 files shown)


Boy132 (commented):
We should probably use the config values here too: https://github.com/pelican-dev/panel/blob/main/config/panel.php#L65-L68

Or at least set the timeout a bit higher, 3 seconds seems very low.

notAreYouScared:
> We should probably use the config values here too: https://github.com/pelican-dev/panel/blob/main/config/panel.php#L65-L68
> 
> Or at least set the timeout a bit higher, 3 seconds seems very low.

Im fine with using the existing config option, didn't know we already had that.

Boy132 (approved):
LGTM

PR #486, opened 2024-07-16
Settings page

Boy132:
Adds a simple Settings page. Settings are directly written to the .env file.
The old admin settings stuff is removed.

This supersedes #259.

grafik
grafik
grafik
grafik
grafik


Poseidon281:
You are kidding me right. After I and with the help of Lance spend all this time working on my PR. You are just going to make another?

app/Filament/Pages/Settings.php
```diff
@@ -0,0 +1,342 @@
...
+                                        ->live()
+                                        ->columnSpanFull()
+                                        ->formatStateUsing(fn ($state): bool => (bool) $state)
+                                        ->afterStateUpdated(fn ($state, Set $set) => $set('PANEL_CLIENT_ALLOCATIONS_ENABLED', (bool) $state))
+                                        ->default(env('PANEL_CLIENT_ALLOCATIONS_ENABLED', config('panel.client_features.allocations.enabled'))),
+                                    TextInput::make('PANEL_CLIENT_ALLOCATIONS_RANGE_START')
+                                        ->label('Starting Port')
+                                        ->required(true)
+                                        ->numeric()
+                                        ->minValue(1024)
```
notAreYouScared, line 236: i think this has to be 1025, as the ports have to be > 1024? i might be wrong

Boy132: No, ports below 1024 are restricted. So 1 - 1023 isn't allowed.

notAreYouScared: Correct. Greater than 1024. So the min would be 1025?

parkervcp: Below 1024 is 1-1023. The usable range is 1024 and above.

RMartinOscar: Ports in an allocation must be greater than 1024 and less than or equal to 65535.
image

RMartinOscar:
Why fully removing the `APP_ENVIRONMENT_ONLY` instead of setting the default to false ?
I find it very useful for troubleshooting

Boy132:
> Why fully removing the `APP_ENVIRONMENT_ONLY` instead of setting the default to false ? I find it very useful for troubleshooting

Because it's no longer doing anything. Settings are directly written to the `.env` file.

RMartinOscar:
> > Why fully removing the `APP_ENVIRONMENT_ONLY` instead of setting the default to false ? I find it very useful for troubleshooting
> 
> Because it's no longer doing anything. Settings are directly written to the `.env` file.

Mybad thought it was still saved in the settings table of the db

app/Filament/Pages/Settings.php
Boy132: > Ports in an allocation must be greater than 1024 and less than or equal to 65535. image

Then we need to update that bit. It's definitely _below_ 1024, so 1024 itself is valid.

notAreYouScared:
I feel this is a good start, overtime i can add everything we have an env value for.

notAreYouScared:
I feel this is good enough in its current state to be added, More settings can be added along the way.

notAreYouScared (approved):

PR #492, opened 2024-07-17
Fix update egg from url

Boy132:
https://github.com/pelican-dev/panel/commit/160ea1ed50231db4719f604d46904b0c4ddfcd14 changed the field name to `update_url` which broke the updating.

### DIFF
--- a/app/Filament/Resources/EggResource/Pages/EditEgg.php
+++ b/app/Filament/Resources/EggResource/Pages/EditEgg.php
@@ -249,9 +249,9 @@ class EditEgg extends EditRecord
                             Tab::make('From URL')
                                 ->icon('tabler-world-upload')
                                 ->schema([
-                                    TextInput::make('update_url')
+                                    TextInput::make('url')
                                         ->label('URL')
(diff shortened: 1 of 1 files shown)


notAreYouScared:
I'm confused what this fixes? It's called `update_url` in the json file and the database.

Boy132:
> I'm confused what this fixes? It's called `update_url` in the json file and the database.
> 
> https://github.com/pelican-dev/panel/blob/833ae30e593ef32032103ac47b0b4998a2339386/app/Services/Eggs/Sharing/EggExporterService.php#L23

It checks if `url` is present.
And the url doesn't have to be the eggs "official" update url. So it's named differently to prevent accidental overwriting.

https://github.com/pelican-dev/panel/blob/main/app/Filament/Resources/EggResource/Pages/EditEgg.php#L282-L284

notAreYouScared:
Ohhh you're talking about the end, Sorry my quick look thought you were changing it on the editnode page, not the import screen,

notAreYouScared (approved):

PR #495, opened 2024-07-18
Command to cleanup docker images

### LINKED ISSUE #250: Automatic cleanup of dangling docker images induced by updated egg containers
### Discussed in https://github.com/pelican-dev/panel/discussions/54

<sup>Originally posted by **DerLev** April 13, 2024</sup>
I have had some issues in the past with dangling docker images filling up my storage on my wings host. An automatic task for cleaning up those dangling images would be nice. A current workaround is to set a cron job, which works fine but if this was to be included in the daemon itself or just as a simple note in the docs it would greatly help new users setting things up</div>

Boy132:
Closes #250

Adds a command to cleanup dangling docker images. This command is run daily.

### DIFF
--- a/app/Console/Commands/Maintenance/PruneImagesCommand.php
+++ b/app/Console/Commands/Maintenance/PruneImagesCommand.php
@@ -18,7 +18,7 @@ class PruneImagesCommand extends Command
         $node = $this->argument('node');
 
         if (empty($node)) {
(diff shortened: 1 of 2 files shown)


notAreYouScared (approved):
LGTM, Ran locally, and cleared out some dangling images.

PR #496, opened 2024-07-18
feat: Webhooks

### LINKED ISSUE #9: Implement Webhooks
This includes an event system that records events to the database. This may replace the activity log.

Poseidon281:
closes #9 

#### TODO

| Task | Status |
|--------|--------|
| Add nodes | Done |
| Add DatabaseHosts  | Done  |


database/migrations/2024_07_18_110844_create_filament-webhook-server_table.php
notAreYouScared, line 1: Should have a down option that drops the table.

Poseidon281: Sorry overlooked that. Added it now

notAreYouScared:
One or more of the packages used brings the panel to its knees. Please re-evaluate
image
image

Poseidon281:
> One or more of the packages used brings the panel to its knees. Please re-evaluate 

I could not fix the plugin so i am making it myself. Not that this PR is inactive, but it could take some time

notAreYouScared:
Okay. Thanks for the update.

Poseidon281:
> Okay. Thanks for the update.

Do you guys want it to be a normal webhook or a Discord webhook?

app/Filament/Pages/Settings.php
```diff
@@ -542,8 +618,6 @@ public function save(): void
 
             $this->rememberData();
 
-            $this->redirect($this->getUrl());
```
Poseidon281, line 545: I dont see the need for this redirect.

notAreYouScared: I think it was added to update the values on the settings page after you saved them? I could be wrong. @Boy132

Boy132: Yes, that's intentional. It's to refresh the page.

Poseidon281: But why? For what i could see the page does not need to refresh and it is rather annoying, because it does not redirect to the tab that you are in, but to the general tab.

Boy132: It depends on the settings you change. The refresh is to instantly show the changes.

notAreYouScared:
Thank-you for the effort and the passion for supporting the project, this is a little more hard coded then we'd like.

PR #499, opened 2024-07-19
Display both 12h & 24h format for Schedules

RMartinOscar:
image
image


notAreYouScared:
This should really be an option on the user's profile, just life timezone and language. Allow them to pick 12/24h for everything

notAreYouScared:
We can revisit this once the client side in closer to completion.

PR #500, opened 2024-07-19
mysql and mariadb client command for docker

Vooperino:
Upon building the dockerfile there is a slight issue then using these env var
```bash
DB_CONNECTION=mariadb
or
DB_CONNECTION=mysql
```

will result a failed command the container or the app will fail to start up, I have created a quick fix that add these command as client only


Boy132:
See https://github.com/pelican-dev/panel/pull/313 - the docker file is being reworked.

PR #502, opened 2024-07-22
Admin Roles

Boy132:
Uses [`spatie/laravel-permission`](https://github.com/spatie/laravel-permission) to add roles and permissions (for admins). This does *not* change subuser permissions! (#193)

A role named "Root Admin" is created during seeding. This role has all permissions and will replace the old `root_admin` bool.

**Example:**
grafik
grafik


dbcaa:
ewwww light modeee im jkk looks good

notAreYouScared:
Looks good, Seems like we will have a lot of updating to do when we add this. 

Things like "restore" & "force delete" aren't useful for us.

alexevladgabriel:
We can still use root_admin, let's say we have a case where we have 2-3 administrators and one of them decides the invited user chooses to remove the original administrator and grief the entire panel, what do we do in that case?

Boy132:
> We can still use root_admin, let's say we have a case where we have 2-3 administrators and one of them decides the invited user chooses to remove the original administrator and grief the entire panel, what do we do in that case?

The same you would now, create a new admin user via artisan cli.
Just having a duplicate and redundant `root_admin` wouldn't change that.

notAreYouScared (approved):
From what i've tested, all seems well.

alexevladgabriel (approved):
LGTM

database/Seeders/DatabaseSeeder.php
```diff
@@ -12,5 +13,7 @@ class DatabaseSeeder extends Seeder
     public function run()
     {
         $this->call(EggSeeder::class);
+
+        Role::firstOrCreate(['name' => Role::ROOT_ADMIN]);
```
lancepioch, line 17: This needs to be in the migration file itself.

config/permission.php
```diff
@@ -0,0 +1,186 @@
+<?php
```
lancepioch, line 1: Did you change any of this file? If not, then we don't need to commit it. If you did, then we only need to add the missing array keys.

app/Filament/Resources/RoleResource.php
```diff
@@ -0,0 +1,178 @@
...
+        'User',
+    ];
+
+    private const PERMISSION_PREFIXES = [
+        'viewList',
+        'view',
+        'create',
+        'update',
+        'delete',
+    ];
```
lancepioch, line 51: Can you move this into new enums

app/Filament/Resources/RoleResource.php
```diff
@@ -0,0 +1,178 @@
...
+            'export',
+        ],
+    ];
+
+    private const SPECIAL_PERMISSIONS = [
+        'Settings' => [
+            'view',
+            'update',
+        ],
+    ];
```
lancepioch, line 65: Put this in the new model, assuming you can't fit it into an enum, I'd say

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+        $columnNames = config('permission.column_names');
+        $pivotRole = $columnNames['role_pivot_key'] ?? 'role_id';
+        $pivotPermission = $columnNames['permission_pivot_key'] ?? 'permission_id';
+
+        if (empty($tableNames)) {
+            throw new \Exception('Error: config/permission.php not loaded. Run [php artisan config:clear] and try again.');
+        }
+        if ($teams && empty($columnNames['team_foreign_key'] ?? null)) {
+            throw new \Exception('Error: team_foreign_key on config/permission.php not loaded. Run [php artisan config:clear] and try again.');
+        }
```
lancepioch, line 25: I'm guessing you ran into these errors yourself? Do you know why they happened?

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+
+        Schema::create($tableNames['permissions'], function (Blueprint $table) {
+            //$table->engine('InnoDB');
```
lancepioch, line 28:
```suggestion
```

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+            $table->bigIncrements('id'); // permission id
+            $table->string('name');       // For MyISAM use string('name', 225); // (or 166 for InnoDB with Redundant/Compact row format)
+            $table->string('guard_name'); // For MyISAM use string('guard_name', 25);
+            $table->timestamps();
+
+            $table->unique(['name', 'guard_name']);
+        });
+
+        Schema::create($tableNames['roles'], function (Blueprint $table) use ($teams, $columnNames) {
+            //$table->engine('InnoDB');
```
lancepioch, line 38:
```suggestion
```

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+                $table->index($columnNames['team_foreign_key'], 'roles_team_foreign_key_index');
+            }
+            $table->string('name');       // For MyISAM use string('name', 225); // (or 166 for InnoDB with Redundant/Compact row format)
+            $table->string('guard_name'); // For MyISAM use string('guard_name', 25);
+            $table->timestamps();
+            if ($teams || config('permission.testing')) {
+                $table->unique([$columnNames['team_foreign_key'], 'name', 'guard_name']);
+            } else {
+                $table->unique(['name', 'guard_name']);
+            }
```
lancepioch, line 51: I definitely don't like this being split up like this.

How can we make this a single statement that works with all the databases?

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+
+            $table->string('model_type');
+            $table->unsignedBigInteger($columnNames['model_morph_key']);
+            $table->index([$columnNames['model_morph_key'], 'model_type'], 'model_has_permissions_model_id_model_type_index');
+
+            $table->foreign($pivotPermission)
+                ->references('id') // permission id
+                ->on($tableNames['permissions'])
+                ->onDelete('cascade');
+            if ($teams) {
```
lancepioch, line 65: $teams?

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+
+            $table->string('model_type');
+            $table->unsignedBigInteger($columnNames['model_morph_key']);
+            $table->index([$columnNames['model_morph_key'], 'model_type'], 'model_has_roles_model_id_model_type_index');
+
+            $table->foreign($pivotRole)
+                ->references('id') // role id
+                ->on($tableNames['roles'])
+                ->onDelete('cascade');
+            if ($teams) {
```
lancepioch, line 89: $teams?

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+            $table->foreign($pivotRole)
+                ->references('id') // role id
+                ->on($tableNames['roles'])
+                ->onDelete('cascade');
+
+            $table->primary([$pivotPermission, $pivotRole], 'role_has_permissions_permission_id_role_id_primary');
+        });
+
+        app('cache')
+            ->store(config('permission.cache.store') != 'default' ? config('permission.cache.store') : null)
```
lancepioch, line 119:
```suggestion
            ->store(config('permission.cache.store') !== 'default' ? config('permission.cache.store') : null)
```

database/migrations/2024_08_01_114538_remove_root_admin_column.php
```diff
@@ -0,0 +1,35 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        $adminUsers = User::whereRootAdmin(1)->get();
```
lancepioch, line 16:
```suggestion
        $adminUsers = User::whereRootAdmin(true)->get();
```

tests/Traits/Http/RequestMockHelpers.php
```diff
@@ -35,13 +35,14 @@ public function setRequestUserModel(User $user = null): void
...
      * Generates a new request user model and also returns the generated model.
      */
-    public function generateRequestUserModel(array $args = []): User
+    public function generateRequestUserModel(bool $isRootAdmin, array $args = []): void
     {
-        /** @var \App\Models\User $user */
         $user = User::factory()->make($args);
-        $this->setRequestUserModel($user);
+        $user = m::mock($user)->makePartial();
+        $user->shouldReceive('isRootAdmin')->andReturn($isRootAdmin);
```
lancepioch, line 42: I don't like mocking and having the assertion this every single time this is called, I'd rather just generate the permission/role for real here.

tests/Unit/Http/Middleware/AdminAuthenticateTest.php
```diff
@@ -13,7 +14,9 @@ class AdminAuthenticateTest extends MiddlewareTestCase
      */
     public function testAdminsAreAuthenticated(): void
     {
-        $user = User::factory()->make(['root_admin' => 1]);
+        $user = User::factory()->make();
+        $user = Mockery::mock($user)->makePartial();
+        $user->shouldReceive('isRootAdmin')->andReturnTrue();
```
lancepioch, line 19: Create the actual role here too.

database/migrations/2024_07_19_130942_create_permission_tables.php
```diff
@@ -0,0 +1,140 @@
...
+                $table->unique(['name', 'guard_name']);
+            }
+        });
+
+        Schema::create($tableNames['model_has_permissions'], function (Blueprint $table) use ($tableNames, $columnNames, $pivotPermission, $teams) {
+            $table->unsignedBigInteger($pivotPermission);
+
+            $table->string('model_type');
+            $table->unsignedBigInteger($columnNames['model_morph_key']);
+            $table->index([$columnNames['model_morph_key'], 'model_type'], 'model_has_permissions_model_id_model_type_index');
```
lancepioch, line 59: This suggestion is optional, but I think we'd probably be fine with the auto generated index names?

lancepioch (changes requested):

database/migrations/2024_07_19_130942_create_permission_tables.php
notAreYouScared: This migration is bundled with the plugin, nothing was changed in it.

database/Seeders/DatabaseSeeder.php
Boy132: Why? That's exactly what seeders are for?

config/permission.php
Boy132: Only thing that changed is the Role class path.

tests/Traits/Http/RequestMockHelpers.php
Boy132: I had to do mocking because we run tests without migrations.

database/Seeders/DatabaseSeeder.php
lancepioch: I meant in addition to, that way it gets added to everybody's. If that's already happening, then you can ignore this comment, thanks.

config/permission.php
lancepioch: Okay, so only keep that single key in this file so it overwrites it (Laravel 11).

```php
<?php

return [
    'models' => [
        'role' => \App\Models\Role::class,
    ],
];
```

So this can be the whole file.

tests/Traits/Http/RequestMockHelpers.php
lancepioch: Ah, forgot about that, then this is fine, thanks

config/permission.php
Boy132: Already done. :)

database/Seeders/DatabaseSeeder.php
Boy132: If the root admin role is missing it's created automatically if needed. The seeding is only for new installs, really. So that should be fine.

lancepioch (approved):

PR #504, opened 2024-07-23
Web Installer

### LINKED ISSUE #5: Show first time admin user registration if no users exist in the database
### Donate 💰 to fund this issue

- You can donate funding to this issue. We receive the money once the issue is completed & confirmed by you.
- The funding will be distributed between the team and contributors to help all aspects of the project.
- Thank you in advance for helping prioritize & fund our progress!

<a href="https://polar.sh/pelican-dev/panel/issues/5">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/pelican-dev/panel/issues/5/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Boy132:
Adds a web installer (app url + `/installer`) with 4 steps (5 if redis is selected):
grafik
grafik
grafik
grafik
grafik

It also handles running migrations/ seeding and setting up the queue worker service file. The app settings command (`php artisan p:environment:setup`) has been simplified to now only copy the example .env and generate an app key.

The installer is not enforced, there is no automatic redirect if the panel isn't "installed".

Also closes #5.


RMartinOscar:
Can’t proceed with installation if both sqlite3 and mysql extension are not installed even though only one will be used in the end.

fpm is not in the extension_loaded so it will always fail and how can you get to this page without it anyway ?

I believe we should keep the CLI in case we want to migrate db/cache without editing .env manually or reinstalling

RMartinOscar:
I also wonder how would one proceed with a migration from ptero with this in place without a guide telling you about the installed file

RMartinOscar:
> > Can’t proceed with installation if both sqlite3 and mysql extensions are not installed even though only one will be used
> Yeah but there is no nice way to handle that. So just require both. This also makes switching easier.

Could remove them both from composer.json then make the requirements step skip checking for sqlite3 if mysql is installed (and the other way around)
I’v managed to do that by changing the acceptedColor and forcing it to true.

notAreYouScared (approved):
Worked for me on a fresh install <3

alexevladgabriel (approved):

PR #510, opened 2024-07-28
Added NavigationItem to and from admin

Poseidon281:
I also fixed pint in this PR


lancepioch:
This PR branch needs to be cleaned up, can't contain other commits

PR #528, opened 2024-08-06
Patch Env CLI

RMartinOscar:
- [x] Remove `--use-redis` following [this commit](https://github.com/pelican-dev/panel/commit/953ee940aa5955dfdad452dbdc46f48ca5e5bb55#diff-ee666c86898d686c606a20796d57bbb363ead4dd6dfa14bbd1a6764d513b5edfL17)
- [x] Add Redis Username (default if empty)
- [x] Move confirm prompt to lang for reuse

### DIFF
--- a/app/Traits/Commands/RequestRedisSettingsTrait.php
+++ b/app/Traits/Commands/RequestRedisSettingsTrait.php
@@ -17,13 +17,12 @@ trait RequestRedisSettingsTrait
 
         if (!empty(config('database.redis.default.user'))) {
             $this->variables['REDIS_USERNAME'] = config('database.redis.default.user');
-            $askForRedisUser = $this->confirm('It seems a user is already defined for Redis, would you like to change it?');
+            $askForRedisUser = $this->confirm(__('commands.appsettings.redis.confirm', ['field' => 'user']));
         }
         if (!empty(config('database.redis.default.password'))) {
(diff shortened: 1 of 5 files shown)


app/Traits/Commands/RequestRedisSettingsTrait.php
```diff
@@ -12,19 +12,34 @@ protected function requestRedisSettings(): void
...
+        }
         if ($askForRedisPassword) {
             $this->output->comment(__('commands.appsettings.redis.comment'));
             $this->variables['REDIS_PASSWORD'] = $this->option('redis-pass') ?? $this->output->askHidden(
                 'Redis Password'
             );
         }
 
+        if (empty($this->variables['REDIS_USERNAME'])) {
+            $this->variables['REDIS_USERNAME'] = 'default';
```
Boy132, line 41:
```suggestion
            $this->variables['REDIS_USERNAME'] = 'null';
```

Boy132 (changes requested):
Looks good, but the default value for the redis user should be empty/ null.

app/Traits/Commands/RequestRedisSettingsTrait.php
RMartinOscar: Tried that and it didn't work i had to set it to 'default'

Boy132 (approved):
LGTM

PR #537, opened 2024-08-09
Patch file explorer in pelican-dev/panel/issue/353

RMartinOscar:
Add is_ in front of bools so `(bool) self::$is_directory` and `(string) self::$directory` are different

Change `File::$file File::$directory File::$symlink` to `File::$is_file File::$is_directory File::$is_symlink`

PHPStan is still made cause of [Sushi](https://github.com/calebporzio/sushi)


Boy132:
Thanks, but this was only the start of the file explorer.
It will be changed/ expanded.

PR #540, opened 2024-08-11
Update default image for new eggs

Boy132:


RMartinOscar:
Why not use `ghcr.io/pelican-eggs/installers:alpine` instead ? Alpine is way more lightweight then debian

parkervcp:
> Why not use `ghcr.io/pelican-eggs/installers:alpine` instead ? Alpine is way more lightweight then debian

I would say use debian over alpine because alpine/ash/muslc do not behave the same was bash/glibc do and most people won't understand why a script they wrote doesn't work like it does when they run it on their own.

notAreYouScared (approved):

PR #543, opened 2024-08-14
Add authentik oauth support

EkiciLP:
related to #385


EkiciLP:

EkiciLP:
related to #385

RMartinOscar:
https://github.com/pelican-dev/panel/pull/435#issuecomment-2194608450
>Thanks but we'll add more providers once the oauth frontend is done. It will also be less hardcoded.

Boy132:
Thank you for your contribution but the way how providers are added will be changed once we work on the oauth front end.
We will also add more default providers then.

PR #544, opened 2024-08-14
remove 'guest' middleware to enable oauth account linking

EkiciLP:
The ``guest`` middleware did not allow for authenticated users to call the /auth/oauth/redirect endpoint resulting in not being able to link an oauth account to yours.


Boy132:
Those are not the endpoints for linking but for logging in.
Linking is https://github.com/pelican-dev/panel/blob/main/routes/base.php#L13-L14

For authenticated users it doesn't make sense to login again with oauth.

EkiciLP:
Yes but the link endpoint currently just redirects to the oauth/redirect endpoint which would fail.

Boy132:
It doesn't. The link endpoint redirects to the oauth provider page which then redirects back to the callback endpoint.
The redirect endpoint is only used for login.

PR #546, opened 2024-08-14
Oauth basic user provisioning

EkiciLP:
This pr aims to add a temporary solution for people wanting to use oauth as the main user backend untel the frontend is done.

- Environment variable ``OAUTH_USER_PROVISIONING=(bool)`` turns this on or off


lancepioch:
Sorry, not looking for a temporary solution here.

PR #549, opened 2024-08-14
Add missing foreign keys to fix #545

### LINKED ISSUE #545: Egg variable name needs to be unique per egg
Currently you can create multiple variables with the same name. This should not be possible.
Same applies to the env variable name.

RMartinOscar:
- [x] Make name & env_variables unique for each egg

> Throws a 500 if you try to add a duplicate

### DIFF
--- a/database/migrations/2024_08_14_182737_fix_egg_variable_foreign_key.php
+++ b/database/migrations/2024_08_14_182737_fix_egg_variable_foreign_key.php
@@ -1,0 +1,57 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Only needed for sqlite
+        if (Schema::getConnection()->getDriverName() !== 'sqlite') {
+            return;
+        }
+
+        // Disable foreign checks
+        // legacy_alter_table needs to be 'ON' so existing foreign key table references aren't renamed when renaming the table, see https://www.sqlite.org/lang_altertable.html
+        DB::statement('PRAGMA foreign_keys = OFF');
+        DB::statement('PRAGMA legacy_alter_table = ON');
+
+        DB::transaction(function () {
+            DB::statement('ALTER TABLE egg_variables RENAME TO _egg_variables_old');
+            DB::statement('CREATE TABLE egg_variables 
+                ("id" integer primary key autoincrement not null, 
+                "egg_id" integer not null, 
+                "name" varchar not null, 
+                "description" text not null, 
+                "env_variable" varchar not null, 
+                "default_value" text not null, 
+                "user_viewable" integer not null, 
+                "user_editable" integer not null, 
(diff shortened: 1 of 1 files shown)


database/migrations/2024_08_14_182737_fix_egg_variable_foreign_key.php
```diff
@@ -0,0 +1,57 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Only needed for sqlite
+        if (Schema::getConnection()->getDriverName() !== 'sqlite') {
```
Boy132, line 15: Fix needs to apply to all drivers, not just sqlite

database/migrations/2024_08_14_182737_fix_egg_variable_foreign_key.php
```diff
@@ -0,0 +1,57 @@
...
+                "rules" text not null, 
+                "created_at" datetime, 
+                "updated_at" datetime, 
+                "sort" integer, 
+                foreign key("egg_id") references "eggs"("id") on delete cascade)');
+            DB::statement('INSERT INTO egg_variables SELECT * FROM _egg_variables_old');
+            DB::statement('DROP TABLE _egg_variables_old');
+            DB::statement('CREATE UNIQUE INDEX "egg_variables_env_variable_unique" on "egg_variables" ("egg_id", "env_variable")');
+            DB::statement('CREATE UNIQUE INDEX "egg_variables_name_unique" on "egg_variables" ("egg_id", "name")');
+        });
```
Boy132, line 44: You don't need to drop and recreate the table to create an index. That is only needed when changing foreign keys on sqlite.

Create a new migration with something like this:
```php
Schema::table('egg_variables', function (Blueprint $table) {
    $table->unique(['egg_id', 'env_variable']);
    $table->unique(['egg_id', 'name']);
});
```

database/migrations/2024_08_14_182737_fix_egg_variable_foreign_key.php
```diff
@@ -0,0 +1,57 @@
...
+        DB::statement('PRAGMA foreign_keys = ON');
+        DB::statement('PRAGMA legacy_alter_table = OFF');
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        // Reverse not needed
```
Boy132, line 55:
```php
Schema::table('egg_variables', function (Blueprint $table) {
    $table->dropUnique(['egg_id', 'env_variable']);
    $table->dropUnique(['egg_id', 'name']);
});
```

Boy132 (changes requested):
Besides the backend changes we should also handle the frontend properly.

RMartinOscar:
@Boy132 sqlite tests are still failling

Boy132:
> @Boy132 sqlite tests are still failling

I found the reason: the `sort` for the bungeecord variables is `null`. For all other eggs the `sort` is filled.

notAreYouScared:
@RMartinOscar Can you merge main, and see if the tests pass?

lancepioch:
We should programmatically enforce this instead, I think sqlite has an issue with composite unique keys.

PR #576, opened 2024-09-15
Update placeholders & panel error command

ash-development:
The version placeholders are easier to remove rather than maintain. Also the panel error upload command was incorrect, so I fixed that.


ash-development:

PR #580, opened 2024-09-24
Add artisan command for setting up redis

Boy132:
Instead of having to run 3 commands (with the same input multiple times) you can simply run this one command.

### DIFF
--- a/app/Console/Commands/Environment/RedisSetupCommand.php
+++ b/app/Console/Commands/Environment/RedisSetupCommand.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace App\Console\Commands\Environment;
+
+use App\Traits\Commands\RequestRedisSettingsTrait;
+use App\Traits\EnvironmentWriterTrait;
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Console\Kernel;
+
+class RedisSetupCommand extends Command
+{
+    use EnvironmentWriterTrait;
+    use RequestRedisSettingsTrait;
+
+    protected $description = 'Configure the Panel to use Redis as cache, queue and session driver.';
+
+    protected $signature = 'p:redis:setup
+                            {--redis-host= : Redis host to use for connections.}
+                            {--redis-user= : User used to connect to redis.}
+                            {--redis-pass= : Password used to connect to redis.}
+                            {--redis-port= : Port to connect to redis over.}';
+
+    protected array $variables = [];
+
+    /**
+     * RedisSetupCommand constructor.
+     */
+    public function __construct(private Kernel $console)
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Handle command execution.
+     */
+    public function handle(): int
+    {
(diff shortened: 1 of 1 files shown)


RMartinOscar:
Shouldn't we remove those then as we can only have one redis cache instance ?
https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/CacheSettingsCommand.php#L25-L28
https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/QueueSettingsCommand.php#L25-L28
https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/SessionSettingsCommand.php#L26-L29

Boy132:
> Shouldn't we remove those then as we can only have one redis cache instance ?
> 
> https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/CacheSettingsCommand.php#L25-L28
> 
> 
> https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/QueueSettingsCommand.php#L25-L28
> 
> 
> https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/SessionSettingsCommand.php#L26-L29

No, you still should be able to run the other commands, e.g. if you only want to use redis for one thing.

RMartinOscar:
> No, you still should be able to run the other commands, e.g. if you only want to use redis for one thing.

Can't you just change --driver ?
https://github.com/pelican-dev/panel/blob/fc643f57f9c2d862fc0d3644ca98945a1afcd599/app/Console/Commands/Environment/CacheSettingsCommand.php#L24

RMartinOscar:
> No, you still should be able to run the other commands, e.g. if you only want to use redis for one thing.
You first do `php artisan p:redis:setup` then you set Cache, Queue and/or Session so you just have to specify the driver not redis user,pass,host for each of them

So you can have Cache and Queue as redis and Session as file/db for some reason.
But you don't have to enter --redis-user --redis-pass --redis-host every single time you want to set one to use redis

Boy132:
> > No, you still should be able to run the other commands, e.g. if you only want to use redis for one thing.
> > You first do `php artisan p:redis:setup` then you set Cache, Queue and/or Session so you just have to specify the driver not redis user,pass,host for each of them
> 
> So you can have Cache and Queue as redis and Session as file/db for some reason. But you don't have to enter --redis-user --redis-pass --redis-host every single time you want to set one to use redis

In that case you can either run the new command and change the session driver later back or use only the cache & queue driver commands to change it to redis. That's the point of this new command - to simply the process of setting multiple drivers to redis at once.

lancepioch (approved):

PR #595, opened 2024-09-29
Patch User check & Auto Login in the Installer

RMartinOscar:


lancepioch (approved):

lancepioch:
Being replaced with #614

PR #599, opened 2024-10-02
Add optimize filament to docker

lancepioch:


RMartinOscar:
Should we also add to The installer ?
`Artisan::call('filament:optimize');`

https://github.com/pelican-dev/panel/blob/6c205a744d01ee4eebe21f74244de9181d03a9be/app/Filament/Pages/Installer/PanelInstaller.php#L117-L118
and `Artisan::call('filament:optimize-clear');`
https://github.com/pelican-dev/panel/blob/6c205a744d01ee4eebe21f74244de9181d03a9be/app/Filament/Pages/Installer/PanelInstaller.php#L67-L71

lancepioch:
No because it should be ran each time PHP is restarted instead of just the once.

RMartinOscar:
> No because it should be ran each time PHP is restarted instead of just the once.

So like a cron for the standalone install ?

Boy132:
> > No because it should be ran each time PHP is restarted instead of just the once.
> 
> Could we take advantage of `* * * * * php /var/www/pelican/artisan schedule:run >> /dev/null **2>&1` cron and create a custom schedule ?

It's only needed for docker. For the non-dockerized install you only need to run it once. (and we plan on adding it to the release build workflow)

PR #601, opened 2024-10-02
Add can check to fix #600

### LINKED ISSUE #600: Admin Role Users Without Permission Can Change Primary Allocation
### Current Behavior

Users given an admin role that only allows servers to be viewed are still able to change the primary allocation even without permission.

Extra info:  the user itself is also not able to add, remove, or modify current allocations on the server either, only view.

### Expected Behavior

The user should not be able to change the default allocation without the permission to change allocations on a server.

### Steps to Reproduce

Create a user role in the admin panel and give it server view permissions, log in as said user and look at the servers page, the user is able to change the primary allocation even though they do not have the required permissions to do so, or if this is even intended.

### Panel Version

1.0.0-beta10

### Wings Version

1.0.0-beta5

Minecraft (Fabric)

### Docker Image

ghcr.io/parkervcp/yolks:java_21

### Error Logs

_No response_

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

RMartinOscar:
Fixes #600

### DIFF
--- a/app/Filament/Resources/ServerResource/Pages/ListServers.php
+++ b/app/Filament/Resources/ServerResource/Pages/ListServers.php
@@ -64,9 +64,8 @@ class ListServers extends ListRecords
                     ->searchable(),
                 SelectColumn::make('allocation_id')
                     ->label('Primary Allocation')
-                    ->options(fn (Server $server) => $server->allocations->mapWithKeys(
-                        fn ($allocation) => [$allocation->id => $allocation->address])
-                    )
+                    ->disabled(fn () => !auth()->user()->can('update server'))
+                    ->options(fn (Server $server) => $server->allocations->mapWithKeys(fn ($allocation) => [$allocation->id => $allocation->address]))
                     ->selectablePlaceholder(false)
                     ->sortable(),
                 TextColumn::make('image')->hidden(),
@@ -81,12 +80,7 @@ class ListServers extends ListRecords
                 Action::make('View')
                     ->icon('tabler-terminal')
                     ->url(fn (Server $server) => "/server/$server->uuid_short")
-                    ->visible(function (Server $server) {
-                        /** @var User $user */
-                        $user = auth()->user();
-
-                        return $user->isRootAdmin() || $user->id === $server->owner_id;
-                    }),
+                    ->visible(fn () => auth()->user()->can('view server')),
                 EditAction::make(),
             ])
             ->emptyStateIcon('tabler-brand-docker')
--- a/app/Filament/Resources/ServerResource/Pages/ListServers.php
+++ b/app/Filament/Resources/ServerResource/Pages/ListServers.php
@@ -67,7 +67,7 @@ class ListServers extends ListRecords
                     ->options(fn (Server $server) => $server->allocations->mapWithKeys(fn ($allocation) => [$allocation->id => $allocation->address]))
                     ->selectablePlaceholder(false)
                     ->sortable(),
-                TextColumn::make('allocation_id_can')
+                TextColumn::make('allocation_id_readonly')
                     ->label('Primary Allocation')
                     ->hidden(auth()->user()->can('update server'))
                     ->state(fn (Server $server) => $server->allocation->address),
--- a/app/Filament/Resources/ServerResource/Pages/ListServers.php
+++ b/app/Filament/Resources/ServerResource/Pages/ListServers.php
@@ -4,7 +4,6 @@ namespace App\Filament\Resources\ServerResource\Pages;
 
 use App\Filament\Resources\ServerResource;
 use App\Models\Server;
-use App\Models\User;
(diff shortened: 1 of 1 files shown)


app/Filament/Resources/ServerResource/Pages/ListServers.php
```diff
@@ -81,12 +80,7 @@ public function table(Table $table): Table
                 Action::make('View')
                     ->icon('tabler-terminal')
                     ->url(fn (Server $server) => "/server/$server->uuid_short")
-                    ->visible(function (Server $server) {
-                        /** @var User $user */
-                        $user = auth()->user();
-
-                        return $user->isRootAdmin() || $user->id === $server->owner_id;
-                    }),
+                    ->visible(fn () => auth()->user()->can('view server')),
```
Boy132, line 83:
```suggestion
                    ->authorize(fn () => auth()->user()->can('view server')),
```

app/Filament/Resources/ServerResource/Pages/ListServers.php
```diff
@@ -64,9 +64,8 @@ public function table(Table $table): Table
                     ->searchable(),
                 SelectColumn::make('allocation_id')
                     ->label('Primary Allocation')
-                    ->options(fn (Server $server) => $server->allocations->mapWithKeys(
-                        fn ($allocation) => [$allocation->id => $allocation->address])
-                    )
+                    ->disabled(fn () => !auth()->user()->can('update server'))
+                    ->options(fn (Server $server) => $server->allocations->mapWithKeys(fn ($allocation) => [$allocation->id => $allocation->address]))
```
Boy132, line 68: Instead of disabling the select itself it would probably look nicer to hide the column and show a normal TextColumn instead.

Boy132 (changes requested):

app/Filament/Resources/ServerResource/Pages/ListServers.php
RMartinOscar: Something like that ?
```php
auth()->user()->can('update server')
                    ? SelectColumn::make('allocation_id')
                        ->label('Primary Allocation')
                        ->options(fn (Server $server) => $server->allocations->mapWithKeys(fn ($allocation) => [$allocation->id => $allocation->address]))
                        ->sortable()
                    : TextColumn::make('allocation_id')
                        ->label('Primary Allocation')
                        ->formatStateUsing(fn (Server $server) => $server->allocation->address)
                        ->sortable(),
```

Boy132: > Something like that ?
> 
> ```
> auth()->user()->can('update server')
>                     ? SelectColumn::make('allocation_id')
>                         ->label('Primary Allocation')
>                         ->options(fn (Server $server) => $server->allocations->mapWithKeys(fn ($allocation) => [$allocation->id => $allocation->address]))
>                         ->selectablePlaceholder(false)
>                         ->sortable()
>                     : TextColumn::make('allocation_id')
>                         ->label('Primary Allocation')
>                         ->formatStateUsing(fn (Server $server) => $server->allocation->address)
>                         ->sortable(),
> ```

Yes but controlled with `hidden`/ `visible`.

app/Filament/Resources/ServerResource/Pages/ListServers.php
RMartinOscar: Just noticed the _View_ button leads to the react part which doesn't know what to do with Roles should i mke it aware of it or should we wait for the Filament frontend ?
Also rows in ListServers are not clickable if not rootAdmin

app/Filament/Resources/ServerResource/Pages/ListServers.php
RMartinOscar: I doubt we can have multiple fields with the same name just using `hidden` / `visible` which is why i went with ternary

Boy132: I should work since only of them is actually visible at a time. But let's try and see.

RMartinOscar: Well after testing i see that its procedural so the first one got overwritten.
With permission it shows TextColumn and without it just removes that column

Boy132: Rename the `TextColimn` to something like `allocation_id_readonly`. It's using `formatStateUsing` to overwrite the state anyways so it should be fine.

Boy132 (approved):

lancepioch:
Thank you @RMartinOscar - your changes are much appreciated, @Boy132 is handling this in his pr though: #614

PR #611, opened 2024-10-07
Update API for roles

Boy132:
- Makes sure you can't update/ delete/ assign the `Root Admin` role.
- Removes `guard_name` from role transformer.
- Adds `id` to role transformer.
- Adds endpoint to remove user roles.


lancepioch (approved):
Awesome, thanks!

PR #614, opened 2024-10-08
Update web installer

Boy132:
- Removed user check in installer
- Added back old .env variable for installer
- Simplified `p:environment:setup` command even more
- Fixes auto login
- The `APP_URL` is now auto filled
- The setup work is now done after each step (e.g. writing to .env, running migrations, creating the admin user) rather than at the very end
- Removed "unsaved changes" alert
- Improved error handling and reporting
- New "Setup complete" step that handles the queue and crontab stuff:
grafik
- Changed `.env.example` to set a `SESSION_COOKIE`. This way the page doesn't expire when changing the app name. [(By default, Laravel uses the `APP_NAME` to generate the name of the cookie)](https://github.com/laravel/framework/blob/11.x/config/session.php#L130-L133)

Supersedes #595


app/Filament/Pages/Installer/PanelInstaller.php
```diff
@@ -32,54 +34,43 @@ class PanelInstaller extends SimplePage implements HasForms
...
     {
         return MaxWidth::SevenExtraLarge;
     }
 
-    public static function show(): bool
+    public static function isInstalled(): bool
     {
-        if (User::count() <= 0) {
-            return true;
-        }
```
lancepioch, line 51: I definitely need this in there, it's a much better user experience.

app/Filament/Pages/Installer/PanelInstaller.php
```diff
@@ -32,54 +34,43 @@ class PanelInstaller extends SimplePage implements HasForms
...
-
-        if (config('panel.client_features.installer.enabled')) {
-            return true;
-        }
-
-        return false;
+        // This defaults to true so existing panels count as "installed"
+        return env('APP_INSTALLED', true);
```
lancepioch, line 53: Should be using config instead of env.

.env.example
```diff
@@ -4,6 +4,7 @@ APP_KEY=
 APP_TIMEZONE=UTC
 APP_URL=http://panel.test
 APP_LOCALE=en
+APP_INSTALLED=false
```
lancepioch, line 7: I really don't like this being necessary for a default app.

Anybody who copies this .env over will automatically have the installer enabled which might be a problem.

lancepioch (changes requested):

app/Filament/Pages/Installer/PanelInstaller.php
notAreYouScared: I disagree. This ONLY applies to docker... You're also forcing docker users to use SQLite in the beginning.

notAreYouScared (approved):

.env.example
notAreYouScared: We can revisit this. For now i'm going to merge it so the installer isn't broken for every new install.

app/Filament/Pages/Installer/PanelInstaller.php
notAreYouScared: We can revisit this. For now i'm going to merge it so the installer isn't broken for every new install.

PR #620, opened 2024-10-12
Auto-check eggs for update

Boy132:
Adds a new command to check installed eggs for updates. This just fetches the json from the `update_url` and compares it with the current json. (or rather the hashes of the json files) It runs automatically every hour.

If an update is found it will also display a button on the eggs list that will simply re-import the updated egg json.

_This will also mean that if you edit an egg it will detect it as "outdated"!_

### DIFF
--- a/app/Console/Commands/Egg/CheckEggUpdates.php
+++ b/app/Console/Commands/Egg/CheckEggUpdates.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace App\Console\Commands\Egg;
+
+use App\Models\Egg;
+use App\Services\Eggs\Sharing\EggExporterService;
+use Exception;
+use Illuminate\Console\Command;
+
+class CheckEggUpdates extends Command
+{
+    protected $signature = 'p:egg:check-updates';
+
+    public function handle(): void
+    {
+        /** @var EggExporterService $exporterService */
+        $exporterService = app(EggExporterService::class);
+
+        $eggs = Egg::all();
+        foreach ($eggs as $egg) {
+            try {
+                if (is_null($egg->update_url)) {
+                    $this->comment("{$egg->name}: Skipping (no update url set)");
+
+                    continue;
+                }
+
+                $currentJson = json_decode($exporterService->handle($egg->id));
+                unset($currentJson['exported_at']);
+
+                $updatedJson = json_decode(file_get_contents($egg->update_url));
+                unset($updatedJson['exported_at']);
+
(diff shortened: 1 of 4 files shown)


app/Console/Commands/Egg/CheckEggUpdates.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace App\Console\Commands\Egg;
+
+use App\Models\Egg;
+use App\Services\Eggs\Sharing\EggExporterService;
+use Exception;
+use Illuminate\Console\Command;
+
+class CheckEggUpdates extends Command
```
RMartinOscar, line 10:
```suggestion
class CheckEggUpdatesCommand extends Command
```

app/Console/Kernel.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Console;
 
+use App\Console\Commands\Egg\CheckEggUpdates;
```
RMartinOscar, line 5:
```suggestion
use App\Console\Commands\Egg\CheckEggUpdatesCommand;
```

app/Console/Kernel.php
```diff
@@ -35,6 +36,7 @@ protected function schedule(Schedule $schedule): void
 
         $schedule->command(CleanServiceBackupFilesCommand::class)->daily();
         $schedule->command(PruneImagesCommand::class)->daily();
+        $schedule->command(CheckEggUpdates::class)->hourly();
```
RMartinOscar, line 39:
```suggestion
        $schedule->command(CheckEggUpdatesCommand::class)->hourly();
```

RMartinOscar (commented):
`CheckEggUpdates.php` should be `CheckEggUpdatesCommand.php`

app/Console/Commands/Egg/CheckEggUpdates.php
```diff
@@ -0,0 +1,46 @@
...
+                if (is_null($egg->update_url)) {
+                    $this->comment("{$egg->name}: Skipping (no update url set)");
+
+                    continue;
+                }
+
+                $currentJson = json_decode($exporterService->handle($egg->id));
+                unset($currentJson->exported_at);
+
+                $updatedJson = json_decode(file_get_contents($egg->update_url));
```
RMartinOscar, line 31: Shouldn't use `file_get_contents`

Boy132: `file_get_contents` is fine, we also use it in the egg importer.

lancepioch (approved):

Issue #623, opened 2024-10-14
Broken Web Installer


RMartinOscar:
### Current Behavior

![image](https://github.com/user-attachments/assets/e954108a-ee44-43fc-9a37-57a814b6ef52)
The last step is displayed before it should.
`env_database.DB_DATABASE` default will always be `database.sqlite` eventhough `mariadb` or `sqlite` is selected.
The installer still writes `DB_HOST`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD` eventhough `sqlite` is selected.
The page expires at the Admin step which forces you to refresh the page before processing.

### Expected Behavior

Should only get displayed at the end.
We shouldn't `writeToEnv('env_general')` before the redis step otherwise it won't work if the defaults don't match redis settings.
It should create the admin user then go to the Setup complete step (without refresh) and Auto login.

### Steps to Reproduce

Install as usual

### Panel Version

latest (c0eedc1)

### Wings Version

1.0.0-beta5 (not relevent)

### Error Logs

```bash
[2024-10-14 01:22:00] production.ERROR: `SELECT` failed: NOAUTH Authentication required. [tcp://localhost:6379] {"exception":"[object] (Predis\\Connection\\ConnectionException(code: 0): `SELECT` failed: NOAUTH Authentication required. [tcp://localhost:6379] at /var/www/pelican/vendor/predis/predis/src/Connection/AbstractConnection.php:131)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

RMartinOscar:
I really don't get the use of `env_general`, `env_database` `env_redis` when its all gonna get saved under `env`.

Boy132:
> The last step is displayed before it should. `env_database.

That looks like a filament bug.

> DB_DATABASE` default will always be `database.sqlite` eventhough `mariadb` or `sqlite` is selected.

Yeah, you are right. It checks for the env variable instead of using `$get`.

> The installer still writes `DB_HOST`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD` eventhough `sqlite` is selected.

That's because hidden fields are written too. But that's not a problem.

> The page expires at the Admin step which forces you to refresh the page before processing.

[...]

PR #625, opened 2024-10-14
Remove no longer needed `View::share`

Boy132:
Fixes a redis error on the installer, ref #623

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -31,6 +31,7 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot(): void
     {
+        // TODO: remove when old admin area gets yeeted
         View::share('appVersion', config('app.version'));
         View::share('appIsGit', false);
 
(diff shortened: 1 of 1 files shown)


Boy132:
> I believe this breaks `/legacy`
> 
> https://github.com/pelican-dev/panel/blob/f95ba6447c1b05c8a255a3b91c642f2d3ae53db2/resources/views/layouts/admin.blade.php#L138

Well, it's called "legacy" for a reason. Better the old admin area is "broken" instead of the installer.

RMartinOscar:
> > I believe this breaks `/legacy`
> > https://github.com/pelican-dev/panel/blob/f95ba6447c1b05c8a255a3b91c642f2d3ae53db2/resources/views/layouts/admin.blade.php#L138
> 
> Well, it's called "legacy" for a reason. Better the old admin area is "broken" instead of the installer.

Following your logic we should just get rid of it entirely as this file is included everywhere it will 500 all the old admin area

RMartinOscar:
That works too

lancepioch (approved):

PR #592, opened 2024-09-27
Adds Auto Deploy Command

notAreYouScared:
Co-authored By: @RMartinOscar 

Adds the Auto Deploy button

PR #627, opened 2024-10-14
Add back auto deploy

RMartinOscar:
Co-authored By: @notAreYouScared

image

image
image

image
image

Supersedes #592
Needs https://github.com/pelican-dev/wings/pull/46

### DIFF
--- a/app/Filament/Resources/ApiKeyResource/Pages/CreateApiKey.php
+++ b/app/Filament/Resources/ApiKeyResource/Pages/CreateApiKey.php
@@ -23,7 +23,7 @@ class CreateApiKey extends CreateRecord
         return $form
             ->schema([
                 Hidden::make('identifier')->default(ApiKey::generateTokenIdentifier(ApiKey::TYPE_APPLICATION)),
-                Hidden::make('token')->default(str_random(ApiKey::KEY_LENGTH)),
+                Hidden::make('token')->default(str_random(config('api.key.secret_length', 32))),
 
                 Hidden::make('user_id')
                     ->default(auth()->user()->id)
--- a/app/Filament/Resources/ApiKeyResource/Pages/CreateApiKey.php
+++ b/app/Filament/Resources/ApiKeyResource/Pages/CreateApiKey.php
@@ -23,7 +23,7 @@ class CreateApiKey extends CreateRecord
         return $form
             ->schema([
                 Hidden::make('identifier')->default(ApiKey::generateTokenIdentifier(ApiKey::TYPE_APPLICATION)),
-                Hidden::make('token')->default(str_random(config('api.key.secret_length', 32))),
+                Hidden::make('token')->default(str_random(ApiKey::KEY_LENGTH)),
 
                 Hidden::make('user_id')
                     ->default(auth()->user()->id)
--- a/app/Http/Controllers/Api/Client/ApiKeyController.php
+++ b/app/Http/Controllers/Api/Client/ApiKeyController.php
@@ -29,7 +29,7 @@ class ApiKeyController extends ClientApiController
      */
     public function store(StoreApiKeyRequest $request): array
     {
-        if ($request->user()->apiKeys->count() >= config('api.key.limit', 25)) {
+        if ($request->user()->apiKeys->count() >= config('panel.api.key_limit')) {
             throw new DisplayException('You have reached the account limit for number of API keys.');
         }
 
--- a/app/Http/Controllers/Api/Client/ApiKeyController.php
+++ b/app/Http/Controllers/Api/Client/ApiKeyController.php
@@ -29,7 +29,7 @@ class ApiKeyController extends ClientApiController
      */
     public function store(StoreApiKeyRequest $request): array
     {
-        if ($request->user()->apiKeys->count() >= 25) {
+        if ($request->user()->apiKeys->count() >= ApiKey::API_KEYS_LIMIT) {
             throw new DisplayException('You have reached the account limit for number of API keys.');
         }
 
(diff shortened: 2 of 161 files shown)


RMartinOscar:
Would be cool if we had an env variable like ```ENV APP_DOCKER=true``` in [Dockerfile](https://github.com/pelican-dev/panel/blob/main/Dockerfile) so we can set the default value of `get('docker')`
Like https://github.com/pelican-dev/panel/blob/a9e11cd246493d713e67ac8c182683e61ff16866/app/Filament/Resources/NodeResource/Pages/EditNode.php#L440
```php
->default(env('APP_DOCKER', false))
```

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -0,0 +1,58 @@
...
+     * Generates a new API key for the logged-in user with only permission to read
+     * nodes, and returns that as the deployment key for a node.
+     *
+     * @throws \App\Exceptions\Model\DataValidationException
+     */
+    public function handle(Request $request, Node $node, ?bool $docker = false): ?string
+    {
+        /** @var ApiKey|null $key */
+        $key = ApiKey::query()
+            ->where('user_id', $request->user()->id)
```
Boy132, line 30: Why the user check? For application keys this shouldn't be needed.

Boy132 (changes requested):

app/Services/Nodes/NodeAutoDeployService.php
RMartinOscar: It won't find the keys without it thus creating thousand duplicates.

RMartinOscar: Im gonna try again but for some reason it didn't work on my side when i was testing that.

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -47,7 +46,7 @@ public function handle(Request $request, Node $node, ?bool $docker = false): ?st
             sprintf(
                 '%s wings configure --panel-url %s --token %s --node %d%s',
                 $docker ? 'docker compose exec -it' : 'sudo',
-                config('app.url'),
+                route('index'),
```
Boy132, line 49: Revert that. The `APP_URL` is correct here.

app/Models/ApiKey.php
```diff
@@ -66,15 +66,18 @@ class ApiKey extends Model
 
     /**
      * Maximum number of Api keys that a user can have.
+     *
+     * @deprecated
```
Boy132, line 70: You can remove the consts instead of deprecating them.

Boy132 (dismissed):

app/Filament/Resources/ApiKeyResource/Pages/CreateApiKey.php
```diff
@@ -23,7 +23,7 @@ public function form(Form $form): Form
         return $form
             ->schema([
                 Hidden::make('identifier')->default(ApiKey::generateTokenIdentifier(ApiKey::TYPE_APPLICATION)),
-                Hidden::make('token')->default(str_random(ApiKey::KEY_LENGTH)),
+                Hidden::make('token')->default(str_random(config('api.key.secret_length', 32))),
```
lancepioch, line 26: This should NOT be configurable!

app/Http/Controllers/Api/Client/ApiKeyController.php
```diff
@@ -29,7 +29,7 @@ public function index(ClientApiRequest $request): array
      */
     public function store(StoreApiKeyRequest $request): array
     {
-        if ($request->user()->apiKeys->count() >= 25) {
+        if ($request->user()->apiKeys->count() >= config('api.key.limit', 25)) {
```
lancepioch, line 32: Use the constant from before

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -0,0 +1,57 @@
...
+     * nodes, and returns that as the deployment key for a node.
+     *
+     * @throws \App\Exceptions\Model\DataValidationException
+     */
+    public function handle(Request $request, Node $node, ?bool $docker = false): ?string
+    {
+        /** @var ApiKey|null $key */
+        $key = ApiKey::query()
+            ->where('key_type', ApiKey::TYPE_APPLICATION)
+            ->where('r_nodes', 1)
```
lancepioch, line 31:
```suggestion
            ->where('r_nodes', true)
```

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -0,0 +1,57 @@
...
+            ->first();
+
+        // We couldn't find a key that exists for this user with only permission for
+        // reading nodes. Go ahead and create it now.
+        if (!$key) {
+            $key = $this->keyCreationService->setKeyType(ApiKey::TYPE_APPLICATION)->handle([
+                'memo' => 'Automatically generated node deployment key.',
+                'user_id' => $request->user()->id,
+            ], ['r_nodes' => 1]);
```
lancepioch, line 40:
```suggestion
            ], ['r_nodes' => true]);
```

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        $token = $key->identifier . $key->token;
+
+        return $token ?
```
lancepioch, line 45: Make it easier to read by getting rid of the ternary

```php
if (! $token) {
    return null;
}
```

config/api.php
lancepioch, line 1: I think this will only cause issues for people that are now allowed to modify it.

lancepioch (changes requested):
I would say remove the api config, put the max api keys in the panel configuration file, but the others (identifier_length and secret_length) should remain constants. Otherwise, looks good!

lancepioch (approved):

Issue #634, opened 2024-10-16
Role permissions "Servers"


XdelaDiscord:
### Current Behavior

When the user has permission to see the server list he does not have the "SHOWING OTHERS' SERVERS" button to see the list of other servers on the panel. I don't know if it's a bug or an oversight.

### Expected Behavior

The "SHOWING OTHERS' SERVERS" button is displayed when you are on the panel in client mode and have permission

### Steps to Reproduce

1. create a role with server permissions
2. add the role to a third party
3. check with this person that they have the "SHOWING OTHERS' SERVERS" button

### Panel Version

Version: 1.0.0-beta10

### Wings Version

wings v1.0.0-beta5

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

RMartinOscar:
Hey the Role Permissions are only effective for the admin area at this time, it will be good when #416 will get merged.
The current client area still uses the legacy permission system.

PR #416, opened 2024-06-18
feat: Client UI translate to Filament (from React)

lancepioch:
Edit By Charles
Below are things that need to be done...Good Luck!

## Profile 
 - [x]  2FA
   	 - [x] Enable
   	 - [x] Disable
   	 - [x] Backup codes
  - [x] Timezone Changing
  - [x] 12/24hr Time Changing
  - [x] Change Password
  - [x] Change Email
  - [x] Change Language
  - [x]  API
	 - [x] Create API Keys
	 - [x] Delete API Keys
	 - [x] Last Used Timestamp
 - [x] SSH Keys
	 - [x] Adding of SSH Keys
	 - [x] Deleting of SSH Keys
 - [x] Activity Logs
 - [x] OAuth Link/ Unlink

## Server List

Needs to be redone... I'll just leave it at that...

## Server Area

 - [x] Console Page
	 - [x] Console
	 - [x] Power Actions
	 - [x] Stats + Graphs
 - [x] File Manager
	 - [x] Directory Listing / Explorer
	 - [x] File Management
		 - [x] Edit Files ( Monaco Code Editor ) 
		 - [x] Delete Files / Folders
		 - [x] Create Files / Folders
		 - [x] Manage File / Folder Permissions
		 - [x] Copy Files
		 - [x] Upload Files
		 - [x] Move Files / Folders
		 - [x] Create / Extract Archives
		 - [x] Download Files
 - [x] Databases
	 - [x] List Databases
	 - [x] View Database Details
	 - [x] Rotate Database Password
	 - [x] Delete Database
	 - [x] Create Database
 - [x] Schedules
	 - [x] List Schedules
	 - [x] Create Schedules
	 - [x] Edit Schedules
	 - [x] Edit Schedule Tasks
 - [x] Users
	 - [x] List Subusers
	 - [x] Create Subusers
	 - [x] Delete Subusers
	 - [x] Edit Subusers
	 - [x] Edit Existing Subuser permissions
- [x]  Backups
	- [x] List Backups
	- [x] Create Backup
	- [x] Delete Backup
	- [x] Lock/Unlock Backup
	- [x] Download Backup
	- [x] Restore from Backup
- [x] Network
	- [x] List Allocations
	- [x] Create new Allocation
	- [x] Change Primary Allocation
- [x] Startup
	- [x] List Startup Command
	- [x] Docker Image ( dropdown + read-only if manually set ) 
	- [x] Variables
- [x] Settings
	- [x] SFTP Details
[...]

PR #636, opened 2024-10-17
Allow rootadmin role

RMartinOscar:
This make it so you can add "root_admin" users again, without this you can't grant any privilege on the client area until #416 gets merged cause the legacy still checks for `Auth::user()->root_admin` not for `Auth::user()->can('resource permission')`.

https://github.com/pelican-dev/panel/issues/634#issuecomment-2416530074


Boy132:
No, this is intentional. There should (ideally) only ever be one root admin - the initial admin account.
And while the current client area still only checks for `root_admin` there are better workarounds than simply giving people root admin.

PR #640, opened 2024-10-20
Fix #606 - Prevent database hosts bulk selection if host has any databases

### LINKED ISSUE #606: [500s] Using "Bulk Delete" on Database Hosts Results in 500
Trying to use the Bulk Delete option on the List Database Hosts, you get a 500. Ideally, this shouldn't be checkable if the database host has databases. There is logic on the Edit Database Host page, but not on the List.

lancepioch:
Fixes #606

### DIFF
--- a/app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
+++ b/app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
@@ -32,13 +32,15 @@ class ListDatabaseHosts extends ListRecords
                     ->sortable(),
                 TextColumn::make('username')
                     ->searchable(),
-                TextColumn::make('max_databases')
-                    ->numeric()
-                    ->sortable(),
+                TextColumn::make('databases_count')
+                    ->counts('databases')
+                    ->icon('tabler-database')
+                    ->label('Databases'),
                 TextColumn::make('node.name')
                     ->numeric()
                     ->sortable(),
             ])
+            ->checkIfRecordIsSelectableUsing(fn (DatabaseHost $databaseHost) => !$databaseHost->databases_count)
             ->actions([
                 EditAction::make(),
             ])
--- a/app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
+++ b/app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
@@ -31,7 +31,6 @@ class ListDatabaseHosts extends ListRecords
                 TextColumn::make('port')
                     ->sortable(),
                 TextColumn::make('username')
-                    ->icon('tabler-user')
                     ->searchable(),
                 TextColumn::make('databases_count')
                     ->counts('databases')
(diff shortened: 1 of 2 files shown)


app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
```diff
@@ -31,14 +31,17 @@ public function table(Table $table): Table
                 TextColumn::make('port')
                     ->sortable(),
                 TextColumn::make('username')
+                    ->icon('tabler-user')
                     ->searchable(),
-                TextColumn::make('max_databases')
-                    ->numeric()
-                    ->sortable(),
```
Boy132, line 37: Why was this removed?

app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
```diff
@@ -31,14 +31,17 @@ public function table(Table $table): Table
...
+                TextColumn::make('databases_count')
+                    ->counts('databases')
+                    ->icon('tabler-database')
+                    ->label('Databases'),
                 TextColumn::make('node.name')
-                    ->numeric()
+                    ->icon('tabler-server-2')
                     ->sortable(),
```
Boy132, line 42: Why should add a placeholder here so the column isn't empty when no node is linked.

app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
```diff
@@ -31,14 +31,17 @@ public function table(Table $table): Table
                 TextColumn::make('port')
                     ->sortable(),
                 TextColumn::make('username')
+                    ->icon('tabler-user')
```
Boy132, line 34: Wouldn't adding a user icon suggest that it's a panel user?

Boy132 (changes requested):

app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
lancepioch, line 34:
```suggestion
```

Good point

lancepioch: Good point

app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
lancepioch: "Max" limit for a resource isn't a good metric to display on the overview. You display current metrics instead.

app/Filament/Resources/DatabaseHostResource/Pages/ListDatabaseHosts.php
lancepioch: Good idea!

lancepioch: I tried it out and I don't think it looks better.

image

image

Boy132: Did you use `->placeholder`?

Boy132 (approved):

PR #647, opened 2024-10-20
Speed up docker start

Freddo3000:
Starting the docker container is hampered due to setting `chown -R www-data:www-data /var/www/html/` on every start, causing it to traverse the entire directory which in our use case is very slow. This PR instead changes it to set permissions as part of the build process.

Sidenote: Is `LE_EMAIL` supposed to be used in addition to `ADMIN_EMAIL`?

### DIFF
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -58,7 +58,7 @@ else
   echo "Starting PHP-FPM only"
 fi
 
-chown -R www-data:www-data . /pelican-data/.env /pelican-data/database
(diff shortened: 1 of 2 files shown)


RMartinOscar:
Not sure that would work.
No `LE_EMAIL` is not in use anymore cause we ditched Nginx (at least for now), `ADMIN_EMAIL` is the Let'sencrypt Email used by Caddy to generate certs.

lancepioch:
Yes, you can remove LE_EMAIL from the readme

parkervcp (approved):
LGTM

PR #670, opened 2024-10-27
Add Soft Deletes to webhooks config table

RMartinOscar:


notAreYouScared (approved):
Tested on MySQL. Fixes the issue.

PR #676, opened 2024-10-27
Update Egg Export Version to PLCN_V1

### LINKED ISSUE #675: Update Egg Export Version tag
We have shifted away from the pterodactyl egg standard.

We should at least change the exported version so when people try to import into pterodactyl they get the correct error.

parkervcp:
resolves #675


RMartinOscar (approved):

RMartinOscar (approved):
LGTM

PR #677, opened 2024-10-27
Allow path in Connection Address

snipcola:
Currently, if the connection address has a path, the port will end up after the path.
001150 Zen Browser 2024-10-27 16 41 06

001152 Zen Browser 2024-10-27 20 58 48

001154 Zen Browser 2024-10-27 20 59 50

This PR ensures the path comes after the port.
001146 Zen Browser 2024-10-27 16 39 58

001132 Zen Browser 2024-10-27 15 55 48

001156 Zen Browser 2024-10-27 21 00 18

### DIFF
--- a/app/Filament/Resources/NodeResource/Pages/CreateNode.php
+++ b/app/Filament/Resources/NodeResource/Pages/CreateNode.php
@@ -93,7 +93,7 @@ class CreateNode extends CreateRecord
                                     $fqdn = array_get($url, 'host', $state);
                                     $scheme = array_get($url, 'scheme', 'https');
                                     $validRecords = gethostbynamel(array_get(parse_url("$scheme://$fqdn"), 'host'));
-                                    
+
                                     if ($validRecords) {
                                         $set('dns', true);
 
--- a/app/Filament/Resources/NodeResource/Pages/CreateNode.php
+++ b/app/Filament/Resources/NodeResource/Pages/CreateNode.php
@@ -89,10 +89,7 @@ class CreateNode extends CreateRecord
                                         return;
                                     }
 
-                                    $url = parse_url($state);
-                                    $fqdn = array_get($url, 'host', $state);
-                                    $scheme = array_get($url, 'scheme', 'https');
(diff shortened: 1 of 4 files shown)


snipcola:

RMartinOscar:
We should probably change Domain Name to Enpoint if we add support for Paths, you shouldn't put port in this field also the scheme part is useless filamentphp gets rid of it

snipcola:
> you shouldn't put port in this field
Can you elaborate, what field are you referencing?

RMartinOscar:
> > you shouldn't put port in this field
> 
> Can you elaborate, what field are you referencing?

The domaine Name field shouldn't include port if you want to use a custom port use the Port field
image
image

snipcola:
> > > you shouldn't put port in this field
> > 
> > 
> > Can you elaborate, what field are you referencing?
> 
> The domaine Name field shouldn't include port if you want to use a custom port use the Port field image image

Oh, right, I already assumed that was the case.
Only for the example, I've predefined the port. `daemon_listen` would be used, like usual.

Boy132:
Why would you use a fqdn with a path? This sounds like a super edge case.

snipcola:
> Why would you use a fqdn with a path? This sounds like a super edge case.

I understand it might be an edge case. There are certain scenarios where it could be useful; for example, if users want to host multiple applications under the same domain. In my use case, I've set up all of my services under dokploy, which uses traefik, and allowing a path in the address allows me to do something like:
- game.domain.com
- game.domain.com/node

It also allows traefik to request one less certificate from letsencrypt, which probably doesn't make too much of a difference - but I do prefer it.

001164 Zen Browser 2024-10-28 10 55 08

lancepioch:
> It also allows traefik to request one less certificate from letsencrypt, which probably doesn't make too much of a difference - but I do prefer it.

You always have the option of requesting a wildcard from LE if needed. The current limit is 50 per week per IP which is plenty for 99% of people.

Subdomains are free, please use them instead of paths! Sorry, this would just add too much overhead for no appreciable benefit. We do appreciate the contributions though!

PR #687, opened 2024-10-29
Fix Node Health not live & Add tooltip

RMartinOscar:
- [x] Add livewire poll so it auto refresh
- [x] Add back version tooltip

image
image

### DIFF
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -307,7 +307,9 @@ class Node extends Model
 
                 if ($message->startsWith('cURL error 6: Could not resolve host')) {
                     $message = str('Could not resolve host');
-                } elseif ($message->startsWith('cURL error 28: Failed to connect to ')) {
+                }
+                
+                if ($message->startsWith('cURL error 28: Failed to connect to ')) {
                     $message = $message->after('cURL error 28: ')->before(' after ');
                 }
 
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -308,7 +308,6 @@ class Node extends Model
                 if ($message->startsWith('cURL error 6: Could not resolve host')) {
                     $message = str('Could not resolve host');
                 }
-                
                 if ($message->startsWith('cURL error 28: Failed to connect to ')) {
                     $message = $message->after('cURL error 28: ')->before(' after ');
                 }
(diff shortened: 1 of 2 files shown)


Boy132:
> Nothing is shown on hover if the node is offline should i add a `Unknown` ?

Yes, something similar to the tooltip from the old admin area.

We should also [lower the refresh rate](https://livewire.laravel.com/docs/wire-poll#timing-control) to 5 seconds (or maybe 10 or 15 even?)

RMartinOscar:
image
image

Maybe its a bit long ?

I searched for a way to define [$pollingInterval](https://filamentphp.com/docs/3.x/widgets/stats-overview#live-updating-stats-polling) in php but it didn't care so i hardcoded it in the blade

Boy132:
> image image
> 
> Maybe its a bit long ?

Is there a way to add a line break?

RMartinOscar:
> Is there a way to add a line break?

I tried without success

Boy132:
On a second thought, do we even print errors to the browser console?

RMartinOscar:
> On a second thought, do we even print errors to the browser console?

LMAO we don't

app/Models/Node.php
```diff
@@ -307,9 +307,7 @@ public function systemInformation(): array
 
                 if ($message->startsWith('cURL error 6: Could not resolve host')) {
                     $message = str('Could not resolve host');
-                }
-
-                if ($message->startsWith('cURL error 28: Failed to connect to ')) {
+                } elseif ($message->startsWith('cURL error 28: Failed to connect to ')) {
```
lancepioch, line 310: You're triggering me, this was fine before

```suggestion
                }
                
                if ($message->startsWith('cURL error 28: Failed to connect to ')) {
```

app/Models/Node.php
```diff
@@ -308,7 +308,7 @@ public function systemInformation(): array
                 if ($message->startsWith('cURL error 6: Could not resolve host')) {
                     $message = str('Could not resolve host');
                 }
-
+
```
lancepioch, line 311:
```suggestion

```

app/Models/Node.php
```diff
@@ -308,7 +308,6 @@ public function systemInformation(): array
                 if ($message->startsWith('cURL error 6: Could not resolve host')) {
                     $message = str('Could not resolve host');
                 }
```
lancepioch, line 310:
```suggestion
                }

```

lancepioch (approved):

PR #691, opened 2024-10-29
Update egg-rust.json to use Pterodactyl's image

RMartinOscar:
parker's yolk is broken


lancepioch:
Fix the yolk instead.

RMartinOscar:
> Fix the yolk instead.

https://github.com/pelican-eggs/yolks/pull/267

PR #692, opened 2024-10-29
Fix trusted proxies settings & Move ips to config & Add ipv6

RMartinOscar:
image
This was due to `config('trustedproxy.proxies')` being a string so you have to implode it otherwise it will put a blank string in by default
image


config/trustedproxy.php
```diff
@@ -25,4 +25,22 @@
...
+        '108.162.192.0/18',
+        '190.93.240.0/20',
+        '188.114.96.0/20',
+        '197.234.240.0/22',
+        '198.41.128.0/17',
+        '162.158.0.0/15',
+        '104.16.0.0/13',
+        '104.24.0.0/14',
+        '172.64.0.0/13',
+        '131.0.72.0/22',
```
notAreYouScared, line 44: Screenshots show ipv6? But they're not listed in any of the changed files?

RMartinOscar: You must have cherrypick the commit cause its in the second one
image

notAreYouScared: They were 100% not there when i looked lol.

notAreYouScared (approved):

app/Filament/Pages/Settings.php
```diff
@@ -159,23 +159,7 @@ private function generalSettings(): array
...
-                            '188.114.96.0/20',
-                            '197.234.240.0/22',
-                            '198.41.128.0/17',
-                            '162.158.0.0/15',
-                            '104.16.0.0/13',
-                            '104.24.0.0/14',
-                            '172.64.0.0/13',
-                            '131.0.72.0/22',
-                        ])),
+                        ->action(fn (Set $set) => $set('TRUSTED_PROXIES', config('trustedproxy.cloudflare'))),
```
lancepioch, line 162: Pull from the official source instead: https://www.cloudflare.com/ips-v4/ and https://www.cloudflare.com/ips-v6/

lancepioch (dismissed):

app/Filament/Pages/Settings.php
RMartinOscar: I'll pull from https://api.cloudflare.com/client/v4/ips so i get both in a single query

>See that tomorrow

RMartinOscar:
Should i make it pull regularly or leave it to only manual via Settings ?

Boy132:
> Should i make it pull regularly or leave it to only manual via Settings ?

Only manual, there is no reason to pull it regulalry.

> Also should i add ENV variables ?

No, it's fine.


But remove the service. That's unnecessary.

app/Services/Helpers/TrustedProxyService.php
```diff
@@ -0,0 +1,42 @@
...
+                $result = json_decode($response->getBody(), true);
+                foreach (config('trustedproxy.auto.keys') as $value) {
+                    $ips->push(...data_get($result, $value));
+                }
+                $ips->unique();
+            }
+        } catch (GuzzleException $e) {
+        }
+
+        return $ips->values()->all();
```
Boy132, line 40: Remove the service and do this in the action on the settings page.

config/trustedproxy.php
```diff
@@ -25,4 +25,15 @@
...
+    /*
+    * Automatically pull ips from url
+    */
+    'auto' => [
+        'url' => 'https://api.cloudflare.com/client/v4/ips',
+        'keys' => [
+            'result.ipv4_cidrs',
+            'result.ipv6_cidrs',
+        ],
+    ],
```
Boy132, line 38: Doesn't need to be in a config.

Boy132 (changes requested):

Boy132 (approved):

PR #699, opened 2024-11-01
Write special env variables at the very end of the installer

### LINKED ISSUE #696: [Installer] Database type should not be persisted until after all the db info is submitted
Otherwise if you hit the back button, it will error out with Access Denied

Boy132:
Closes #696

This makes sure that "special" env variables (in this case the `SESSION_DRIVER`) are only written to the .env file at the very end of the installer.
The `SESSION_DRIVER` variable is the only variable that directly affects the installer when being changed.

### DIFF
--- a/app/Filament/Pages/Installer/PanelInstaller.php
+++ b/app/Filament/Pages/Installer/PanelInstaller.php
@@ -71,7 +71,7 @@ class PanelInstaller extends SimplePage implements HasForms
                 EnvironmentStep::make($this),
                 DatabaseStep::make($this),
                 RedisStep::make($this)
-                    ->hidden(fn (Get $get) => $get('env_general.SESSION_DRIVER') != 'redis' && $get('env_general.QUEUE_CONNECTION') != 'redis' && $get('env_general.CACHE_STORE') != 'redis'),
+                    ->hidden(fn (Get $get) => $get('env_special.SESSION_DRIVER') != 'redis' && $get('env_general.QUEUE_CONNECTION') != 'redis' && $get('env_general.CACHE_STORE') != 'redis'),
                 AdminUserStep::make($this),
                 CompletedStep::make(),
             ])
--- a/app/Filament/Pages/Installer/PanelInstaller.php
+++ b/app/Filament/Pages/Installer/PanelInstaller.php
@@ -104,6 +104,9 @@ class PanelInstaller extends SimplePage implements HasForms
         $this->user ??= User::all()->filter(fn ($user) => $user->isRootAdmin())->first();
         auth()->guard()->login($this->user, true);
 
+        // Special handling for special env variables (looking at you, SESSION_DRIVER)
+        $this->writeToEnv('env_special');
+
         // Redirect to admin panel
         return redirect(Dashboard::getUrl());
     }
--- a/app/Filament/Pages/Installer/Steps/EnvironmentStep.php
+++ b/app/Filament/Pages/Installer/Steps/EnvironmentStep.php
@@ -68,7 +68,7 @@ class EnvironmentStep
                     ->inline()
                     ->options(self::CACHE_DRIVERS)
                     ->default(config('cache.default', 'file')),
-                ToggleButtons::make('env_general.SESSION_DRIVER')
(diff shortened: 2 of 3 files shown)


RMartinOscar:
https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/PanelInstaller.php#L74
```
$get('env_special.SESSION_DRIVER')
$get('env_special.CACHE_STORE')
```
https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/Steps/EnvironmentStep.php#L63

We have to change those too

RMartinOscar (changes requested):
https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/PanelInstaller.php#L74
```
$get('env_special.SESSION_DRIVER')
$get('env_special.CACHE_STORE')
```
https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/Steps/EnvironmentStep.php#L63

We have to change those too

Boy132:
> https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/PanelInstaller.php#L74
> 
> ```
> ->hidden(fn (Get $get) => $get('env_special.SESSION_DRIVER') != 'redis' && $get('env_general.QUEUE_CONNECTION') != 'redis' && $get('env_special.CACHE_STORE') != 'redis'), 
> ```
> 
> https://github.com/pelican-dev/panel/blob/382dcb3868395a7753940e49d358f51719241927/app/Filament/Pages/Installer/Steps/EnvironmentStep.php#L63
> 
> ```
> ToggleButtons::make('env_special.CACHE_STORE') 
> ```
> 
> We have to change those too

Hidden check is fixed. And the cache driver is fine, it doesn't need special treatment.

RMartinOscar:
I had the same 500 as before when hitting next in the db step with redis selected as cache store it errors immediately i can't even fill in the redis pass so it uses no password cause no redis env is set yet.
This is fixed with my proposed changes.

Boy132:
> I had the [same 500](https://github.com/pelican-dev/panel/issues/623#issuecomment-2411617308) as before when hitting next in the db step with redis selected as cache store it errors immediately i can't even fill in the redis host & pass so it uses no password cause no redis env is set yet. This is fixed with my proposed changes.
> 
> > Using redis which is on something else but localhost will also throw

Please try again and if you still get a 500 post the error.
The error you linked is also from the session driver and not the cache.

RMartinOscar:
> > I had the [same 500](https://github.com/pelican-dev/panel/issues/623#issuecomment-2411617308) as before when hitting next in the db step with redis selected as cache store it errors immediately i can't even fill in the redis host & pass so it uses no password cause no redis env is set yet. This is fixed with my proposed changes.
> > > Using redis which is on something else but localhost will also throw
> 
> Please try again and if you still get a 500 post the error. The error you linked is also from the session driver and not the cache.

https://paste.pelistuff.com/kin4

lancepioch:
We should validate the Redis connection too and make sure it works before saving it in the .env, good catch!

Boy132:
> > > I had the [same 500](https://github.com/pelican-dev/panel/issues/623#issuecomment-2411617308) as before when hitting next in the db step with redis selected as cache store it errors immediately i can't even fill in the redis host & pass so it uses no password cause no redis env is set yet. This is fixed with my proposed changes.
> > > > Using redis which is on something else but localhost will also throw
> > 
> > 
> > Please try again and if you still get a 500 post the error. The error you linked is also from the session driver and not the cache.
> 
> https://paste.pelistuff.com/kin4 I only selected cache store in the first step

Yeah so that's a totally different error...

> We should validate the Redis connection too and make sure it works before saving it in the .env, good catch!

The redis connection is validated - in the Redis step.

The problem is the `SoftwareVersionService` that gets called when we run the config clear command. (I assume that's because of dependency injection)

PR #700, opened 2024-11-04
Fixing Docker Environment variable only getting checked for existence instead of value

BlockyBlockling:
This is fixing possible misinterpretations of the provided docker compose file.
It was only checking for the existence of the environment variable „SKIP_CADDY“ instead of checking for its value.

Now it‘s making sure that the Environment variable is actually set to true so if a value like false is set, caddy will still be active.


RMartinOscar:
If you wish to use caddy then you comment it or set it to false if you don't want it set it to whatever but those two values and it will be disabled, i hardly see this condition useful.

PR #701, opened 2024-11-07
Fix AutoDeploy & KeyCreationService

RMartinOscar:
Second param of handle() is a 1d array


RMartinOscar:
is there any reason to be able to pass "permissions" in both `$data` and `$pemissions` ?

Boy132:
> is there any reason to be able to pass "permissions" in both `$data` and `$pemissions` ?

That's probably a leftover from before. You should be fine to combine it/ just use `data`.

notAreYouScared (approved):

PR #702, opened 2024-11-07
Default to 8443 if https

RMartinOscar:


Boy132:
Why? There is no reason to _always_ default to `8443` if using https.

RMartinOscar:
Feels wrong to server https over a http port; Maybe i should set it to 4443 instead of 8443 even 🤔

Boy132:
8080 is not a http port per se. Just with cloudflare.

RMartinOscar:
> 8080 is not a http port per se. Just with cloudflare.

So you would serve https over 81 without thinking about it ?

Boy132:
> > 8080 is not a http port per se. Just with cloudflare.
> 
> So you would serve https over 81 without thinking about it ?

Wings doesn't care about the port. 81 would work with https, 444 with http.
As long as there is no software limitation it makes no sense to have two default ports. That just adds necessary complexity.

PR #704, opened 2024-11-08
Cleanup `SoftwareVersionService`

Boy132:
General cleanup for the `SoftwareVersionService`.

~This also fixes the [non-file cache driver problem](https://github.com/pelican-dev/panel/pull/699#issuecomment-2462943816) with the Installer. (#699)~


RMartinOscar:
I wouldn't use hard coded discord & donate links maybe make it a var in `config/panel.php` ?

notAreYouScared (approved):

PR #705, opened 2024-11-08
Update web installer (again)

### LINKED ISSUE #696: [Installer] Database type should not be persisted until after all the db info is submitted

Boy132:
Closes #696

- `Redis` step has been renamed to `Cache`.
- Database driver and cache driver are now written with their corresponding step.
- Session driver will be written to .env file at the very end to avoid page expire errors.
- NULL values are no longer written to the .env file.
- The database name/ path now updates correctly depending on the database driver.
- ~~Added info for dockerized install on `Setup Complete` step.~~
- Added own step for `Queue`. ~~(Creating the admin user is now the last step.)~~
- Added own step for `Session`.
- `AdminUser` step removed, user data is now entered during `Environment` step.


RMartinOscar (changes requested):
I'd add back defaults for `port`, `host`, `username` (db & cache).
Possibly prevent users from skipping step using the url.
Add
```php
->visible(fn () => is_file('/.dockerenv'))
```
at the Completed Step

Other then that it fixes every issues i had 👍

RMartinOscar (changes requested):
We have to enforce that Queue is redis if cache is redis otherwise you can't enter creds.

app/Filament/Pages/Installer/Steps/SessionStep.php
```diff
@@ -0,0 +1,38 @@
...
+                    ->hintIcon('tabler-question-mark')
+                    ->hintIconTooltip('The driver used for storing sessions. We recommend "Filesystem" or "Database".')
+                    ->required()
+                    ->inline()
+                    ->options(self::SESSION_DRIVERS)
+                    ->disableOptionWhen(fn ($value, Get $get) => $value === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
+                    ->default(config('session.driver')),
+                TextInput::make('env_session.SESSION_SECURE_COOKIE')
+                    ->hidden()
+                    ->default(str_starts_with(url(''), 'https://') ? 'true' : 'false'),
```
RMartinOscar, line 35:
```suggestion
                    ->default(request()->isSecure()),
```

RMartinOscar (changes requested):
Other then that LGTM

app/Filament/Pages/Installer/Steps/QueueStep.php
```diff
@@ -0,0 +1,64 @@
...
+            ->columns()
+            ->schema([
+                ToggleButtons::make('env_queue.QUEUE_CONNECTION')
+                    ->label('Queue Driver')
+                    ->hintIcon('tabler-question-mark')
+                    ->hintIconTooltip('The driver used for handling queues. We recommend "Database".')
+                    ->required()
+                    ->inline()
+                    ->options(self::QUEUE_DRIVERS)
+                    ->disableOptionWhen(fn ($value, Get $get) => $value === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```
RMartinOscar, line 35:
```suggestion
                    ->disableOptionWhen(fn ($state, Get $get) => $state === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```

app/Filament/Pages/Installer/Steps/SessionStep.php
```diff
@@ -0,0 +1,38 @@
...
+            ->label('Session')
+            ->schema([
+                ToggleButtons::make('env_session.SESSION_DRIVER')
+                    ->label('Session Driver')
+                    ->hintIcon('tabler-question-mark')
+                    ->hintIconTooltip('The driver used for storing sessions. We recommend "Filesystem" or "Database".')
+                    ->required()
+                    ->inline()
+                    ->options(self::SESSION_DRIVERS)
+                    ->disableOptionWhen(fn ($value, Get $get) => $value === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```
RMartinOscar, line 31:
```suggestion
                    ->disableOptionWhen(fn ($state, Get $get) => $state === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```

app/Filament/Pages/Installer/Steps/SessionStep.php
Boy132: `request()->isSecure()` might not always return the correct value, e.g. when using proxies.
Checking the url is not the prettiest but it works.

RMartinOscar: https://github.com/pelican-dev/panel/blob/cb522b24ef79f372454a55042fcb5668e322482c/app/Providers/AppServiceProvider.php#L44-L46 should handle that just fine and we use `request()->isSecure()` everywhere else i don't see why wouldn't it work here.

lancepioch:
Use `request()->isSecure()` for now and THEN we can replace all of those instances with a new helper for the future!

app/Filament/Pages/Installer/Steps/QueueStep.php
RMartinOscar, line 35:
```suggestion
                    ->disableOptionWhen(fn ($state, Get $get) => $state === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```

app/Filament/Pages/Installer/Steps/SessionStep.php
RMartinOscar, line 31:
```suggestion
                    ->disableOptionWhen(fn ($state, Get $get) => $state === 'redis' && $get('env_cache.CACHE_STORE') !== 'redis')
```

RMartinOscar (approved):
LGTM

PR #715, opened 2024-11-15
Update 2024_11_04_185326_revamp_api_keys_permissions.php

Rosox-hosting-account:
I try to download this update but this can't do the update so I tell you the solvation


gOOvER:
pls check first discord:

https://discordapp.com/channels/1218730176297439332/1218753783786639380/1306419264407994440

Boy132:
Thank you for your contribution, but this is not a proper fix.
We know about the problem and are looking into it.

PR #722, opened 2024-11-21
Installer: Shows errors on submit & move `runMigrations`

Boy132:
_Yet Another Installer PR._

Fixes errors on submit not showing because of `Halt` exception.
Moves the `runMigrations` call to submit to avoid cache problems.


RMartinOscar (changes requested):
Errors are now shown but if any error occurs (like if redis is not reachable for a second when clicking finish) it will display
```
Migrations failed
SQLSTATE[HY000]: General error: 1 table "permissions" already exists (Connection: sqlite, SQL: create table "permissions" ("id" integer primary key autoincrement not null, "name" varchar not null, "guard_name" varchar not null, "created_at" datetime, "updated_at" datetime))
```
cause migrations already ran even though redis failed causing the finish step to never work

Boy132:
> Errors are now shown but if any error occurs (like if redis is not reachable for a second when clicking finish) it will display
> 
> ```
> Migrations failed
> SQLSTATE[HY000]: General error: 1 table "permissions" already exists (Connection: sqlite, SQL: create table "permissions" ("id" integer primary key autoincrement not null, "name" varchar not null, "guard_name" varchar not null, "created_at" datetime, "updated_at" datetime))
> ```
> 
> cause migrations already ran even though redis failed causing the finish step to never work and you need to manually empty the database

Not sure if I understand you correctly, you got a redis error and that caused migrations to fail?

notAreYouScared:
It should be noted that it creates a blank `database.sqlite` even if you change from sqlite

Boy132:
> It should be noted that it creates a blank `database.sqlite` even if you change from sqlite
> 
> Edit: Image below, I got to the database screen, change the database name to `databa1se.sqlite` and it created both? but `database.sqlite` is a blank file. image

Can't reproduce that. Make sure you are on the right branch/ have the right files.

notAreYouScared (approved):
Looks good, Did notice that it no longer auto logins for you. Might just be a windows env thing.

PR #727, opened 2024-11-26
Add a primary key to 'password_resets' table to allow compatibility with MariaDB Galera.

iriseden:
*Compatibility with MariaDB Galera Cluster*

Adds a `id` field as the primary key in the `password_resets` table.

A primary key is required on each table when using a clustered database with MariaDB Galera.


RMartinOscar:
> [DELETE](https://mariadb.com/kb/en/delete/) operations are unsupported on tables without a primary key. Also, rows in tables without a primary key may appear in a different order on different nodes. [@mariadb](https://mariadb.com/kb/en/mariadb-galera-cluster-known-limitations/)

I believe its just if you are gonna get rid of `password_resets` which we don't.
Also we need to make the change for the following tables as well: egg_mount, mount_node, mount_server.

lancepioch:
Thank you for the contribution, unfortunately I've thought about it and have decided against these changes.

One of the main reasons is that this table is a standard table with its structure coming from the standard `laravel` repository.

Luckily, you should be able to add a primary key to any tables missing one in your own application and there should be no issues with updating your application or database in the future.

Issue #733, opened 2024-12-01
Server Description can be edited while it is disabled in the Settings (new client area)


Leon-JavaScript:
### Current Behavior

The user can edit their server description in the (new) client area while it's diabled in the settings.

### Expected Behavior

The user should not be able to edit their server description in the (new) client area when it's disabled in the settings.

### Steps to Reproduce

1. Disable the Setting that Users can edit their Server description
2. Go to the Settings of a server in the new client area
3. The server description can still be edited

### Panel Version

1.0.0-beta14

### Wings Version

1.0.0-beta7

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

RMartinOscar:
Great catch !
https://github.com/pelican-dev/panel/pull/734

PR #734, opened 2024-12-01
Add config panel.editable_server_descriptions check

RMartinOscar:
Fixes https://github.com/pelican-dev/panel/issues/733


Boy132 (changes requested):
The field should be hidden and not jus disabled.

Boy132 (approved):

PR #740, opened 2024-12-01
Add (global) sorting to server list

### LINKED ISSUE #249: Add sorting and/or grouping to the front end

Boy132:
Closes #249


notAreYouScared:
So the ONLY issue I can see is, this isn't per a user. If i sort them as an admin, it will also sort them for the user that owns the servers.

Boy132:
> So the ONLY issue I can see is, this isn't per a user. If i sort them as an admin, it will also sort them for the user that owns the servers.

Yeah, but I'm not sure how we would add per-user sorting. That sounds like a lot of custom queries and joins.

lancepioch:
I don't know how easy it would be on the Filament side of things... but all you'd have to do is just add an order column on the pivot model between a Server and User (Subuser).

Boy132:
> I don't know how easy it would be on the Filament side of things... but all you'd have to do is just add an order column on the pivot model between a Server and User (Subuser).

That would only for subusers tho, not for owners/ admins.

Issue #583, opened 2024-09-26
Add View pages for admin resources


Boy132:
In order for the `view` permissions to actually do something we need to add view pages (`ViewRecord`) to the following resources:

- [ ] Egg
- [ ] Node
- [x] Role
- [ ] Server
- [x] User
- [x] API Key (?)
- [x] Database Host
- [x] Mount
- [x] Webhooks

PR #742, opened 2024-12-01
Add role permissions for webhooks

Boy132:


RMartinOscar:
Can't view webhook you should be able to access the edit page but everything should be read only (we can even hide events that are not active for better userfriendlyness)
image
image

image

RMartinOscar (changes requested):
Can't view webhook you should be able to access the edit page but everything should be read only (we can even hide events that are not active for better userfriendlyness)
image
image
image

Boy132:
> Can't view webhook you should be able to access the edit page but everything should be read only (we can even hide events that are not active for better userfriendlyness)

See #583

RMartinOscar (approved):

notAreYouScared (approved):

PR #759, opened 2024-12-03
Add Display Width setting

notAreYouScared:
Allow users to pick how "wide" the panel is.

Some have commented on how "big" the panel is.. this will let them pick how "big" they want it to be.


RMartinOscar (approved):
LGTM

PR #770, opened 2024-12-06
Add config value for display width to other pages

Boy132:
Followup for #759. We missed some places.

### DIFF
--- a/app/Filament/Pages/Installer/PanelInstaller.php
+++ b/app/Filament/Pages/Installer/PanelInstaller.php
@@ -42,7 +42,7 @@ class PanelInstaller extends SimplePage implements HasForms
 
     public function getMaxWidth(): MaxWidth|string
     {
-        return MaxWidth::SevenExtraLarge;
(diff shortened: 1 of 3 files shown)


RMartinOscar (approved):

RMartinOscar:
Was there any reason for it to be [7xl](https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Enums/MaxWidth.php#L27) instead of [screen 2xl](https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Enums/MaxWidth.php#L47C10-L47C30) ?

Boy132:
> Was there any reason for it to be [7xl](https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Enums/MaxWidth.php#L27) instead of [screen 2xl](https://github.com/filamentphp/filament/blob/3.x/packages/support/src/Enums/MaxWidth.php#L47C10-L47C30) ?

That's the default Filament uses.

PR #781, opened 2024-12-08
Show login failure message

### LINKED ISSUE #716: Incorrect Logins Forever Load
Since the old admin area was removed i think the exception was also removed. 

Attempt to login to any panel with invalid username/password, observe infinite loading and errors in browsers console. 

image

lancepioch:
Fixes #716

### DIFF
--- a/resources/scripts/components/auth/LoginContainer.tsx
+++ b/resources/scripts/components/auth/LoginContainer.tsx
@@ -55,7 +55,9 @@ const LoginContainer = ({ history }: RouteComponentProps) => {
                 addError({ message: 'Invalid login, please try again.' });
 
                 setToken('');
-                turnstile.reset();
+                if (turnstile) {
+                    turnstile.reset();
+                }
 
                 setSubmitting(false);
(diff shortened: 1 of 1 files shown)


resources/scripts/components/auth/LoginContainer.tsx
```diff
@@ -51,7 +51,8 @@ const LoginContainer = ({ history }: RouteComponentProps) => {
                 history.replace('/auth/login/checkpoint', { token: response.confirmationToken });
             })
             .catch((error) => {
-                console.error(error);
+                setSubmitting(false);
+                addError({ message: 'Invalid login, please try again.' });
 
                 setToken('');
                 turnstile.reset();
```
RMartinOscar, line 58:
```suggestion
                if (turnstile) {
                    turnstile.reset();
                }
```

RMartinOscar (changes requested):
Turnstile can be disabled and causes the last error in #716 console.

RMartinOscar (approved):

PR #825, opened 2024-12-12
Add CreateUser page

Boy132:


notAreYouScared:
Is there any reason for this change other than making it its page? 
Giving it its own page seems too much for something that's so small of an action.

Boy132:
> Is there any reason for this change other than making it its page? Giving it its own page seems too much for something that's so small of an action.

Mainly consistency. Also as preparation for future stuff.

lancepioch (approved):

PR #838, opened 2024-12-24
Add OAuth Settings to Settings Page

notAreYouScared:
Providers that are enabled auto expand. 
Providers not in use are collapsed.
image
image


Boy132:
The settings should be dynamically generated from the config and not be hardcoded like this.

notAreYouScared:
> The settings should be dynamically generated from the config and not be hardcoded like this.

Feel free to open a new pr to replace this.

PR #858, opened 2025-01-03
Add PHP 8.4 Support

lancepioch:


app/Models/ApiKey.php
```diff
@@ -180,11 +177,6 @@ public function getPermission(string $resource): int
 
     private static array $customResourceNames = [];
 
-    public static function registerCustomResourceName(string $resourceName): void
-    {
-        $customResourceNames[] = $resourceName;
-    }
-
```
Boy132, line 187: Add this back. (and this should be using `static::$customResourceNames`)

Boy132 (dismissed):

Boy132:
Also, add 8.4 to the composer file:
https://github.com/pelican-dev/panel/blob/37f9725f2727778165e00ce47d15e920571e6158/composer.json#L5

app/Models/Allocation.php
```diff
@@ -81,11 +82,6 @@ protected function casts(): array
         ];
     }
 
-    public function getRouteKeyName(): string
```
pelican-vehikl, line 84: Alex: We had that for a long time and now is removed, what changed?

pelican-vehikl (approved):

PR #862, opened 2025-01-05
Discard ipAddresses cache if wings is offline + Switch to Select

RMartinOscar:


notAreYouScared:
Until we can figure out why wings fails to report ips some times. This can stay

alexevladgabriel (approved):
LGTM

app/Models/Node.php
```diff
@@ -384,15 +384,15 @@ public function ipAddresses(): array
                 $addresses = Http::daemon($this)->connectTimeout(1)->timeout(1)->get('/api/system/ips')->json();
                 $ips = $ips->concat(fluent($addresses)->get('ip_addresses'));
             } catch (Exception) {
-                // pass
+                return null;
```
lancepioch, line 387:
```suggestion
                return collect();
```

app/Models/Node.php
```diff
@@ -372,7 +372,7 @@ public function statistics(): array
 
     public function ipAddresses(): array
     {
-        return cache()->remember("nodes.$this->id.ips", now()->addHour(), function () {
+        $ips = cache()->remember("nodes.$this->id.ips", now()->addHour(), function () {
             $ips = collect();
             if (is_ip($this->fqdn)) {
                 $ips = $ips->push($this->fqdn);
```
lancepioch, line 378:
```suggestion
```

lancepioch (changes requested):

lancepioch (approved):

PR #871, opened 2025-01-06
Fix/suspend server offline node

### LINKED ISSUE #867: Can't suspend server if node is Offline

RMartinOscar:
Closes #867
image
image

### DIFF
--- a/app/Http/Controllers/Api/Application/Servers/ServerManagementController.php
+++ b/app/Http/Controllers/Api/Application/Servers/ServerManagementController.php
@@ -32,7 +32,7 @@ class ServerManagementController extends ApplicationApiController
      */
     public function suspend(ServerWriteRequest $request, Server $server): Response
     {
-        $this->suspensionService->toggle($server);
+        $this->suspensionService->handle($server);
 
         return $this->returnNoContent();
     }
@@ -44,7 +44,7 @@ class ServerManagementController extends ApplicationApiController
      */
     public function unsuspend(ServerWriteRequest $request, Server $server): Response
     {
-        $this->suspensionService->toggle($server, SuspensionService::ACTION_UNSUSPEND);
+        $this->suspensionService->handle($server, SuspensionService::ACTION_UNSUSPEND);
 
         return $this->returnNoContent();
     }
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Api/Application/Servers/ServerManagementController.php
```diff
@@ -32,7 +32,7 @@ public function __construct(
      */
     public function suspend(ServerWriteRequest $request, Server $server): Response
     {
-        $this->suspensionService->toggle($server);
+        $this->suspensionService->handle($server);
```
lancepioch, line 35: Let's be explicit:

```suggestion
        $this->suspensionService->handle($server, SuspensionService::ACTION_SUSPEND);
```

app/Services/Servers/SuspensionService.php
```diff
@@ -27,7 +28,7 @@ public function __construct(
      *
      * @throws \Throwable
      */
-    public function toggle(Server $server, string $action = self::ACTION_SUSPEND): void
+    public function handle(Server $server, string $action = self::ACTION_SUSPEND): void
```
lancepioch, line 31: I'd like to remove the default action, while we're in here. Also let's switch it from a string to an enum!

```suggestion
    public function handle(Server $server, SuspendAction $action): void
```

app/Services/Servers/SuspensionService.php
```diff
@@ -55,11 +56,7 @@ public function toggle(Server $server, string $action = self::ACTION_SUSPEND): v
         try {
             // Tell daemon to re-sync the server state.
             $this->daemonServerRepository->setServer($server)->sync();
-        } catch (\Exception $exception) {
-            // Rollback the server's suspension status if daemon fails to sync the server.
-            $server->update([
-                'status' => $isSuspending ? null : ServerState::Suspended,
-            ]);
+        } catch (ConnectionException $exception) {
```
lancepioch, line 59: I agree with this change. Yes, wings does need to know that the server is suspended, but if it's currently not connected, it will simply sync that correctly when it reconnects!

lancepioch (changes requested):

lancepioch (approved):

PR #883, opened 2025-01-07
Fix docker build after vite update

Boy132:


RMartinOscar (approved):

Dockerfile
```diff
@@ -31,12 +30,18 @@ COPY Caddyfile /etc/caddy/Caddyfile
...
 
-COPY --from=yarn /build/public/assets ./public/assets
-
 RUN touch .env
 
 RUN composer install --no-dev --optimize-autoloader
 
+WORKDIR /build
+
+RUN yarn run build
```
RMartinOscar, line 39: yarn isn't in the php image

RMartinOscar (changes requested):

PR #885, opened 2025-01-07
Remove `DaemonConnectionException`

Boy132:
Use the normal `ConnectionException` instead of our custom `DaemonConnectionException`.


RMartinOscar:
Not a fan of Global Exception if we have an existing type & what we catch only throws one type of exception

lancepioch (approved):
Awesome, read my mind!

PR #894, opened 2025-01-08
Speed up Docker build and fix entrypoint permissions

PseudoResonance:
The Docker entrypoint script runs artisan commands as root, which creates directories under `storage/framework/cache/data` that are owned by root. This causes errors when running the panel, as it's unable to write to parts of the cache. By running the commands as www-data instead, the permissions are kept unaffected.

I added the Dockerfile to the compose file, so that the image can easily be built from source if desired.

As the `yarn install` command can take a long time, I changed the Dockerfile to copy just the necessary package/yarn files for the installation first. Once that is completed, it copies the rest and builds the app. This sped up cached builds on my computer by about 80 seconds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,14 @@ FROM node:20-alpine AS yarn
 
 WORKDIR /build
 
-COPY . ./
+COPY package.json yarn.lock ./
 
 RUN yarn config set network-timeout 300000 \
-    && yarn install --frozen-lockfile \
-    && yarn run build:production
+    && yarn install --frozen-lockfile
+
+COPY . ./
+
+RUN yarn run build:production
 
 FROM php:8.3-fpm-alpine
 # FROM --platform=$TARGETOS/$TARGETARCH php:8.3-fpm-alpine
@@ -49,7 +52,7 @@ RUN cp .github/docker/supervisord.conf /etc/supervisord.conf && \
     mkdir /var/log/supervisord/
 
 HEALTHCHECK --interval=5m --timeout=10s --start-period=5s --retries=3 \
-  CMD curl -f http://localhost/up || exit 1
+    CMD curl -f http://localhost/up || exit 1
 
 EXPOSE 80 443
 
--- a/compose.yml
+++ b/compose.yml
@@ -28,6 +28,8 @@ x-common:
 services:
   panel:
     image: ghcr.io/pelican-dev/panel:latest
+    build:
+      dockerfile: ./Dockerfile
     restart: always
     networks:
       - default
--- a/.github/docker/entrypoint.sh
+++ b/.github/docker/entrypoint.sh
@@ -33,17 +33,17 @@ ln -s /pelican-data/database/database.sqlite /var/www/html/database/
 
 if ! grep -q "APP_KEY=" .env || grep -q "APP_KEY=$" .env; then
   echo "Generating APP_KEY..."
-  php artisan key:generate --force
+  su -s /bin/ash -c "php artisan key:generate --force" www-data
 else
   echo "APP_KEY is already set."
 fi
 
 ## make sure the db is set up
 echo -e "Migrating Database"
-php artisan migrate --force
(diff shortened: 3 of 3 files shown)


PseudoResonance:

RMartinOscar:
We should run the entrypoint it self as www-data instead of forcing www-data on each commands.
I believe this change would be in supervisord
Or we can use `su www-data` on top of the entryscript (not sure about that tho)

PseudoResonance:
I didn't think running the whole entrypoint script as www-data would be good as entrypoint itself runs chown at the end, seemingly as a final attempt at making sure everything is good.

RMartinOscar:
> > Also now that we are using vite yarn needs to access vendor
> > https://github.com/pelican-dev/panel/blob/a7a7c5ba4d3ca862a4a3971b5828616f836d84ed/tailwind.config.js#L1
> 
> It shouldn't need it to install the dependencies though, right? I don't know much about Vite.

It does require it to build and its been changed to `yarn build` too

Dockerfile
```diff
@@ -1,41 +1,38 @@
...
     libpng-dev libjpeg-turbo-dev freetype-dev libzip-dev icu-dev \
     zip unzip curl \
     caddy ca-certificates supervisor \
     && docker-php-ext-install bcmath gd intl zip opcache pcntl posix pdo_mysql
 
-# Copy the Caddyfile to the container
-COPY Caddyfile /etc/caddy/Caddyfile
+# Install dependencies with Composer
+COPY composer.json composer.lock ./
+COPY app/helpers.php ./app/
```
lancepioch, line 21: This definitely confuses me even more. Definitely don't want to reference this file anywhere in Docker.

lancepioch (changes requested):
I've left you some feedback, thanks

composer.json
```diff
@@ -15,6 +15,7 @@
         "coderflex/filament-turnstile": "^2.2",
         "dedoc/scramble": "^0.10.0",
         "doctrine/dbal": "~3.6.0",
+        "filament/actions": "^3.2",
```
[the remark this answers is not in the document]
PseudoResonance: Sorry, that got in there somehow and I forgot to remove it.

compose.yml
```diff
@@ -28,6 +28,8 @@ x-common:
 services:
   panel:
     image: ghcr.io/pelican-dev/panel:latest
+    build:
+      dockerfile: ./Dockerfile
```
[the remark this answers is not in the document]
PseudoResonance: No, I just tested, and docker compose build or docker compose run --build will only build images with a build section specified. Although I realized `build: ./` can serve the same purpose as this, if the Dockerfile has the default name.

Dockerfile
PseudoResonance: These 3 files are the bare minimum required to run `composer install`. The reason `app/helpers.php` is required as well, is because it's specified as an autoload file in `composer.json`, so `composer install --optimize-autoloader` looks for it.

PseudoResonance: Also the reason behind splitting the install and build up like this is just optimizing for build caching. By only copying the dependency lists first, those first layers only need to rerun when the dependencies change. Then the entire codebase is copied as late as possible to minimize needlessly rerunning. GitHub actions can also be configured to utilize caching (although I didn't do that in this PR), which, with these changes, should help speed up the builds a bit.

Although this doesn't have as big of an impact anymore because of the Vite frontend rewrite. The majority of the build time is now taken up by `yarn run build`, but regardless, I tried to move it down as far as possible.

JustTNE:
I would very much like to see the permissions entrypoint fix committed to the main repo. Otherwise, the panel errors out after about 5 minutes when trying to access any page.

PseudoResonance:
> We should run the entrypoint it self as www-data instead of forcing www-data on each commands. I believe this change would be in supervisord Or we can use `su www-data` on top of the entryscript (not sure about that tho)

I addressed this in #932, where the entrypoint and all other services are now run as www-data to avoid the su privilege escalation.

PR #896, opened 2025-01-08
PostgreSQL Support

PseudoResonance:
Adds PostgreSQL as an alternative database option. Additionally, adds configurable driver option to the DynamicDatabaseConnection part.

The database driver type is saved in a new column `driver` of the DatabaseHost, allowing for potentially other databases in the future if desired.

I've tried to test everything I could think of, but please let me know if there's anything I missed, or if something should be adjusted. I want this to be as minimal of a maintenance burden as possible, but I do really want Postgres support because there's much better support for running it in a Kubernetes HA setup with CloudnativePG.


database/migrations/2025_01_08_052636_add_driver_to_database_hosts copy.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('database_hosts', function (Blueprint $table) {
+            $table->dropColumn('driver');
+        });
+    }
+};
```
RMartinOscar, line 28: Remove ` copy` with the name

database/migrations/2025_01_08_111850_setup_default_postgresql_permissions.php
```diff
@@ -0,0 +1,33 @@
...
+    public function down(): void
+    {
+        switch (Schema::getConnection()->getDriverName()) {
+            case 'pgsql':
+                $database = DB::connection()->getDatabaseName();
+                DB::statement(sprintf('GRANT CONNECT ON DATABASE "%s" TO public', $database));
+                break;
+        }
+    }
+};
```
RMartinOscar, line 33: The user should do it not migrations

database/migrations/2024_11_04_185326_revamp_api_keys_permissions.php
```diff
@@ -22,7 +22,16 @@
...
+            switch (Schema::getConnection()->getDriverName()) {
+                case 'mysql':
+                case 'mariadb':
+                case 'sqlite':
+                    $table->json('permissions');
+                    break;
+                case 'pgsql':
+                    $table->jsonb('permissions');
+                    break;
+            }
```
RMartinOscar, line 34: Use if/else not switch and if you really don't want to use ifs checkout [match](https://www.php.net/manual/en/control-structures.match.php)

database/migrations/2024_07_25_072050_convert_rules_to_array.php
```diff
@@ -16,9 +16,18 @@ public function up(): void
...
+        switch (Schema::getConnection()->getDriverName()) {
+            case 'sqlite':
+            case 'mysql':
+            case 'mariadb':
+                Schema::table('egg_variables', function (Blueprint $table) {
+                    $table->json('rules')->change();
+                });
+                break;
+            case 'pgsql':
+                DB::statement('ALTER TABLE egg_variables ALTER COLUMN rules TYPE jsonb USING to_jsonb(rules)');
```
RMartinOscar, line 28: Unless absolute necessity you shouldn't use raw statements and what i see from other migrations tells me its not mandatory

database/migrations/2024_03_12_154408_remove_nests_table.php
```diff
@@ -28,10 +28,14 @@ public function up(): void
...
-            } else {
-                $table->dropForeign(['nest_id']);
+            switch (Schema::getConnection()->getDriverName()) {
+                case 'mariadb':
+                case 'mysql':
+                    $table->dropForeign('service_options_nest_id_foreign');
+                    break;
+                case 'sqlite':
+                    $table->dropForeign(['nest_id']);
+                    break;
```
RMartinOscar, line 38:
```suggestion
            $table->dropForeign(['nest_id']);
```
Those are old format

database/migrations/2021_03_21_104718_force_cron_month_field_to_have_value_if_missing.php
```diff
@@ -13,7 +13,7 @@
     public function up(): void
     {
         Schema::table('schedules', function (Blueprint $table) {
-            DB::update("UPDATE `schedules` SET `cron_month` = '*' WHERE `cron_month` = ''");
+            DB::update("UPDATE schedules SET cron_month = '*' WHERE cron_month = ''");
```
RMartinOscar, line 16: Don't use raw statements
```suggestion
            DB::table('schedules')->where('cron_month', '')->update(['cron_month' => '*']);
```

database/migrations/2020_03_22_163911_merge_permissions_table_into_subusers.php
```diff
@@ -65,12 +65,21 @@
...
+                case 'pgsql':
+                    $table->jsonb('permissions')->nullable()->after('server_id');
+                    break;
+            }
         });
 
         $cursor = DB::table('permissions')
             ->select(['subuser_id'])
-            ->selectRaw('GROUP_CONCAT(permission) as permissions')
+            ->selectRaw(Schema::getConnection()->getDriverName() === 'pgsql' ? 'STRING_AGG(permission,\',\') AS permissions' : 'GROUP_CONCAT(permission) as permissions')
```
RMartinOscar, line 82: Avoid ternary in migrations

database/migrations/2017_02_10_171858_UpdateAPIKeyColumnNames.php
```diff
@@ -28,9 +31,12 @@ public function up(): void
...
         Schema::table('api_keys', function (Blueprint $table) {
-            if (Schema::getConnection()->getDriverName() !== 'sqlite') {
-                $table->dropForeign('api_keys_user_id_foreign');
-                $table->dropIndex('api_keys_user_id_foreign');
+            switch (Schema::getConnection()->getDriverName()) {
+                case 'mariadb':
+                case 'mysql':
+                    $table->dropForeign('api_keys_user_id_foreign');
+                    $table->dropIndex('api_keys_user_id_foreign');
+                    break;
```
RMartinOscar, line 39: Use if statement

database/migrations/2017_02_02_175548_UpdateColumnNames.php
```diff
@@ -19,13 +19,16 @@ public function up(): void
...
-                $table->dropIndex('servers_node_foreign');
-                $table->dropIndex('servers_owner_foreign');
-                $table->dropIndex('servers_allocation_foreign');
-                $table->dropIndex('servers_service_foreign');
-                $table->dropIndex('servers_option_foreign');
-                $table->dropIndex('servers_pack_foreign');
+            switch (Schema::getConnection()->getDriverName()) {
+                case 'mariadb':
+                case 'mysql':
+                    $table->dropIndex('servers_node_foreign');
```
RMartinOscar, line 25:
```suggestion
                    $table->dropIndex('node');
```
Old format

composer.json
```diff
@@ -15,6 +15,7 @@
         "coderflex/filament-turnstile": "^2.2",
         "dedoc/scramble": "^0.10.0",
         "doctrine/dbal": "~3.6.0",
+        "filament/actions": "^3.2",
```
RMartinOscar, line 18: Not needed `filament/filament` already requires it

RMartinOscar, line 18:
```suggestion
```
Not needed `filament/filament` already requires it

compose.yml
```diff
@@ -28,6 +28,8 @@ x-common:
 services:
   panel:
     image: ghcr.io/pelican-dev/panel:latest
+    build:
+      dockerfile: ./Dockerfile
```
RMartinOscar, line 32:
```suggestion
```
Useless

app/Services/Databases/DatabasePasswordService.php
```diff
@@ -35,10 +35,18 @@ public function handle(Database|int $database): void
...
+                case 'mariadb':
+                    $database->dropUser($database->username, $database->remote);
+                    $database->createUser($database->database, $database->username, $database->remote, $password, $database->max_connections);
+                    $database->assignUserToDatabase($database->database, $database->username, $database->remote);
+                    $database->flush();
+                    break;
+                case 'pgsql':
+                    $database->updateUserPassword($database->database, $database->username, $database->remote, $password);
+                    break;
+            }
```
RMartinOscar, line 49: Use if/else or match

app/Models/DatabaseHost.php
```diff
@@ -57,6 +64,7 @@ class DatabaseHost extends Model
         'password' => 'nullable|string',
         'node_ids' => 'nullable|array',
         'node_ids.*' => 'required|integer,exists:nodes,id',
+        'driver' => 'required|string',
```
RMartinOscar, line 67:
```suggestion
        'driver' => 'required|string|in:' . self::DRIVERS,
```
`public const DRIVERS = 'mysql,mariadb,pgsql';`

app/Livewire/Installer/Steps/DatabaseStep.php
```diff
@@ -102,34 +112,66 @@ public static function make(PanelInstaller $installer): Step
 
     private static function testConnection(string $driver, ?string $host, null|string|int $port, ?string $database, ?string $username, ?string $password): bool
     {
-        if ($driver === 'sqlite') {
-            return true;
-        }
+        switch ($driver) {
```
RMartinOscar, line 115: Use match

app/Extensions/DynamicDatabaseConnection.php
```diff
@@ -6,30 +6,42 @@
...
+            'DB_CHARSET' => 'utf8',
+            'DB_COLLATION' => 'utf8_unicode_ci',
+            'DEFAULT_DB' => 'mysql',
+        ],
+        'pgsql' => [
+            'DB_CHARSET' => 'utf8',
+            'DB_COLLATION' => 'en_US',
+            'DEFAULT_DB' => 'postgres',
+        ],
+    ];
```
RMartinOscar, line 25: Use config instead

Dockerfile
```diff
@@ -1,41 +1,38 @@
...
+    zip unzip curl libpq-dev postgresql \
     caddy ca-certificates supervisor \
-    && docker-php-ext-install bcmath gd intl zip opcache pcntl posix pdo_mysql
+    && docker-php-ext-install bcmath gd intl zip opcache pcntl posix pdo_mysql pdo_pgsql
 
-# Copy the Caddyfile to the container
-COPY Caddyfile /etc/caddy/Caddyfile
+# Install dependencies with Composer
+COPY composer.json composer.lock ./
+COPY app/helpers.php ./app/
```
RMartinOscar, line 21:
```suggestion
# Copy the application code to the container
COPY . .
```

Dockerfile
```diff
@@ -1,41 +1,38 @@
...
 
-# Copy the application code to the container
-COPY . .
+RUN composer install --no-dev --optimize-autoloader
 
-COPY --from=yarn /build/public/assets ./public/assets
+# Install dependencies with Composer
+COPY package.json yarn.lock ./
```
RMartinOscar, line 26:
```suggestion
```

Dockerfile
```diff
@@ -1,41 +1,38 @@
...
+RUN apk add --no-cache yarn
+RUN yarn config set network-timeout 300000 \
+    && yarn install --frozen-lockfile
 
-RUN touch .env
+# Copy the application code to the container
+COPY . .
```
RMartinOscar, line 32:
```suggestion
```

.github/workflows/lint.yaml
```diff
@@ -3,7 +3,7 @@ name: Lint
 on:
   pull_request:
     branches:
-      - '**'
+      - "**"
```
RMartinOscar, line 6:
```suggestion
      - '**'
```

RMartinOscar (changes requested):
Almost there

RMartinOscar:
Also please rebase to not use the PR https://github.com/pelican-dev/panel/pull/894

database/migrations/2025_01_08_111850_setup_default_postgresql_permissions.php
PseudoResonance: Or should it be part of the installer?

database/migrations/2016_09_04_182835_create_notifications_table.php
```diff
@@ -14,7 +14,16 @@ public function up(): void
...
             $table->morphs('notifiable');
-            $table->text('data');
+            switch (Schema::getConnection()->getDriverName()) {
+                case 'sqlite':
+                case 'mysql':
+                case 'mariadb':
+                    $table->text('data');
+                    break;
+                case 'pgsql':
+                    $table->jsonb('data');
```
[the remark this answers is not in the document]
PseudoResonance: Are the notifications only ever storing plain text?

database/migrations/2024_11_04_185326_revamp_api_keys_permissions.php
PseudoResonance: Alright, but I am just curious why you think switch is not more appropriate in this case? Is it because of the worry of accidental fall-through with a missing break?
I figured switch would more elegantly separate the variety of driver types, while still allowing the reuse of code for multiple drivers if desired.

lancepioch:
Sorry, I'm not comfortable with changing that many past migrations and adding switch statements to that many of them, especially from that many years back. Most likely what would have to happen is migration squashing to get it to work properly with Postgresql. Please see here: https://laravel.com/docs/11.x/migrations#squashing-migrations

database/migrations/2016_09_04_182835_create_notifications_table.php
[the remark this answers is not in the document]
PseudoResonance: To answer this, no, the notifications are JSON, and [jsonb is required for PostgreSQL](https://filamentphp.com/docs/3.x/notifications/database-notifications#setting-up-the-notifications-database-table).

app/Livewire/Installer/Steps/DatabaseStep.php
```diff
@@ -102,34 +112,66 @@ public static function make(PanelInstaller $installer): Step
...
+                        'driver' => $driver,
+                        'host' => $host,
+                        'port' => $port,
+                        'database' => $database,
+                        'username' => $username,
+                        'password' => $password,
+                        'charset' => 'utf8mb4',
+                        'collation' => 'utf8mb4_unicode_ci',
+                        'strict' => true,
+                    ]);
```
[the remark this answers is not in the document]
PseudoResonance: Is there a reason why DynamicDatabaseConnection doesn't just have a static `set` function?

database/migrations/2024_11_04_185326_revamp_api_keys_permissions.php
PseudoResonance: I was thinking/reading about this, and I can see why match is nicer than switch. I didn't realize you could match multiple values.

Although, would it make more sense to make an enum for the driver types and just match with those?

app/Models/DatabaseHost.php
lancepioch: There's an enum rule specifically

app/Transformers/Api/Client/DatabaseTransformer.php
```diff
@@ -21,6 +21,7 @@ public function transform(Database $model): array
         $model->loadMissing('host');
 
         return [
+            'driver' => $model->getRelation('host')->driver,
             'id' => $model->id,
             'host' => [
                 'address' => $model->getRelation('host')->host,
```
lancepioch, line 27: Not your fault, but should quick fix this nonetheless:

```suggestion
            'driver' => $model->host->driver,
            'id' => $model->id,
            'host' => [
                'address' => $model->host->host,
```

app/Extensions/DynamicDatabaseConnection.php
```diff
@@ -6,30 +6,24 @@
...
             $host = DatabaseHost::query()->findOrFail($host);
         }
 
         config()->set('database.connections.' . $connection, [
-            'driver' => self::DB_DRIVER,
+            'driver' => $host->driver->value,
             'host' => $host->host,
             'port' => $host->port,
-            'database' => $database,
+            'database' => $database !== '' ? $database : $host->driver->getDefaultOption('test_database'),
```
lancepioch, line 22: Can we switch `'test_database'` to a constant?

app/Enums/DatabaseDriver.php
```diff
@@ -0,0 +1,124 @@
...
+            ],
+            self::Sqlite => [
+                'driver' => 'sqlite',
+                'url' => '',
+                'database' => 'database.sqlite',
+                'prefix' => '',
+                'foreign_key_constraints' => true,
+            ],
+        })[$key] ?? null;
+        return $useEnv ? config()->get(sprintf('database.connections.%s.%s', $this->value, $key), $defaultValue) : $defaultValue;
```
lancepioch, line 100: Please use string interpolation as much as possible!

```suggestion
        return $useEnv ? config()->get("database.connections.{$this->value}.$key", $defaultValue) : $defaultValue;
```

app/Enums/DatabaseDriver.php
```diff
@@ -0,0 +1,124 @@
...
+
+    public function isRemote(): bool
+    {
+        return match($this) {
+            self::Mariadb, self::Mysql, self::Postgresql => true,
+            self::Sqlite => false,
+        };
+    }
+
+    public function getDefaultOption(string $key, bool $useEnv = false): mixed
```
lancepioch, line 39: I like this enum, but the config values definitely should be stored in the config file. The getDefaultOption can just pull using config(). I know this won't work for the port for example, but feel free to just make a getDefaultPort method.

lancepioch (changes requested):
Awesome work so far, thank you for updating the migration(s).

app/Enums/DatabaseDriver.php
PseudoResonance: Wait I was thinking PHP still didn't have string interpolation... Whoops, I'll have to fix that everywhere.

app/Enums/DatabaseDriver.php
PseudoResonance: I wasn't sure what would be best, so I just tried it out.

lancepioch:
Replaced with #1066

PR #898, opened 2025-01-09
Make DynamicDatabaseConnection static

PseudoResonance:
I thought it was odd that this class, which has no internal state or anything, wasn't a static method and required instantiating the class first.
Does this seem reasonable to change, or is there some plan for it?

### DIFF
--- a/app/Extensions/DynamicDatabaseConnection.php
+++ b/app/Extensions/DynamicDatabaseConnection.php
@@ -15,7 +15,7 @@ class DynamicDatabaseConnection
     /**
      * Adds a dynamic database connection entry to the runtime config.
      */
-    public function set(string $connection, DatabaseHost|int $host, string $database = 'mysql'): void
+    public static function set(string $connection, DatabaseHost|int $host, string $database = 'mysql'): void
     {
         if (!$host instanceof DatabaseHost) {
             $host = DatabaseHost::query()->findOrFail($host);
--- a/app/Services/Databases/DatabaseManagementService.php
+++ b/app/Services/Databases/DatabaseManagementService.php
@@ -31,7 +31,6 @@ class DatabaseManagementService
 
     public function __construct(
         protected ConnectionInterface $connection,
-        protected DynamicDatabaseConnection $dynamic,
     ) {}
 
     /**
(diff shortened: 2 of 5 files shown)


Boy132:
We should get rid of `DynamicDatabaseConnection` completely and update `Database` to handle it directly.
Ideally using `DB::build` instead of `DB::connection` and the custom config.

PseudoResonance:
> We should get rid of `DynamicDatabaseConnection` completely and update `Database` to handle it directly. Ideally using `DB::build` instead of `DB::connection` and the custom config.

What about any logic for generating the connection config?

This is just an example from my initial commit, but I needed to get a few different default values depending on the driver. 
https://github.com/PseudoResonance/pelican-panel/blob/support-postgresql/app/Extensions/DynamicDatabaseConnection.php#L9

PseudoResonance:
> Do we even need to set the charset and collation? Worst case we just use the default values from the normal config.

I have no idea. Postgres almost certainly doesn't need it, but the one thing that is really important is the default database. On postgres, you usually want to test a connection by connecting to `postgres`, while mariadb and mysql, you connect to `mysql` to validate it.

PseudoResonance:
What about [something like this](https://github.com/pelican-dev/panel/pull/896/commits/754a28d6816528ea56ce42279e5296be4f5b6348#diff-21bfd1674161421a5cfbda6b4671d1506653f9d1ab4840ac48afe631c15210b8R108)?
The defaults and all other database related things are all [defined in this enum](https://github.com/pelican-dev/panel/pull/896/commits/754a28d6816528ea56ce42279e5296be4f5b6348#diff-502fb64066febb921de5850fb3d9a5810877ac2e04153e5e60417a8f1ec21f55R39). Then it can be fetched fairly easily if you need it for `DB::build`, like in that first example.
I was also thinking perhaps I can just return the whole default options array, and then the desired values can be overwritten. Then the `DB::build`s can be 100% driver agnostic and will still fetch and use all the driver specific options.

These are just some things I was trying out, I'm not sure what would be best. I tried to move all the defaults to a central place where they can be referenced fairly easily from anywhere else, instead of redefining them all over the place.

lancepioch:
@PseudoResonance I think the best path forward is to remove the `DynamicDatabaseConnection` class. Thoughts?

PR #899, opened 2025-01-09
Fix deleting database host when it has assigned nodes

PseudoResonance:
Deleting a database host with assigned nodes currently throws a foreign key constraint error. This migration will cascade delete the node assignments as well, but maybe it'd be preferable to just show an error and ask the user to remove the node assignments?

### DIFF
--- a/database/migrations/2025_01_09_143607_database_host_node_foreign_delete_cascade.php
+++ b/database/migrations/2025_01_09_143607_database_host_node_foreign_delete_cascade.php
@@ -20,7 +20,5 @@ return new class extends Migration
     /**
      * Reverse the migrations.
      */
-    public function down(): void
-    {
-    }
+    public function down(): void {}
 };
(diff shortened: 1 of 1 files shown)


RMartinOscar:
I thought #844 already took care of that

database/migrations/2025_01_09_143607_database_host_node_foreign_delete_cascade.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('database_host_node', function (Blueprint $table) {
+            $table->dropForeign(['database_host_id']);
+            $table->foreign('database_host_id')->references('id')->on('database_hosts')->noActionOnDelete();
+        });
```
RMartinOscar, line 28:
```suggestion
```
No rollback needed

RMartinOscar (changes requested):
No rollback needed

RMartinOscar (approved):

Boy132 (approved):

PR #901, opened 2025-01-09
Fix server reinstall action

Boy132:
image


RMartinOscar:
WHy not use ReinstallServerService instead of the raw repo ?

Boy132:
> WHy not use ReinstallServerService instead of the raw repo ?

That's a good question. :D

RMartinOscar (approved):

PR #902, opened 2025-01-10
Make sure variables are unique per egg

### LINKED ISSUE #545: Egg variable name needs to be unique per egg

RMartinOscar:
Closes #545


Boy132 (commented):
image

Works, but can you also add a custom message?
Something like this:
```php
->validationMessages([
    'unique' => 'A variable with this name already exists.',
])
```

notAreYouScared:
How many eggs have, server password, map, slots, players, etc... 

What's the reason for them having to be unique...

Boy132:
> How many eggs have, server password, map, slots, players, etc...
> 
> What's the reason for them having to be unique...
> 
> Edit, or is this per egg?

It's per egg of course.

Boy132 (approved):

PR #906, opened 2025-01-12
Use Filament labels.

notAreYouScared:
Use filament labels, which will also use filaments translations.


Boy132 (commented):
Use `trans()` or `__()` instead of `Lang::get()`

lancepioch:
Yeah let's stick to `trans('key.here')`

Boy132 (approved):

PR #907, opened 2025-01-13
Fix translations for activity logs

### LINKED ISSUE #897: Translate activity log event with properties
Currently the activity log event is translated without properties, so placeholders aren't replaced.
Example, `:file` is not being replaced:
grafik

https://github.com/pelican-dev/panel/blob/main/app/Filament/Server/Resources/ActivityResource/Pages/ListActivities.php#L23

Old client area code: https://github.com/pelican-dev/panel/blob/release/v1.0.0-beta17/resources/scripts/components/elements/activity/ActivityLogEntry.tsx#L20-L39

Boy132:
Closes #897

**TODO:**
- [ ] Backwards compat for old activity logs
- [ ] SFTP logs (probably requires wings update)


app/Filament/Server/Resources/ActivityResource/Pages/ListActivities.php
```diff
@@ -20,7 +20,11 @@ public function table(Table $table): Table
             ->columns([
                 TextColumn::make('event')
                     ->html()
-                    ->formatStateUsing(fn ($state, ActivityLog $activityLog) => __('activity.'.str($state)->replace(':', '.'))) // TODO: convert properties to a format that trans likes, see ActivityLogEntry.tsx - wrapProperties
+                    ->formatStateUsing(function ($state, ActivityLog $activityLog) {
+                        $properties = $activityLog->wrapProperties();
+
+                        return trans_choice('activity.'.str($state)->replace(':', '.'), array_key_exists('count', $properties) ? $properties['count'] : 1, $properties);
```
RMartinOscar, line 26:
```suggestion
                        return trans_choice('activity.'.str($state)->replace(':', '.'), array_get($properties, 'count', 1));
```

app/Filament/Server/Resources/ActivityResource/Pages/ListActivities.php
```diff
@@ -20,7 +20,11 @@ public function table(Table $table): Table
...
+                    })
```
RMartinOscar, line 27:
```suggestion
                    ->formatStateUsing(fn ($state, ActivityLog $activityLog) => trans_choice('activity.'.str($state)->replace(':', '.'), array_get($activityLog->wrapProperties(), 'count', 1)))
```

RMartinOscar (dismissed):
image
>Try deleting a file using the recyclebin on the right, it works for bulkaction but not the single one

image
>Try renaming a file within a dir

Man idk what's wrong with my gh it keeps Approving & clear my comment when i request changes

Running on wings canary with Quinten
```go
    metadata := make(map[string]interface{})

    // Check if the target is empty
    if fa.Target == "" {
        // Only include the "files" key when no target is provided
        metadata["files"] = []string{fa.Entity}
    } else {
        // Include "from" and "to" fields when a target exists
        metadata["from"] = fa.Entity
        metadata["to"] = fa.Target
    }
```

lancepioch (approved):

PR #914, opened 2025-01-15
Actually use nodeUpdateService not only for keys

RMartinOscar:
image
image

So when you change the port wings updates it automatically via api call

### DIFF
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -202,7 +202,6 @@ class Node extends Model
             ],
             'allowed_mounts' => $this->mounts->pluck('source')->toArray(),
             'remote' => route('filament.app.resources...index'),
-            'ignore_panel_config_updates' => $this->behind_proxy,
         ];
     }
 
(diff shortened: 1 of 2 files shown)


notAreYouScared:
Don't you still need to manually restart wings?

RMartinOscar:
> Don't you still need to manually restart wings?

Ye without wings edit it still gonna be this way

app/Models/Node.php
```diff
@@ -202,6 +202,7 @@ public function getConfiguration(): array
             ],
             'allowed_mounts' => $this->mounts->pluck('source')->toArray(),
             'remote' => route('filament.app.resources...index'),
+            'ignore_panel_config_updates' => $this->behind_proxy,
```
Boy132, line 205: Why?

RMartinOscar: I'v heard people wanted that

lancepioch: Separate PR

Boy132 (approved):

PR #915, opened 2025-01-15
Fix Docker build by readding existing alias for now

lancepioch:
Fixes these string of errors: https://github.com/pelican-dev/panel/actions/workflows/docker-publish.yml

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "type": "module",
     "scripts": {
         "build": "vite build",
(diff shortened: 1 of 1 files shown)


alexevladgabriel:
@lancepioch This will not solve the actual docker workflow.

alexevladgabriel (changes requested):
Required changes to Dockerfile to install deps first and then build the UI assets.

PR #917, opened 2025-01-16
Fix the docker build

QuintenQVD0:
# Changes

- Implement a multi-stage build: first, use Composer to handle dependencies, then build assets with Yarn and Vite, followed by creating the final container.
- The sequence is critical since Vite requires Composer dependencies to function correctly.
- Specify the platform explicitly where necessary for better compatibility.
- Streamline file transfers between stages, ensuring only a single `COPY . .` is required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,65 @@
 # Pelican Production Dockerfile
 
-FROM node:20-alpine AS yarn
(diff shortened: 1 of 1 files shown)


alexevladgabriel (approved):
LGTM

PR #935, opened 2025-01-23
Add roles to owner selector on Create/Edit Server page

RMartinOscar:
Before it shown
`email | user`
`email | user (Admin)`

now it shows
`email | user`
`email | user (Root Admin)`
`email | user (Role1, Role2)`

> I would'v added a screenshot but the select shits the bed the moment i do it


notAreYouScared:
We should limit it to a single role... No point is listing them all when the box is only so big.

RMartinOscar:
> We should limit it to a single role... No point is listing them all when the box is only so big.

Seems fine to me it makes the record use an extra line the overflow is clean but if you really want to limit it i'll do ->first() instead

notAreYouScared (approved):

PR #960, opened 2025-01-29
Add front end badges

notAreYouScared:
Green/Yellow/Red
Yellow is >70% of limit

image
image

stays red if over the limit
image

### DIFF
--- a/app/Filament/Server/Resources/BackupResource.php
+++ b/app/Filament/Server/Resources/BackupResource.php
@@ -20,6 +20,27 @@ class BackupResource extends Resource
 
     protected static bool $canCreateAnother = false;
 
+    public static function getNavigationBadge(): ?string
+    {
+        /** @var Server $server */
+        $server = Filament::getTenant();
+
+        if ($server->backup_limit === 0) {
+            return null;
+        }
+
+        return $server->backups->count() . ' / ' . $server->backup_limit;
+    }
+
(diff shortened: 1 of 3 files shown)


Boy132 (approved):

RMartinOscar:
Should i push those ?
https://github.com/RMartinOscar/panel/tree/charles/front-end-badges

notAreYouScared:
> Should i push those ? https://github.com/RMartinOscar/panel/tree/charles/front-end-badges

I thought about adding them, and I decided not to as schedules don't have limits, so kinda useless, and allocations people normally only have what they get when the server is made.

RMartinOscar:
> > Should i push those ? https://github.com/RMartinOscar/panel/tree/charles/front-end-badges
> 
> I thought about adding them, and I decided not to as schedules don't have limits, so kinda useless, and allocations people normally only have what they get when the server is made.

Got it and how about displaying others even if they are unlimited ?

notAreYouScared:
> > > Should i push those ? https://github.com/RMartinOscar/panel/tree/charles/front-end-badges
> > 
> > 
> > I thought about adding them, and I decided not to as schedules don't have limits, so kinda useless, and allocations people normally only have what they get when the server is made.
> 
> Got it and how about displaying others even if they are unlimited ?

What others? You can't assign unlimited backups or databases

Boy132 (approved):

PR #963, opened 2025-01-31
Add view pages for "simple" resources

Boy132:
Adds view pages (#583) for the "simple" resources which means **everything except Egg, Node and Server**.

`table` and `form` methods were moved to the Resource class so the view pages can use a disabled form and to avoid duplicate code.

---

\+ the table for mounts got a little rework:
grafik


lancepioch (dismissed):
Not sure this will ultimately be needed, but I don't see why not for now

Boy132:
PR is currently on hold until translations are merged.

lancepioch (approved):
Code looks good, but before merging, let's discuss this in team chat

PR #972, opened 2025-02-05
Add default import `url` on `EditEgg`

RMartinOscar:


app/Filament/Components/Actions/ImportEggAction.php
```diff
@@ -46,6 +47,7 @@ protected function setUp(): void
                             TextInput::make('url')
                                 ->label('URL')
                                 ->hint('This URL should point to a single json file')
+                                ->default(fn (Egg $egg) => $egg->update_url)
```
Boy132, line 50: Egg can be null.

```suggestion
                                ->default(fn (Egg $egg) => $egg?->update_url)
```

Boy132 (changes requested):

app/Filament/Components/Actions/ImportEggAction.php
RMartinOscar: As i said in CI, it doesn't bring any error should i nullsafe anyway ?

Boy132: Yes add it anyways, just to be sure.

RMartinOscar:
```php
:50    Using nullsafe property access on non-nullable type App\Models\Egg. Use -> instead.  
         🪪  nullsafe.neverNull
```

Boy132 (approved):

PR #975, opened 2025-02-07
Fix `OAuthProvider` & Add `ColorPicker` for `Authentik`

RMartinOscar:
- [x] Fix 500 Driver [XXXXX] not supported.
- [x] Fix Authentik getting Steam client_secret
- [x] Add ColorPicker for Authentik
image
- [ ] `Make a button on Login that links to an other page if we have to many enabled providers` Or `collapse them somehow`
image

Tested all providers beside Linkedin & Gitlab.


app/Extensions/OAuth/Providers/CommonProvider.php
```diff
@@ -4,7 +4,7 @@
 
 final class CommonProvider extends OAuthProvider
 {
-    protected function __construct(private string $id, private ?string $providerClass, private ?string $icon, private ?string $hexColor)
```
Boy132, line 7: Don't remove the `$providerClass` as parameter.

database/testing.sqlite-journal
Boy132, line 1: Remove journal file.

Boy132 (changes requested):

app/Extensions/OAuth/Providers/CommonProvider.php
RMartinOscar: Why, its always null so i thought i'd get rid of it.

Boy132 (approved):

PR #982, opened 2025-02-07
feat: Adds Azure AD/Entra ID as an OAuth provider

cyberdev-tech:
Gives the user the ability to add Entra ID as an OAuth provider to allow simpler collaboration when working in teams.

Display name is changeable due to the fact that people call Entra ID different things due to their name change and use cases. Some call it Azure Active Directory (old name) and some call it Entra ID or just Microsoft.


lancepioch:
This should be implemented as a plugin in the future, thank you for your contribution.

PR #993, opened 2025-02-10
fix(docker): enable multi-arch builds

QuintenQVD0:
## Changes & Explanation  

- Since building PHP on QEMU is very slow, we use `build-php-base` to create ARM64 and AMD64 base images on their respective native runners for faster builds.  
- These images are then exported as workflow artifacts.  

- In `build-and-push`, the two artifacts are downloaded, tagged, and pushed to a temporary registry (explained below).  

For some reason, even though the images appear in `docker images`, they cannot be used directly ([see issue](https://github.com/docker/build-push-action/issues/1176)). The only way to use them is to push them to a temporary registry running on port `5000`. Because of this, Docker Buildx must be started in `host` mode, otherwise, it cannot access the registry on that port.  

- This change reduces the first uncached workflow run to about **10 minutes**, and subsequent runs to approximately **3 minutes**.

afbeelding


parkervcp (approved):
LGTM. I hate that this is needed but what can we do beyond that?

RMartinOscar (approved):

PR #995, opened 2025-02-10
Fix incorrect `Allocation` permission in Node's `AllocationsRelationManager`

RMartinOscar:
https://github.com/pelican-dev/panel/blob/cba4cf11aa25b0fbd64e3985634bb2878f8dd332/app/Filament/Admin/Resources/NodeResource/RelationManagers/AllocationsRelationManager.php#L110


Boy132:
We don't use role permissions for allocations.
If you can update a node you can create/ update allocations.

RMartinOscar:
Then why is there a delete allocation permission on BulkdeleteAction of The node's AllocationRelationManager ?

PR #999, opened 2025-02-11
Fix `Health` `ScheduleCheck` 

RMartinOscar:
https://github.com/pelican-dev/panel/blob/a6963ad8026a7d7f31fc6fc6423e8eeba4022de8/app/Console/Kernel.php#L59 Will always fail cause [it checks if ScheduleCheck is registered](https://github.com/spatie/laravel-health/blob/main/src/Commands/ScheduleCheckHeartbeatCommand.php#L17-L25)
image

image

### DIFF
--- a/app/Checks/ScheduleCheck.php
+++ b/app/Checks/ScheduleCheck.php
@@ -4,10 +4,10 @@ namespace App\Checks;
 
 use Carbon\Carbon;
 use Composer\InstalledVersions;
-use Spatie\Health\Checks\Checks\ScheduleCheck as Check;
+use Spatie\Health\Checks\Checks\ScheduleCheck as BaseCheck;
 use Spatie\Health\Checks\Result;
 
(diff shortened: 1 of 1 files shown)


app/Checks/ScheduleCheck.php
```diff
@@ -4,48 +4,11 @@
 
 use Carbon\Carbon;
 use Composer\InstalledVersions;
-use Spatie\Health\Checks\Check;
+use Spatie\Health\Checks\Checks\ScheduleCheck as Check;
 use Spatie\Health\Checks\Result;
 
 class ScheduleCheck extends Check
```
Boy132, line 10:
```suggestion
use Spatie\Health\Checks\Checks\ScheduleCheck as BaseCheck;
use Spatie\Health\Checks\Result;

class ScheduleCheck extends BaseCheck
```

Boy132 (approved):

PR #1005, opened 2025-02-12
Chore Upgrade Dependencies

RMartinOscar:
Tackle https://github.com/pelican-dev/panel/security/dependabot/93
And update other dependencies


resources/views/filament/plugins/monaco-editor.blade.php
notAreYouScared, line 1: I still don't agree that the preview should be added back.

Boy132: The preview is disabled in the config anyways?

https://github.com/pelican-dev/panel/blob/1e7a901371328c21467bf35ae61826f28e98d560/config/filament-monaco-editor.php#L5

notAreYouScared: Then why re add the code?

Boy132: Because it's the default code of the plugin. Makes updating the view easier if we are as close to the original as possible.

notAreYouScared: If thats the argument then why are we manually updating the js versions in the code.... the plugon author has not updated the plugon to 0.52.2

Boy132: Because that's just a minor change. And if the plugin author updates we can just overwrite it.

PR #1009, opened 2025-02-14
Interpret `Server` `StartupCommand` variables

rmartinoscar:
image
image
image

### DIFF
--- a/app/Filament/Server/Pages/Startup.php
+++ b/app/Filament/Server/Pages/Startup.php
@@ -47,7 +47,7 @@ class Startup extends ServerFormPage
                         'lg' => 4,
                     ])
                     ->autosize()
-                    ->formatStateUsing(fn (Server $server, StartupCommandService $StartupCommandService) => $StartupCommandService->handle($server))
+                    ->formatStateUsing(fn (Server $server, StartupCommandService $service) => $service->handle($server))
                     ->readOnly(),
                 TextInput::make('custom_image')
                     ->label('Docker Image')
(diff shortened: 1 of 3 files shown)


notAreYouScared:
Honestly not a fan.

app/Filament/Server/Pages/Startup.php
```diff
@@ -46,6 +47,7 @@ public function form(Form $form): Form
                         'lg' => 4,
                     ])
                     ->autosize()
+                    ->formatStateUsing(fn (Server $server, StartupCommandService $StartupCommandService) => $StartupCommandService->handle($server))
```
Boy132, line 50: _(suggestion below not pinted)_
```suggestion
                    ->formatStateUsing(fn (Server $server, StartupCommandService $service) => $service->handle($server, !auth()->user()->can(Permission::ACTION_STARTUP_READ, $server)))
```
https://github.com/pelican-dev/panel/blob/b355830db4b0a58e595d85ea82cc9d68511304c4/app/Transformers/Api/Client/ServerTransformer.php#L64

Boy132 (dismissed):

Boy132:
> Honestly not a fan. How do I know its using a variable and not a typed in number?

The old client area did this too. I think it's fine. Especially for more complex startup commands it's nicer this way.

app/Filament/Server/Pages/Startup.php
RMartinOscar: I thought of this aswell but the page isn't even accessible without it so why double check ?
https://github.com/pelican-dev/panel/blob/b355830db4b0a58e595d85ea82cc9d68511304c4/app/Filament/Server/Pages/Startup.php#L156

notAreYouScared (approved):

PR #1015, opened 2025-02-17
Don't log duplicated `OauthProvider`s during tests

rmartinoscar:
image

https://github.com/pelican-dev/panel/blob/5be4e22a0c932f11c5aba430942ae470bdd04970/app/Extensions/OAuth/Providers/OAuthProvider.php#L22-L26
I purposely left that in for future plugins

### DIFF
--- a/app/Extensions/OAuth/Providers/OAuthProvider.php
+++ b/app/Extensions/OAuth/Providers/OAuthProvider.php
@@ -20,7 +20,10 @@ abstract class OAuthProvider
     protected function __construct()
     {
         if (array_key_exists($this->getId(), static::$providers)) {
-            logger()->warning("Tried to create duplicate OAuth provider with id '{$this->getId()}'");
+            // @phpstan-ignore-next-line
+            if (!app()->runningUnitTests()) {
+                logger()->warning("Tried to create duplicate OAuth provider with id '{$this->getId()}'");
+            }
 
             return;
         }
--- a/app/Extensions/OAuth/Providers/OAuthProvider.php
+++ b/app/Extensions/OAuth/Providers/OAuthProvider.php
@@ -4,6 +4,7 @@ namespace App\Extensions\OAuth\Providers;
 
 use Filament\Forms\Components\TextInput;
 use Filament\Forms\Components\Wizard\Step;
+use Illuminate\Foundation\Application;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Str;
 use SocialiteProviders\Manager\SocialiteWasCalled;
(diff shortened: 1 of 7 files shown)


Boy132:
Why is it trying to register them multiple times? That would mean that `boot` runs multiple times, which it shouldn't.

rmartinoscar:
> Why is it trying to register them multiple times? That would mean that `boot` runs multiple times, which it shouldn't.

Everytime i run `php artisan test` or `phpstan analyze` it floods my logs

Boy132:
> > Why is it trying to register them multiple times? That would mean that `boot` runs multiple times, which it shouldn't.
> 
> Everytime i run `php artisan test` or `phpstan analyze` it floods my logs

If you run tests that's normal. But in a normal production env the logs shouldn't get flooded.

rmartinoscar:
> If you run tests that's normal. But in a normal production env the logs shouldn't get flooded.

Ok but its gonna flood github too especially when we move to Pest with `--parallel`

Boy132:
> > If you run tests that's normal. But in a normal production env the logs shouldn't get flooded.
> 
> Ok but its gonna flood github too especially when we move to Pest with `--parallel`

Then only [write to the log](https://github.com/pelican-dev/panel/blob/5be4e22a0c932f11c5aba430942ae470bdd04970/app/Extensions/OAuth/Providers/OAuthProvider.php#L23) when not running tests (`$app->runningUnitTests()`)

Boy132 (dismissed):

app/Extensions/OAuth/Providers/OAuthProvider.php
```diff
@@ -20,7 +20,10 @@ public static function get(?string $id = null): array|self
     protected function __construct()
     {
         if (array_key_exists($this->getId(), static::$providers)) {
-            logger()->warning("Tried to create duplicate OAuth provider with id '{$this->getId()}'");
+            // @phpstan-ignore-next-line
```
lancepioch, line 23: I'm going to disable this lol

lancepioch, line 23: I'm going to disable this lol...

Dependency inject the `Illuminate\Foundation\Application $app` instance

lancepioch (changes requested):

app/Extensions/OAuth/Providers/OAuthProvider.php
rmartinoscar: I thought i'd throw it as `app()` for the time being then we'd replace every helpers at once, like i plan on doing with `request()`

lancepioch (approved):

PR #1018, opened 2025-02-19
Validate eggVariable `default_value` using its `rules`

rmartinoscar:
image
image
image

I know about but i need it raw otherwise you have to save first.
https://github.com/pelican-dev/panel/blob/324fc4b7d57d27e7b20ebcae9686af94ba2f0faa/app/Models/EggVariable.php#L85-L88

### DIFF
--- a/app/Filament/Admin/Resources/EggResource/Pages/CreateEgg.php
+++ b/app/Filament/Admin/Resources/EggResource/Pages/CreateEgg.php
@@ -201,6 +201,7 @@ class CreateEgg extends CreateRecord
                                         ->required(),
                                     TextInput::make('default_value')
                                         ->label(trans('admin/egg.default_value'))
+                                        ->required(fn (Get $get) => in_array('required', $get('rules')))
                                         ->rules(fn (Get $get) => $get('rules'))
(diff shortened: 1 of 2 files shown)


app/Filament/Admin/Resources/EggResource/Pages/CreateEgg.php
```diff
@@ -199,7 +199,11 @@ public function form(Form $form): Form
                                             '*' => trans('admin/egg.error_reserved'),
                                         ])
                                         ->required(),
-                                    TextInput::make('default_value')->label(trans('admin/egg.default_value'))->maxLength(255),
+                                    TextInput::make('default_value')
+                                        ->label(trans('admin/egg.default_value'))
+                                        ->required(fn (Get $get) => in_array('required', $get('rules')))
```
Boy132, line 204: Not sure if the default value should be required. Sometimes you only want it to be required when creating a server but not on the egg itself.

What do you think, @QuintenQVD0 ?

QuintenQVD0: I do agree. For example for password fields.

PR #1031, opened 2025-02-25
Fix `StoreServerRequest` for deployment

Boy132:
- Add rules for `tags`
- Add `sometimes` rule to `locations` 
- Fix type for `locations`, `tags` and `dedicated_ip`

### DIFF
--- a/app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
+++ b/app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
@@ -55,9 +55,11 @@ class StoreServerRequest extends ApplicationApiRequest
 
             // Automatic deployment rules
             'deploy' => 'sometimes|required|array',
-            'deploy.locations' => 'array',
-            'deploy.locations.*' => 'required_with:deploy.locations,integer|min:1',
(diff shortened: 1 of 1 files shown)


app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
```diff
@@ -55,9 +55,11 @@ public function rules(): array
 
             // Automatic deployment rules
             'deploy' => 'sometimes|required|array',
-            'deploy.locations' => 'array',
-            'deploy.locations.*' => 'required_with:deploy.locations,integer|min:1',
-            'deploy.dedicated_ip' => 'required_with:deploy,boolean',
+            'deploy.locations' => 'sometimes|array',
+            'deploy.locations.*' => 'required_with:deploy.locations|integer|min:1',
+            'deploy.tags' => 'array',
```
rmartinoscar, line 60: Shouldn't the tags be sometimes as well ?

Boy132: No, tags need to be present. (but can be an empty array)

PR #1034, opened 2025-02-26
Make sure tests fails on composer error

rmartinoscar:


composer.lock
Boy132, line 1: Revert file change (no longer needed after #1036)

Boy132 (changes requested):

PR #1036, opened 2025-02-27
Fix composer.lock

Boy132:
`myclabs/deep-copy` was under dev packages but is a normal requirement.


QuintenQVD0 (approved):
I can confirm that `composer install --no-dev --optimize-autoloader` now work without any issues

PR #1049, opened 2025-03-01
Add missing `Database` address field

rmartinoscar:
image
image
image
image

### DIFF
--- a/app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
+++ b/app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
@@ -665,11 +665,13 @@ class EditServer extends EditRecord
                                             ->label(trans('admin/databasehost.table.host'))
                                             ->disabled()
                                             ->formatStateUsing(fn ($record) => $record->address())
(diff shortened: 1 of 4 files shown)


notAreYouScared:
Why use the display name? And not th3 actual host....

rmartinoscar:
> Why use the display name? And not the actual host....
You guys didn't want me to add an alias so i used displayname instead like Lance told me.
https://github.com/pelican-dev/panel/pull/403#issuecomment-2203908336

notAreYouScared:
Hmm. I dont think people will use display name as an alias... 

If I put US East. As the display name... what are they going to do with `US East:3306`

Boy132:
The display name is intended to be the alias. The description is just wrong/ outdated.

rmartinoscar:
> The display name is intended to be the alias. The description is just wrong/ outdated.

What should i set it to ?

lancepioch:
> > The display name is intended to be the alias. The description is just wrong/ outdated.
> 
> What should i set it to ?

Either `Nickname` or `Display Name` or `Display` or `Name`

Boy132 (approved):

PR #1052, opened 2025-03-03
Avoid unnecessary node fqdm queries

JoanFo1456:
Using cache avoids to do unnecesary queries.


JoanFo1456:

rmartinoscar:
I don't see this being a problem and caching will just avoid hitting the database not the dns queries.

PR #1058, opened 2025-03-05
Force `2fa_required` to no one by default during Tests

rmartinoscar:


Boy132 (commented):
Why?

rmartinoscar:
> Why?

If im testing on a panel that uses a different value 90% of tests fail.
And this will allow us to do additionnal testing like test if its requested when we enforce only admins or every body.

tests/TestCase.php
```diff
@@ -30,6 +30,7 @@ protected function setUp(): void
         // "an error occurred" message), we can probably assume that the exception isn't one that
         // is recognized as being user viewable.
         config()->set('app.debug', false);
+        config()->set('panel.auth.2fa_required', 0);
```
lancepioch, line 33: ?

```suggestion
        config()->set('panel.auth.2fa_required', false);
```

rmartinoscar: No 0 = no one, 1 = admin only, 2 = every body

lancepioch: Completely forgot, we should make an enum for that now.

lancepioch (approved):

PR #1060, opened 2025-03-05
Add missing activity loggers on client area

notAreYouScared:
Closes #1055,  Backup Download will come in another pr when downloading is reworked for backups.

PR #1063, opened 2025-03-06
Fix subuser activity log

Boy132:
- Fix activity log subject when inviting subuser (followup for #1060)
- Add permissions to properties of activity log (followup for #1060)
- Always add `websocket.connect` permission to new subusers (it's a [default permission](https://github.com/pelican-dev/panel/blob/a9e4495c915133aa705f56994523a13a3a314ab6/app/Http/Controllers/Api/Client/Servers/SubuserController.php#L139-L153))

### DIFF
--- a/app/Filament/Server/Resources/UserResource.php
+++ b/app/Filament/Server/Resources/UserResource.php
@@ -121,11 +121,13 @@ class UserResource extends Resource
                     ->action(function (array $data, SubuserUpdateService $subuserUpdateService, User $user) use ($server) {
                         $subuser = Subuser::query()->where('user_id', $user->id)->where('server_id', $server->id)->first();
 
-                        if (in_array('console', $data['control'])) {
-                            $data['websocket'][0] = 'connect';
-                        }
+                        $permissions = collect($data)
+                            ->forget('email')
+                            ->flatMap(fn ($permissions, $key) => collect($permissions)->map(fn ($permission) => "$key.$permission"))
+                            ->push(Permission::ACTION_WEBSOCKET_CONNECT)
+                            ->unique()
+                            ->all();
 
-                        $permissions = collect($data)->forget('email')->map(fn ($permissions, $key) => collect($permissions)->map(fn ($permission) => "$key.$permission"))->flatten()->all();
                         $subuserUpdateService->handle($subuser, $server, $permissions);
 
                         Notification::make()


app/Filament/Server/Resources/UserResource/Pages/ListUsers.php
```diff
@@ -387,19 +387,17 @@ protected function getHeaderActions(): array
...
                     $permissions = collect($data)->forget('email')->map(fn ($permissions, $key) => collect($permissions)->map(fn ($permission) => "$key.$permission"))->flatten()->all();
+                    $permissions = array_unique(array_merge($permissions, [Permission::ACTION_WEBSOCKET_CONNECT]));
 
                     try {
-                        $service->handle($server, $email, $permissions);
+                        $subuser = $service->handle($server, $email, $permissions);
 
                         Activity::event('server:subuser.create')
-                            ->subject($data['email'])
+                            ->subject($subuser->user)
```
rmartinoscar, line 397:
```suggestion
                            ->subject($subuser->user->email)
```
The raw user is stored in actor

app/Filament/Server/Resources/UserResource/Pages/ListUsers.php
```diff
@@ -387,19 +387,17 @@ protected function getHeaderActions(): array
                 ->action(function (array $data, SubuserCreationService $service) use ($server) {
                     $email = strtolower($data['email']);
 
-                    if (in_array('console', $data['control'])) {
-                        $data['websocket'][0] = 'connect';
-                    }
-
                     $permissions = collect($data)->forget('email')->map(fn ($permissions, $key) => collect($permissions)->map(fn ($permission) => "$key.$permission"))->flatten()->all();
+                    $permissions = array_unique(array_merge($permissions, [Permission::ACTION_WEBSOCKET_CONNECT]));
```
rmartinoscar, line 391:
```suggestion
                    $permissions = collect($data)
                        ->forget('email')
                        ->flatMap(fn ($permissions, $key) => collect($permissions)->map(fn ($permission) => "$key.$permission"))
                        ->push(Permission::ACTION_WEBSOCKET_CONNECT)
                        ->unique()
                        ->all();
```

rmartinoscar:
If you uncheck console when editing the subuser then it doesn't have websocket.connect
I think we should show it to the end user but disable it so they can't uncheck it, that would avoid the merging garbage and would explain *why a no perm user has 1 perm*.

lancepioch (approved):

rmartinoscar (approved):

PR #1065, opened 2025-03-07
Add Laravel Data package, also some small fixes

lancepioch:


app/Console/Commands/Egg/CheckEggUpdatesCommand.php
```diff
@@ -34,15 +34,19 @@ private function check(Egg $egg, EggExporterService $exporterService): void
...
         unset($currentJson->exported_at);
 
-        $updatedJson = json_decode(file_get_contents($egg->update_url));
+        $updatedEgg = file_get_contents($egg->update_url);
+        assert($updatedEgg !== false);
+        $updatedJson = json_decode($updatedEgg);
         unset($updatedJson->exported_at);
 
-        if (md5(json_encode($currentJson)) === md5(json_encode($updatedJson))) {
+        if (md5((string) json_encode($currentJson)) === md5((string) json_encode($updatedJson))) {
```
rmartinoscar, line 42: Maybe fluent is the move here otherwise Pest mutations will fail.

app/Models/Mount.php
```diff
@@ -41,28 +40,27 @@ class Mount extends Model implements Validatable
...
+        'target' => ['required', 'string'],
+        'read_only' => ['sometimes', 'boolean'],
+        'user_mountable' => ['sometimes', 'boolean'],
     ];
 
     /**
      * Implement language verification by overriding Eloquence's gather rules function.
      *
-     * @return array<string|string[]>
+     * @return array<array-key, string[]|>
```
rmartinoscar, line 57:
```suggestion
     * @return array<array-key, string[]>
```

app/Console/Commands/Egg/CheckEggUpdatesCommand.php
lancepioch: There aren't any mutation tests for this.

app/Models/Mount.php
lancepioch: Thanks!

app/Console/Commands/Egg/CheckEggUpdatesCommand.php
rmartinoscar, line 42:
```suggestion
        if (md5(json_encode($currentJson, JSON_THROW_ON_ERROR)) === md5(json_encode($updatedJson, JSON_THROW_ON_ERROR))) {
```

rmartinoscar (approved):

PR #1068, opened 2025-03-08
Refactor captcha

Boy132:
Adds a similar system like oauth for captacha. Allows to create additional captcha providers.
Removes `coderflexx/filament-turnstile`.


resources/views/filament/components/turnstile-captcha.blade.php
```diff
@@ -33,13 +28,7 @@
...
         })()"
     >
-        <div data-sitekey="{{config('turnstile.turnstile_site_key')}}"
-            data-theme="{{ $theme }}"
-            data-language="{{ $language }}"
-            data-size="{{ $size }}"
-            x-ref="turnstile"
-            >
-        </div>
+        <div data-sitekey="{{env('CAPTCHA_TURNSTILE_SITE_KEY')}}" x-ref="turnstile"></div>
```
rmartinoscar, line 31: I don't like getting rid of config and using raw env() call in blade
> Boy: Don't worry. Both oauth and captcha will switch to the database eventually.

app/Extensions/Captcha/Providers/CaptchaProvider.php
```diff
@@ -0,0 +1,103 @@
+<?php
```
rmartinoscar, line 1: Do we really need to make a whole CaptchaProvider parent class when turnstile is already backward compatible with reCaptcha & hCaptcha ?
> Boy: Yes, to allow custom providers. Just like with oauth

notAreYouScared:
Appears to still be broken :(

image

app/Extensions/Captcha/Providers/TurnstileProvider.php
```diff
@@ -0,0 +1,96 @@
...
+        $response = Http::asJson()
+            ->timeout(15)
+            ->connectTimeout(5)
+            ->throw()
+            ->post('https://challenges.cloudflare.com/turnstile/v0/siteverify', [
+                'secret' => $secret,
+                'response' => $captchaResponse,
+            ]);
+
+        return count($response->json()) ? $response->json() : [
```
rmartinoscar, line 79:
```suggestion
        return [
```
It will throw if it fails

app/Filament/Components/Forms/Fields/TurnstileCaptcha.php
```diff
@@ -0,0 +1,26 @@
...
+
+    protected string $view = 'filament.components.turnstile-captcha';
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->hiddenLabel();
+
+        $this->dehydrated(false);
```
rmartinoscar, line 20:
```suggestion
```
Components are hydrated by default this is unneeded

bootstrap/app.php
```diff
@@ -44,7 +44,7 @@
             'bindings' => \Illuminate\Routing\Middleware\SubstituteBindings::class,
             'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
             'node.maintenance' => \App\Http\Middleware\MaintenanceMiddleware::class,
-            'recaptcha' => \App\Http\Middleware\VerifyReCaptcha::class,
+            'captcha' => \App\Http\Middleware\VerifyCaptcha::class,
```
rmartinoscar, line 47:
```suggestion
```
Nothing calls for it so i don't see the point in making an alias

notAreYouScared:
image

 After the latest commit.

resources/views/filament/components/turnstile-captcha.blade.php
lancepioch: Switch back to config, I don't think these should be in the database.

bootstrap/app.php
lancepioch: If true, I agree.

app/Http/Middleware/VerifyCaptcha.php
```diff
@@ -0,0 +1,38 @@
...
+use Illuminate\Http\Response;
+use App\Events\Auth\FailedCaptcha;
+use App\Extensions\Captcha\Providers\CaptchaProvider;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+readonly class VerifyCaptcha
+{
+    public function __construct(private Application $app) {}
+
+    public function handle(Request $request, \Closure $next): mixed
```
lancepioch, line 16:
```suggestion
    public function handle(Request $request, Closure $next): mixed
```

lancepioch: Import please

app/Filament/Components/Forms/Fields/TurnstileCaptcha.php
```diff
@@ -0,0 +1,26 @@
...
+
+        $this->required();
+
+        $this->after(fn (TurnstileCaptcha $component) => $component->rule(new ValidTurnstileCaptcha()));
```
lancepioch, line 24: This would be the issue, the `after` method takes a $date for this.

lancepioch (changes requested):

bootstrap/app.php
Boy132: > Nothing calls for it so i don't see the point in making an alias

Then this is a bug, this middleware should be called somewhere.

resources/views/filament/components/turnstile-captcha.blade.php
Boy132: There is no longer a config.
But like I said all this env variables will be moved to the database eventually.

lancepioch (approved):

PR #1069, opened 2025-03-08
Add hint about overhead when using memory limit

notAreYouScared:
Add hint about overhead as it's not documented anywhere except in the wings source code.

### DIFF
--- a/lang/en/admin/server.php
+++ b/lang/en/admin/server.php
@@ -36,6 +36,7 @@ return [
     'disabled' => 'Disabled',
     'memory' => 'Memory',
     'memory_limit' => 'Memory Limit',
+    'memory_helper' => 'Wings will add overhead to this value when creating the container. <=2048MB = +10%, <=4069MB = +15%, anything else = +5%',
     'disk' => 'Disk Space',
     'disk_limit' => 'Disk Space Limit',
     'advanced_limits' => 'Advanced Limits',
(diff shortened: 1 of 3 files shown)


lang/en/admin/server.php
```diff
@@ -36,6 +36,7 @@
     'disabled' => 'Disabled',
     'memory' => 'Memory',
     'memory_limit' => 'Memory Limit',
+    'memory_helper' => 'Wings will add overhead to this value when creating the container. <=2048MB = +10%, <=4069MB = +15%, anything else = +5%',
```
Boy132, line 39: I would keep the text neutral. Especially because the overhead can be individually set in the wings config.

Something like:
```suggestion
    'memory_helper' => 'Wings will add overhead to this value when creating the container to make sure it doesn't starve out when using max memory.',
```

notAreYouScared: 99.8% of users have no idea it exists... let alone an actual config setting for it.

lang/en/admin/server.php
```diff
@@ -36,6 +36,7 @@
     'disabled' => 'Disabled',
     'memory' => 'Memory',
     'memory_limit' => 'Memory Limit',
+    'memory_helper' => 'Wings will add overhead to this value when creating the container to make sure it doesn\'t starve out when using max memory.',
```
lancepioch, line 39: Rewrite

```suggestion
    'memory_helper' => 'While the Server is being created, additional memory will be temporarily used to help prevent issues.',
```

Boy132: That description is wrong. The overhead is added to the normal container, not the install container.

rmartinoscar (approved):

PR #1071, opened 2025-03-08
Add url `Repeater` to `ImportEggHeaderAction`

rmartinoscar:
Multiple
image
image
image
image

Single
image

### DIFF
--- a/app/Filament/Components/Actions/ImportEggAction.php
+++ b/app/Filament/Components/Actions/ImportEggAction.php
@@ -45,15 +45,16 @@ class ImportEggAction extends Action
                     Tab::make(trans('admin/egg.import.url'))
                         ->icon('tabler-world-upload')
                         ->schema([
-                            Repeater::make('url')
-                                ->label('Egg')
+                            Repeater::make('urls')
+                                ->itemLabel(fn (array $state) => $state['url'] ? str($state['url'])->afterLast('/')->before('.json') : null)
                                 ->hint(trans('admin/egg.import.url_help'))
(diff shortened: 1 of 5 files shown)


app/Filament/Components/Actions/ImportEggAction.php
```diff
@@ -44,27 +45,39 @@ protected function setUp(): void
...
-                                ->default(fn (Egg $egg) => $egg->update_url)
-                                ->label(trans('admin/egg.import.url'))
+                            Repeater::make('url')
+                                ->label('Egg')
                                 ->hint(trans('admin/egg.import.url_help'))
-                                ->url(),
+                                ->addActionLabel(trans('admin/egg.import.add_url'))
+                                ->unique()
+                                ->reorderable(false)
+                                ->schema([
```
Boy132, line 54: Make the repeater [simple](https://filamentphp.com/docs/3.x/forms/fields/repeater#simple-repeaters-with-one-field)

Boy132 (dismissed):

notAreYouScared:
If you add 5 urls, and the 2nd one fails, do the rest get imported? Or do they fail after that?

rmartinoscar:
> If you add 5 urls, and the 2nd one fails, do the rest get imported? Or do they fail after that?

They all fail i'll edit the catch block and make sure you can't have multiple Files & Urls when using the import action on an existing egg

JoanFo1456:
Good job Martin 👍

lancepioch (approved):

PR #1075, opened 2025-03-09
Fix null `Node` `Stats`

rmartinoscar:
Happened if you were talking to the wrong instance of wings (token missmatch) or any other reachable website that wouldn't timeout.

### DIFF
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -364,6 +364,10 @@ class Node extends Model implements Validatable
         ];
 
         try {
+            (new DaemonConfigurationRepository())
+                ->setNode($this)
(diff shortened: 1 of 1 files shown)


app/Models/Node.php
```diff
@@ -364,6 +364,10 @@ public function statistics(): array
         ];
 
         try {
+            (new DaemonConfigurationRepository())
+                ->setNode($this)
+                ->getSystemInformation();
+
```
lancepioch, line 370:
```suggestion
            $this->systemInformation();

```

lancepioch (changes requested):

lancepioch (approved):

PR #1080, opened 2025-03-10
Fix `FindViableNodeService` to actually filter `Tags`

thibaultjunin:
Hi,

I found out that tags (or locations_id) were ignored in the endpoint `/nodes/deployable`.

This PR fixes the issue by checking if the collection is not empty.

### DIFF
--- a/app/Services/Deployment/FindViableNodesService.php
+++ b/app/Services/Deployment/FindViableNodesService.php
@@ -29,7 +29,7 @@ class FindViableNodesService
             ->get();
 
         return $nodes
-            ->filter(fn (Node $node) => !tags || collect($node->tags)->intersect($tags)->isNotEmpty())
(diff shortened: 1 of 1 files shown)


app/Services/Deployment/FindViableNodesService.php
```diff
@@ -29,7 +29,7 @@ public function handle(int $memory = 0, int $disk = 0, int $cpu = 0, array $tags
             ->get();
 
         return $nodes
-            ->filter(fn (Node $node) => !$tags || collect($node->tags)->intersect($tags))
+            ->filter(fn (Node $node) => !$tags || !collect($node->tags)->intersect($tags)->isEmpty())
```
rmartinoscar, line 32:
```suggestion
            ->filter(fn (Node $node) => collect($node->tags)->intersect($tags)->isNotEmpty())
```
`$tags` can't be `null` or `false` since its an `array` even when not provided

Boy132 (approved):

PR #1104, opened 2025-03-16
Add App Logo

rmartinoscar:
image
image
image
image
image

I didn't want to flood but Auth logo works the same way nav does, if the logo is null its gonna print app.name and if its invalid it will print `Pelican logo` (app.name + 'logo')


Boy132 (dismissed):

Boy132 (dismissed):
Please make a separate PR for the footer.

Boy132 (approved):

PR #1119, opened 2025-03-18
Add UI for server transfers

### LINKED ISSUE #289: Implement Server Transfers
This would add a new tab or section similar to the old interface

Boy132:
Closes #289

Allocations probably need more validation but for now I think it's fine.
In the future we should replace the Selects with a Repeater that shows all old allocations and options to "connect" them to new allocations.

_Note: I have only tested the UI itself, not the actual transfer because I don't have a setup with more than one node._

### DIFF
--- a/app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
+++ b/app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
@@ -861,7 +861,7 @@ class EditServer extends EditRecord
                                                 Forms\Components\Actions::make([
                                                     Action::make('transfer')
                                                         ->label(trans('admin/server.transfer'))
-                                                        ->disabled(fn () => Node::count() <= 1)
+                                                        ->disabled(fn (Server $server) => Node::count() <= 1 || $server->isInConflictState())
                                                         ->modalheading(trans('admin/server.transfer'))
                                                         ->action(function (TransferServerService $transfer, Server $server, $data) {
                                                             try {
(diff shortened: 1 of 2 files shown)


SirEndii:
I can test the transfer for you but I first need to build the docker image.

app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
```diff
@@ -858,34 +861,53 @@ public function form(Form $form): Form
                                                 Forms\Components\Actions::make([
                                                     Action::make('transfer')
                                                         ->label(trans('admin/server.transfer'))
-                                                        // ->action(fn (TransferServerService $transfer, Server $server) => $transfer->handle($server, []))
-                                                        ->disabled() //TODO!
-                                                        ->form([ //TODO!
-                                                            Select::make('newNode')
-                                                                ->label('New Node')
+                                                        ->disabled(fn () => Node::count() <= 1)
```
rmartinoscar, line 864: Disable it when a transfer has already been asked and possibly display new node name.

rmartinoscar:
I created a server with `0.0.0.0:10000` then transferred it to my other node with allocation `0.0.0.0:2000` now the server has both
image
but Node is still the first one, the server `isInConflictState` but neither wings received any request on the transfer endpoint

app/Filament/Admin/Resources/ServerResource/Pages/EditServer.php
Boy132: For now I added a check on the conflict state.
Later we can add a "Cancel Transfer" button.

rmartinoscar (approved):
We should fix the backend before merging this so we can push a beta before next year.

lancepioch (approved):

PR #1137, opened 2025-03-20
Pest Test Improvements

pelican-vehikl:


phpunit.xml
lancepioch, line 1: This should be reverted

package.json
lancepioch, line 1: This should be reverted

tests/Feature/SettingsControllerTest.php
```diff
@@ -0,0 +1,183 @@
...
+
+    $server = $server->refresh();
+    expect()->toLogActivities(1)
+        ->and($server->name)->not()->toBe($originalName);
+});
+
+//test('subuser cannot change server name without permission', function () {
+//
+//});
+
```
lancepioch, line 81:
```suggestion
```

PR #1073, opened 2025-03-09
Use composer to copy `.env.example` after its done installing packages.

rmartinoscar:

PR #1138, opened 2025-03-21
Make `Installer` handle `APP_KEY` generation

rmartinoscar:
Needs https://github.com/pelican-dev/panel/pull/1073

### DIFF
--- a/app/Livewire/Installer/PanelInstaller.php
+++ b/app/Livewire/Installer/PanelInstaller.php
@@ -95,6 +95,9 @@ class PanelInstaller extends SimplePage implements HasForms
             // Disable installer
             $this->writeToEnvironment(['APP_INSTALLED' => 'true']);
 
+            // Generate app key if not present
+            Artisan::call('p:environment:setup');
+
             // Run migrations
             $this->runMigrations();
 


Boy132 (commented):
`p:environment:setup` is not only generating the app key but also copies the example .env file.
Both of these steps need to happen before the installer part.

notAreYouScared:
> > `p:environment:setup` is not only generating the app key but also copies the example .env file. Both of these steps need to happen before the installer part.
> 
> #1138
> 
> https://github.com/pelican-dev/panel/blob/4ae0523d44a74a419feaf5e72a9a52ddc882c1ca/composer.json#L84-L86

That only fixes copying the env. Not generating the app key, which is still needed

rmartinoscar:
> That only fixes copying the env. Not generating the app key, which is still needed

Oops it only sets an empty one when running from console should i remove this check or cancel [pelican-dev/panel#1138](https://github.com/pelican-dev/panel/pull/1138) & https://github.com/pelican-dev/docs/pull/123 PR ? 
https://github.com/pelican-dev/panel/blob/50f42759a94a3a10462a7d370d99fb3f2356b719/app/Providers/AppServiceProvider.php#L58-L60

Boy132:
> > That only fixes copying the env. Not generating the app key, which is still needed
> 
> Oops it only sets an empty one when running from console should i remove this check or cancel [pelican-dev/panel#1138](https://github.com/pelican-dev/panel/pull/1138) & [pelican-dev/docs#123](https://github.com/pelican-dev/docs/pull/123) PR ?
> 
> https://github.com/pelican-dev/panel/blob/50f42759a94a3a10462a7d370d99fb3f2356b719/app/Providers/AppServiceProvider.php#L58-L60

Close the PRs.

PR #1139, opened 2025-03-21
Fix server transfer backend

Boy132:
The notify request was wrong, probably due to bad copy pasta when the repositories got refactored.


lancepioch (approved):

rmartinoscar:
Still broken i tried removing the json cast which made wings go further but still the targetNode is stuck at `receiving archive`
https://github.com/pelican-dev/panel/blob/1af18590e27534d0a6fab770d5da6a3753d5a8b3/app/Services/Servers/TransferServerService.php#L28-L36

rmartinoscar (approved):

PR #1144, opened 2025-03-21
Fix 500 when viewing node

### LINKED ISSUE #1143: Undefined array key "disk_total"
### Current Behavior

There is a 500 error when adding a new node.

### Expected Behavior

No error.

### Steps to Reproduce

Add a new Error

### Panel Version

canary (82b0aff)

### Wings Version

does not apply

_No response_

### Docker Image

_No response_

### Error Logs

```bash
Undefined array key "disk_total" (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php) (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php) (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php) (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php) (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php) (View: /var/www/pelican/resources/views/filament/components/node-storage-chart.blade.php)
```

notAreYouScared:
Catches 500 when connecting to a node, that's already a wings server? Not sure how to explain it.

#1075 Should've fixed it, but it still exists.

Closes #1143

### DIFF
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -374,12 +374,10 @@ class Node extends Model implements Validatable
 
             if ($data['memory_total']) {
                 return $data;
-            } else {
-                return $default;
             }
         } catch (Exception) {
-            return $default;
         }
+        return $default;
     }
 
     /** @return string[] */
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -367,11 +367,17 @@ class Node extends Model implements Validatable
         try {
             $this->systemInformation();
 
-            return Http::daemon($this)
+            $data = Http::daemon($this)
                 ->connectTimeout(1)
                 ->timeout(1)
                 ->get('/api/system/utilization')
-                ->json() ?? $default;
+                ->json();
+
+            if ($data['error']) {
+                return $default;
+            }
+
+            return $data ?? $default;
         } catch (Exception) {
             return $default;
         }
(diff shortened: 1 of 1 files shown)


app/Models/Node.php
```diff
@@ -367,11 +367,17 @@ public function statistics(): array
         try {
             $this->systemInformation();
```
rmartinoscar, line 368:
```suggestion
```
This is useless then

app/Models/Node.php
```diff
@@ -367,11 +367,17 @@ public function statistics(): array
...
+                ->json();
+
+            if ($data['memory_total']) {
+                return $data;
+            } else {
+                return $default;
+            }
         } catch (Exception) {
             return $default;
         }
```
rmartinoscar, line 383:
```suggestion
            if ($data['memory_total']) {
                return $data;
        } catch (Exception) {
        }
        return $default;
```

notAreYouScared: If theres no exception... what are we returning? Nothing? We'd be right back to square on where somthing responds.... and its in json...

rmartinoscar: > If theres no exception... what are we returning? Nothing? We'd be right back to square on where somthing responds.... and its in json...
That's why i moved the `return $default` out of the catch

notAreYouScared: Missed that, my bad

lancepioch (approved):

PR #1145, opened 2025-03-21
Refactor Node Stats

notAreYouScared:
They've been problematic from day one... no need to make them a scheduled job


lancepioch:
I agree that it's not a great solution, but we can't poll every single second either, that's even worse.

notAreYouScared:
> I agree that it's not a great solution, but we can't poll every single second either, that's even worse.

How is 1 second worse than a webhook doing it every second? This isn't a job that runs and only runs on page view.

notAreYouScared:
I still feel we should poll when they're viewing the page.... What's the point of being able to view the last 30 or whatever seconds before you view the page?

Boy132 (approved):

PR #1147, opened 2025-03-22
Remove cursor not allowed rule

PalmarHealer:
In the UI for the permissions (role creation and user invitation) all the checkbox have the:
`cursor: not-allowed;`
rule applied even though it is interactable. It tried to create roles and the seems to be purly visual.

I don't know if this is referenced/used by other elements but since I could not find any other usages I think it can be removed

### DIFF
--- a/public/css/filament/server/console.css
+++ b/public/css/filament/server/console.css
@@ -14,6 +14,10 @@
     padding-right: 10px;
 }
 
+#sendCommand:read-only {
+    cursor: not-allowed;
+}
+
 ::-webkit-scrollbar {
     background: none;
     width: 14px;
--- a/public/css/filament/server/console.css
+++ b/public/css/filament/server/console.css
@@ -14,10 +14,6 @@
     padding-right: 10px;
 }
 
-input:read-only {
-    cursor: not-allowed;
-}
-
 ::-webkit-scrollbar {
     background: none;
     width: 14px;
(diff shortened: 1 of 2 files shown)


PalmarHealer:

public/css/filament/server/console.css
```diff
@@ -14,10 +14,6 @@
     padding-right: 10px;
 }
 
-input:read-only {
-    cursor: not-allowed;
-}
-
```
rmartinoscar, line 20:
```suggestion
#sendCommand:read-only {
    cursor: not-allowed;
}
```
Add `id="sendCommand"` bellow class to
https://github.com/pelican-dev/panel/blob/23f39acd4eab3f70e855f48c13adfea61096592a/resources/views/filament/components/server-console.blade.php#L20-L21

rmartinoscar (changes requested):
Good catch !
We should only apply this to the `SendCommand` input.

rmartinoscar (approved):

resources/views/filament/components/server-console.blade.php
```diff
@@ -18,6 +18,7 @@
                 icon="tabler-chevrons-right"
             />
             <input
+                id="sendCommand"
```
lancepioch, line 21:
```suggestion
                id="send-command"
```

public/css/filament/server/console.css
```diff
@@ -14,7 +14,7 @@
     padding-right: 10px;
 }
 
-input:read-only {
+#sendCommand:read-only {
```
lancepioch, line 17:
```suggestion
#send-command:read-only {
```

PR #1157, opened 2025-03-24
Show different emptyStateHeading when `activeTab` is `my` in `ListServers`

rmartinoscar:
image
image
image


JoanFo1456:
I think you can change the message of all servers to something like "You don't own a server and you don't have access to any server"

notAreYouScared (approved):

PR #1165, opened 2025-03-26
Remove duplicated `activity.read` & use relations

rmartinoscar:


app/Filament/Server/Resources/UserResource.php
```diff
@@ -264,6 +261,7 @@ public static function table(Table $table): Table
                                                             })
                                                             ->bulkToggleable()
                                                             ->label('')
+                                                            ->columns(2)
```
Boy132, line 264: Why the `->columns(2)` everywhere?

notAreYouScared: I'm fine with this. The Invite modal has the permissions in a 2 column layout already, this is just making the edit match the create 👍

notAreYouScared: image
image

PR #1173, opened 2025-03-27
Improve alert banner fetching

### LINKED ISSUE #1152: Improve `AlertBanner` fetching
Currently the `AlertBannerContainer` is a livewire component that runs `pullFromSessions` every second. This causes a lot of unnecessary requests.
Ideally this should be switched to an event based system.

https://github.com/pelican-dev/panel/blob/61cbe5465f6a18959ec28ef023a86f3f94233287/resources/views/livewire/alerts/alert-banner-container.blade.php#L1-L5
https://github.com/pelican-dev/panel/blob/61cbe5465f6a18959ec28ef023a86f3f94233287/app/Livewire/AlertBannerContainer.php#L19-L25

Boy132:
Closes #1152

Use an event instead of pulling every second. Based on [filaments notifications](https://github.com/filamentphp/filament/blob/fed5c56ae7dca0454035f77be8a811dbcbf9dbd8/packages/notifications/src/NotificationsServiceProvider.php#L39-L53).

### DIFF
--- a/app/Filament/Server/Pages/Console.php
+++ b/app/Filament/Server/Pages/Console.php
@@ -38,10 +38,10 @@ class Console extends Page
         try {
             $server->validateCurrentState();
         } catch (ServerStateConflictException $exception) {
-            AlertBanner::make()
-                ->warning()
+            AlertBanner::make('server_conflict')
                 ->title('Warning')
                 ->body($exception->getMessage())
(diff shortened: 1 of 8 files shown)


notAreYouScared (approved):
Appears to work :)

lancepioch (approved):

PR #1177, opened 2025-03-27
add a all in one docker compose with example.env and README.md

TheDoDoo:
Hello everyone,
This PR introduces an all-in-one `docker-compose.yml` that directly includes the `database` and `Redis`. Additionally, there is a `.env` file where all settings can be configured.

The docker/README.md explains how to install the panel using docker-compose and the .env file.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,105 @@
+
+# Pelican Panel - Docker Image
+
+## Requirements
+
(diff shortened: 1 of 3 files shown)


TheDoDoo:
¯\_(ツ)_/¯

lancepioch:
Thanks for the contribution! While Pelican natively supports Redis and other relational databases (like mysql and mariadb), the majority of our users will have a solid experience without those services. Those services are optional, but may be used for a better experience for people hosting larger and even commercial instances. In that case, we want to leave it up to the individual user to set up their own docker compose files and images.

PR #1181, opened 2025-03-27
Must use `Closure` since `MenuItem` does not leverage Concerns

rmartinoscar:
https://github.com/filamentphp/filament/issues/13342
I hate cache it should have been catched before https://github.com/pelican-dev/panel/pull/1178

### DIFF
--- a/app/Providers/Filament/ServerPanelProvider.php
+++ b/app/Providers/Filament/ServerPanelProvider.php
@@ -43,7 +43,9 @@ class ServerPanelProvider extends PanelProvider
             ->maxContentWidth(config('panel.filament.display-width', 'screen-2xl'))
             ->login(Login::class)
             ->userMenuItems([
-                'profile' => MenuItem::make()->label('Profile')->url(fn () => EditProfile::getUrl(panel: 'app')),
+                'profile' => MenuItem::make()
+                    ->label('Profile')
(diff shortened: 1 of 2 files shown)


app/Providers/Filament/ServerPanelProvider.php
```diff
@@ -43,7 +43,9 @@ public function panel(Panel $panel): Panel
             ->maxContentWidth(config('panel.filament.display-width', 'screen-2xl'))
             ->login(Login::class)
             ->userMenuItems([
-                'profile' => MenuItem::make()->label('Profile')->url(fn () => EditProfile::getUrl(panel: 'app')),
+                'profile' => MenuItem::make()
+                    ->label('Profile')
```
Boy132, line 47:
```suggestion
                    ->label(fn () => trans('filament-panels::pages/auth/edit-profile.label'))
```

Boy132 (changes requested):

app/Providers/Filament/ServerPanelProvider.php
rmartinoscar: Do we translate the Client Area right now ?

Boy132: For such common things that already have filament translation I would say yes.

PR #1190, opened 2025-03-30
Switch inserts to proper creates

lancepioch:


app/Http/Controllers/Api/Remote/ActivityProcessingController.php
```diff
@@ -71,19 +69,17 @@ public function __invoke(ActivityEventRequest $request): void
         }
 
         foreach ($logs as $key => $data) {
-            Assert::isInstanceOf($server = $servers->get($key), Server::class);
+            $server = $servers->get($key);
+            assert($server instanceof Server);
```
rmartinoscar, line 73: Thought we were moving out of helpers ?

lancepioch: assert isn't a helper :)

rmartinoscar (approved):

PR #1192, opened 2025-03-31
Add avatar providers

Boy132:
By default Gravatar, UI Avatars and "Local Avatars" (allow users to upload their own avatar) are available.
Gravatar is the default.

_(requires filament 3.3.7+)_


rmartinoscar:
I force-pushed cause we'll properly deal with Openspout & Filament in https://github.com/pelican-dev/panel/pull/1191

notAreYouScared:
It seems to mostly work except for local, Image uploads, saving, and refreshing the page; nothing happens and the image disappears. Where is this `avatars` folder supposed to be?

PR #1221, opened 2025-04-06
Add Translations from Crowdin

notAreYouScared:


rmartinoscar (changes requested):
`be`, `el`, `fi`, `he`, `hi`, `hr`, `id`, `it`, `ko`, `lv`, `sk`, `sv`, `th`, `vi` & `zh` are incomplete

And if we use them it will just return the raw `admin.apikey.title` instead of a fallback translation.

PR #1222, opened 2025-04-06
Update original migrations with nullable

lancepioch:
The recommended solution is to use migration squashing, but it's also safe to just explicitly declare these migration columns as nullable (as they were assumed to be beforehand anyways).

### DIFF
--- a/database/migrations/2024_03_12_154408_remove_nests_table.php
+++ b/database/migrations/2024_03_12_154408_remove_nests_table.php
@@ -11,7 +11,7 @@ return new class extends Migration
     public function up(): void
     {
         Schema::table('eggs', function (Blueprint $table) {
-            $table->text('tags');
+            $table->text('tags')->nullable();
         });
 
         DB::table('eggs')->update(['tags' => '[]']);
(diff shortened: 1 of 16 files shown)


database/migrations/2024_03_14_055537_remove_locations_table.php
```diff
@@ -10,7 +10,7 @@
     public function up(): void
     {
         Schema::table('nodes', function (Blueprint $table) {
-            $table->text('tags');
+            $table->text('tags')->nullable();
```
rmartinoscar, line 13: Not nullable
https://github.com/pelican-dev/panel/blob/377b3f170da2fda69941a671c035f140ea9e8ea1/app/Models/Node.php#L43

database/migrations/2024_03_12_154408_remove_nests_table.php
```diff
@@ -11,7 +11,7 @@
     public function up(): void
     {
         Schema::table('eggs', function (Blueprint $table) {
-            $table->text('tags');
+            $table->text('tags')->nullable();
```
rmartinoscar, line 14: Not nullable
https://github.com/pelican-dev/panel/blob/377b3f170da2fda69941a671c035f140ea9e8ea1/app/Models/Egg.php#L49

PR #1224, opened 2025-04-08
CSS calls for lowercase

rmartinoscar:
image


rmartinoscar:
> Remove bold, never used by the panel...

It will be in https://github.com/pelican-dev/panel/tree/charles/change-font

Boy132:
> > Remove bold, never used by the panel...
> 
> It will be in https://github.com/pelican-dev/panel/tree/charles/change-font

This can be added to charles' branch instead of being it's own PR then.

PR #1231, opened 2025-04-10
Allow to assign nodes to roles (node ownership)

### LINKED ISSUE #613: Assign Nodes to Roles
### Discussed in https://github.com/pelican-dev/panel/discussions/21

<sup>Originally posted by **Oldbear200** April 12, 2024</sup>
Hey please add that admins can assign a node to a user so he can create / manage / delete services with just that node.
Would love that</div>

Scope roles to specific nodes. (many to many)

Boy132:
Closes #613

Allows to assign certain nodes to roles. If a user has a role that only has node A assigned they can only manage node A and servers on node A. (still needs the according permissions like create, update etc.) Also applies to mounts and database hosts.

Example:
grafik
grafik

Manager A:
grafik

Manager B:
grafik

Admin: (Non root admin)
grafik

Note: the assigned role nodes also do "stack". Meaning if you give a user both the manager A and the manager B roles they will get access to both nodes.


app/Models/Node.php
```diff
@@ -268,6 +270,11 @@ public function databaseHosts(): BelongsToMany
         return $this->belongsToMany(DatabaseHost::class);
     }
 
+    public function roles(): HasManyThrough
+    {
+        return $this->hasManyThrough(Role::class, RoleNode::class, 'node_id', 'id', 'id', 'role_id');
```
lancepioch, line 275: Since we're using the expected column names, can we shorten this?

```suggestion
        return $this->hasManyThrough(Role::class, RoleNode::class);
```

app/Models/Role.php
```diff
@@ -125,4 +128,9 @@ public static function getRootAdmin(): self
 
         return $role;
     }
+
+    public function nodes(): BelongsToMany
+    {
+        return $this->belongsToMany(Node::class, RoleNode::class, 'role_id', 'node_id', 'id', 'id');
```
lancepioch, line 134: Since we're using the expected column names, can we shorten this?

```suggestion
        return $this->belongsToMany(Node::class, RoleNode::class);
```

app/Models/RoleNode.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class RoleNode extends Model
```
lancepioch, line 7: This is only a Pivot table, correct? Then this is more correct:

```suggestion
class RoleNode extends Pivot
```

app/Models/RoleNode.php
```diff
@@ -0,0 +1,14 @@
...
+{
+    protected $table = 'role_node';
```
lancepioch, line 9: Should be alphabetical

```suggestion
    protected $table = 'node_role';
```

database/migrations/2025_04_08_113556_create_role_node_table.php
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('role_node', function (Blueprint $table) {
```
lancepioch, line 14:
```suggestion
        Schema::create('node_role', function (Blueprint $table) {
```

lancepioch (changes requested):
Awesome, great work, just have a couple nitpicks that should make things a little cleaner!

app/Models/RoleNode.php
rmartinoscar: I believe he followed what we'v had for MountNode

https://github.com/pelican-dev/panel/blob/98d8510f11f3d007bb46b57519a09595a8216e22/app/Models/MountNode.php#L7
https://github.com/pelican-dev/panel/blob/98d8510f11f3d007bb46b57519a09595a8216e22/app/Models/Node.php#L242-L245

app/Models/RoleNode.php
```diff
@@ -0,0 +1,14 @@
...
+
+    protected $primaryKey = null;
+
+    public $incrementing = false;
```
rmartinoscar, line 13:
```suggestion
```
Unneeded once this is a `Pivot`

lancepioch (approved):

rmartinoscar:
I'd make `accessibleNodes` a QueryBuilder so we can use it in the controller like for `accessibleServers`

https://github.com/rmartinoscar/panel/tree/boy132/node-ownership

app/Models/Node.php
```diff
@@ -268,6 +270,11 @@ public function databaseHosts(): BelongsToMany
         return $this->belongsToMany(DatabaseHost::class);
     }
 
+    public function roles(): HasManyThrough
+    {
+        return $this->hasManyThrough(Role::class, NodeRole::class);
```
rmartinoscar, line 275:
```suggestion
        return $this->hasManyThrough(Role::class, NodeRole::class, 'node_id', 'id', 'id', 'role_id');
```
Default Pivot don't have `id` so we have to specify them.

I'd make `accessibleNodes` a QueryBuilder so we can use it in the controller like for `accessibleServers`
https://github.com/rmartinoscar/panel/tree/boy132/node-ownership

rmartinoscar (changes requested):

notAreYouScared:
Conflicts <3. 
Does this just need to be tested?

Boy132:
> Does this just need to be tested?

Yep. :)

rmartinoscar (approved):

notAreYouScared (approved):

PR #1232, opened 2025-04-10
Cleanup CI

lajczi:


lajczi:

.github/workflows/docker.yaml
```diff
@@ -134,32 +127,30 @@ jobs:
...
-          platforms: 'linux/amd64,linux/arm64'
-          build-args: |
-            VERSION=${{ steps.build_info.outputs.version_tag }}
+          platforms: linux/amd64,linux/arm64
+          build-args: VERSION=${{ steps.build_info.outputs.version_tag }}
           labels: ${{ steps.docker_meta.outputs.labels }}
           tags: ${{ steps.docker_meta.outputs.tags }}
-          cache-from: type=gha,scope=tagged${{ matrix.os }}
-          cache-to: type=gha,scope=tagged${{ matrix.os }},mode=max
+          cache-from: type=gha,scope=release
```
QuintenQVD0, line 141: This can't work. As else they would write over eachother cache

rmartinoscar:
`--no-suggest` was here for a reason.
Lowest nodejs version we support is 18.
We use matrices since we run multiple workflows using different php versions so these must stay.
We use `ubuntu-22.04` cause `ubuntu-latest` is `ubuntu-24.04`.

PR #1236, opened 2025-04-11
chore: Refactor `Mount`s

rmartinoscar:
Move `mount_node`, `server_mount` & `egg_mount` into a single `model_has_mounts` table

!eggs && nodes = available to all servers with associated nodes regardless of eggs 
eggs && !nodes = available to all servers with associated eggs regarless of nodes
!eggs && !nodes = available to all servers regardless of eggs or nodes


lancepioch (dismissed):

database/migrations/2025_04_12_034652_move_to_model_has_mounts.php
```diff
@@ -0,0 +1,93 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('model_has_mounts', function (Blueprint $table) {
```
lancepioch, line 15:
```suggestion
        Schema::create('mountables', function (Blueprint $table) {
```

app/Models/Mount.php
```diff
@@ -102,24 +102,24 @@ protected function casts(): array
     /**
      * Returns all eggs that have this mount assigned.
      */
-    public function eggs(): BelongsToMany
+    public function eggs(): MorphToMany
     {
-        return $this->belongsToMany(Egg::class);
+        return $this->morphedByMany(Egg::class, 'model', 'model_has_mounts');
```
lancepioch, line 107:
```suggestion
        return $this->morphedByMany(Egg::class, 'mountable');
```

app/Models/Egg.php
```diff
@@ -289,6 +290,11 @@ public function getInheritFileDenylistAttribute(): ?array
         return $this->configFrom->file_denylist;
     }
 
+    public function mounts(): MorphToMany
+    {
+        return $this->morphToMany(Mount::class, 'model', 'model_has_mounts');
```
lancepioch, line 295:
```suggestion
        return $this->morphToMany(Mount::class, 'mountable');
```

lancepioch (changes requested):

database/migrations/2025_04_12_034652_move_to_model_has_mounts.php
rmartinoscar: Im really not a fan of this, then we need to change
`egg` > eggable
`allocation` > allocable
`role` > rollable
`permission` > permissible

I much rather keep it `model_has_mounts` since if follows `model_has_permissions`, `model_has_roles`, `role_has_permissions` 

> eventhough Spatie is wrong

lancepioch: You wouldn't change the names for those, only from Model -> Mount (mountable). The rest would stay the same name.

PR #1245, opened 2025-04-12
Use recommended PHP `8.4` for Docker

rmartinoscar:


lajczi:
You can also update https://github.com/pelican-dev/panel/blob/main/Dockerfile#L7C1-L8C1

rmartinoscar:
> You can also update [`main`/Dockerfile#L7C1-L8C1](https://github.com/pelican-dev/panel/blob/main/Dockerfile#L7C1-L8C1)

This is commented out since we don't use the stage system for cache.

lancepioch (approved):

PR #1248, opened 2025-04-13
Remove `NavigationGroups` for Admin Navbar

rmartinoscar:
Before
image
image

After
image
image


lancepioch (approved):

Boy132:
Can't we make the groups conditional? When using sidebar use groups, otherwise only use `Advanced`.
This way it looks nice on both settings.

PR #1252, opened 2025-04-13
Fixing of -> Undefined array key "customization"

Justman100:


Justman100:

Justman100:
recheck

QuintenQVD0:
You can only have that if you did not migrate the db?

Justman100:
> You can only have that if you did not migrate the db?

We've had Pelican running for a while! In fact, I only saw the error because I wanted to create a user API key

QuintenQVD0:
Please run `php artisan migrate --seed --force` withoud this applied and see if it works or if you get any migration errors.

Boy132:
Like Quinten said this looks like something is wrong with your migrations. I can't reproduce the error.
If you are able to reproduce it again you can open an issue with an actual description. Or visit the Discord first so this can be verified.

PR #1253, opened 2025-04-13
Fix storage graph not rendering correctly

LetterN:
Use round instead of `Number::format` since that adds a comma once it reaches 1k which breaks chartjs render.


LetterN:

LetterN:
recheck

Boy132 (approved):
Thanks for your contribution!

rmartinoscar (approved):

PR #1258, opened 2025-04-14
Fix activity log query

### LINKED ISSUE #1024: Activity Log metadata view modal 4-5 entries behind selected activity
When viewing the activity log, and wanting to view the metadata of an activity, the popup modal will display the metadata for 4-5 logs behind the one you selected.

Refer to below video

https://github.com/user-attachments/assets/9eda7512-6943-4af0-af78-8e9fc4ad55ee

Boy132:
Closes #1024

For some reason it didn't like using `$server->activity()` so now the query is built manually.

### DIFF
--- a/app/Filament/Server/Resources/ActivityResource.php
+++ b/app/Filament/Server/Resources/ActivityResource.php
@@ -30,8 +30,7 @@ class ActivityResource extends Resource
         /** @var Server $server */
         $server = Filament::getTenant();
 
-        return $server->activity()
-            ->getQuery()
+        return ActivityLog::whereHas('subjects', fn (Builder $query) => $query->where('subject_id', $server->id))
             ->whereNotIn('activity_logs.event', ActivityLog::DISABLED_EVENTS)
             ->when(config('activity.hide_admin_activity'), function (Builder $builder) use ($server) {
                 // We could do this with a query and a lot of joins, but that gets pretty


app/Filament/Server/Resources/ActivityResource.php
```diff
@@ -30,8 +30,7 @@ public static function getEloquentQuery(): Builder
         /** @var Server $server */
         $server = Filament::getTenant();
 
-        return $server->activity()
-            ->getQuery()
+        return ActivityLog::whereHas('subjects', fn (Builder $query) => $query->where('subject_id', $server->id))
```
rmartinoscar, line 33:
```suggestion
        return QueryBuilder::for($server->activity())
```
https://github.com/pelican-dev/panel/blob/b44411208573132c8be73dc2959ac90130442ddb/app/Http/Controllers/Api/Client/Servers/ActivityLogController.php#L34

Boy132: That's effectively the same as `$server->activity()->getQuery()`

rmartinoscar, line 33:
```suggestion
        $query = parent::getEloquentQuery();

        /** @var Server $server */
        $server = Filament::getTenant();

        return $query->whereHas('subjects', fn (Builder $query) => $query->where('subject_id', $server->id))
```

Boy132: No, can't use `parent::getEloquentQuery()` because that fails with an error. (`The model [App\Models\ActivityLog] does not have a relationship named [server]`)
That's why we overwrite the method in the first place.

lancepioch (approved):

PR #1259, opened 2025-04-14
Chunk `Sushi` inserts based on rows count

### LINKED ISSUE #1254: 500 Server Error: SQLite too many SQL variables when accessing folder with many files
### Current Behavior

SQLSTATE[HY000]: General error: 1 too many SQL variables
...
insert into "files" (...) values (...), (...), ...

### Expected Behavior

I need the folder to load properly without issue. The issue with the folder having too many files inside it, is causing an error 500 for myself and others when trying to access this folder. I need the error resolved.

This used to work fine in a previous update, but seems to have stopped working in a recent update.

### Steps to Reproduce

Create a folder which then contains more than the SQLLite variables limit and try accessing the folder in the panel via the "Files" section of the individual server hosted.

### Panel Version

1.0.0-beta19

### Wings Version

v1.0.0-beta11

_No response_

### Docker Image

_No response_

### Error Logs

```bash
SQLSTATE[HY000]: General error: 1 too many SQL variables
...
insert into "files" (...) values (...), (...), ...
```

rmartinoscar:
Closes #1254
image

Tested using, you exceed memory limit going higher then this
```sh
for i in {1..36000}; do touch $EPOCHREALTIME; done
```

### DIFF
--- a/app/Models/File.php
+++ b/app/Models/File.php
@@ -33,6 +33,8 @@ class File extends Model
 
     protected $keyType = 'string';
 
+    protected int $sushiInsertChunkSize = 100;
+
     public const ARCHIVE_MIMES = [
         'application/vnd.rar', // .rar
         'application/x-rar-compressed', // .rar (2)
@@ -167,7 +169,7 @@ class File extends Model
                 throw new Exception($contents['error']);
             }
 
-            return array_map(function ($file) {
+            $rows = array_map(function ($file) {
                 return [
                     'name' => $file['name'],
                     'created_at' => Carbon::parse($file['created'])->timezone('UTC'),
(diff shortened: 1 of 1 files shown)


app/Models/File.php
```diff
@@ -181,6 +183,10 @@ public function getRows(): array
                     'mime_type' => $file['mime'],
                 ];
             }, $contents);
+
+            $this->sushiInsertChunkSize = min((int) floor(999 / count($this->getSchema())), count($rows));
```
Boy132, line 187: Is there a reason you do this crazy math instead of just setting the chunk size to a fixed value? Like 50 or something?

rmartinoscar: The more you chunk it the lower the loading so i rather only lower it when needed

Boy132 (approved):

PR #1263, opened 2025-04-15
Add back `behind_proxy` to ui

Boy132:
grafik

### DIFF
--- a/app/Filament/Admin/Resources/NodeResource/Pages/CreateNode.php
+++ b/app/Filament/Admin/Resources/NodeResource/Pages/CreateNode.php
@@ -182,7 +182,33 @@ class CreateNode extends CreateRecord
                                     'http' => 'tabler-lock-open-off',
                                     'https' => 'tabler-lock',
                                 ])
-                                ->default(fn () => request()->isSecure() ? 'https' : 'http'),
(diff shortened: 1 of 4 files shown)


lancepioch:
Can this be a hidden checkbox instead?

Boy132:
> Can this be a checkbox instead?

It was a separate thing at first but I didn't like that. I think as one combined elements it makes it easier for users to understand and configure. It happened way too often on pterodactyl that users selected "behind proxy" when not needed.
image

rmartinoscar:
If we add this back i think we should also add a reverse proxy port field that would be what the user uses to contact wings & daemon_listen that would be the port used to communicate between wings & the reverse proxy so you don't have to manually edit the config.yml

We could also add a ToggleButton to not sync updates from the panel by setting `ignore_panel_config_updates` in `$node->getConfiguration()`

lancepioch:
@Boy132 Can you put the 3rd button under the first 2 then? I think that'd look a little nicer then.

Boy132:
> @Boy132 Can you put the 3rd button under the first 2 then? I think that'd look a little nicer then.

Like this? Looks a bit bulky.
grafik

notAreYouScared (approved):

PR #1271, opened 2025-04-17
Add back Egg Features

### LINKED ISSUE #756: Egg feature modals are missing in new client area
The old client area has some modals that display when it finds specific strings in a console line. This should also be implemented in the new client area.

https://github.com/pelican-dev/panel/tree/release/v1.0.0-beta17/resources/scripts/components/server/features

- Minecraft EULA
- GSL Token
- Java Version
- PID Limit
- Steam Disk Space

pelican-vehikl:
Closes #756 
Requires https://github.com/pelican-dev/wings/pull/87


app/Extensions/Features/FeatureProvider.php
```diff
@@ -0,0 +1,51 @@
...
+            return;
+        }
+
+        static::$providers[$this->getId()] = $this;
+    }
+
+    abstract public function getId(): string;
+
+    /**
+     * A matching subset string (case-sensitive) from the console output
```
Boy132, line 44: Is there a reason this isn't case insensitive?

app/Filament/Server/Pages/Console.php
```diff
@@ -47,6 +53,38 @@ public function mount(): void
...
+    }
+
+    #[On('mount-feature')]
+    public function mountFeature(string $data): void
+    {
+        $data = json_decode($data);
+        $feature = data_get($data, 'key');
+        $line = data_get($data, 'line');
+
+        Activity::event('server:feature')
```
Boy132, line 73: Why is there an activity log for this?!

resources/views/filament/components/server-console.blade.php
```diff
@@ -1,3 +1,4 @@
+@php use App\Features\CustomModal; @endphp
```
Boy132, line 1: Leftover?

```suggestion
```

app/Extensions/Features/FeatureProvider.php
lancepioch: Yeah, I wanted it to be case sensitive because console output is finicky.

app/Filament/Server/Pages/Console.php
lancepioch: I think we can remove it.

Paging @rmartinoscar

app/Extensions/Features/GSLToken.php
```diff
@@ -0,0 +1,122 @@
...
+        return 'gsltoken';
+    }
+
+    public function getAction(): Action
+    {
+        /** @var Server $server */
+        $server = Filament::getTenant();
+
+        /** @var ServerVariable */
+        $serverVariable = $server->serverVariables()->where('env_variable', 'STEAM_ACC');
```
pelican-vehikl, line 46:
```suggestion
        /** @var ServerVariable $serverVariable */
        $serverVariable = $server->serverVariables()->where('env_variable', 'STEAM_ACC')->first();
```

lancepioch (approved):

PR #1281, opened 2025-04-21
Resolve issue with avatar storage

parkervcp:
This resolves the issue with getting avatar storage working

updates the entrypoint to create the `pelican-data/storage` folder on start.

Adds a dev dockerfile to build locally instead of needing to update the standard dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,7 +81,7 @@ RUN chown root:www-data ./ \
     && ln -s /pelican-data/.env ./.env \
     && ln -s /pelican-data/database/database.sqlite ./database/database.sqlite \
     && ln -sf /var/www/html/storage/app/public /var/www/html/public/storage \
-    && ln -s  /pelican-data/storage/ /var/www/html/storage/app/public/avatars \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,16 +1,9 @@
...
-# FROM --platform=$TARGETOS/$TARGETARCH php:8.4-fpm-alpine as base
-
-# ADD --chmod=0755 https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
-
-# RUN install-php-extensions bcmath gd intl zip opcache pcntl posix pdo_mysql pdo_pgsql
-
-# RUN rm /usr/local/bin/install-php-extensions
+##
+#  If you want to build this locally you want to run `docker build -f Dockerfile.dev`
+##
```
JustTNE, line 6: I don't think duplicating the entire Dockerfile is a good idea. I'm going to review this in more detail later and come up with a suggestion for this instead.

QuintenQVD0: That's fine for a DEV one

QuintenQVD0 (approved):
Worked localy for me

PR #1282, opened 2025-04-22
feat(console): save command history in session

vinhais:
We can use ``#[Session]`` to have Livewire automatically save the command history in the user's session. 

this will improve the user experience for those who often forget the commands they gave or want to be quick :)


vinhais:

Boy132 (approved):
Good idea, thanks!

rmartinoscar:
Awesome idea ! Even better could we store it in cache ? so every person accessing the server can see the history, and it also works in incognito mode 👌

lancepioch (approved):

lancepioch:
Very good idea, thanks!

PR #1283, opened 2025-04-22
Add back network chart

### LINKED ISSUE #956: Add `Network` statistics to client server site
This still needs to be added to the new front end. 

### Discussed in https://github.com/pelican-dev/panel/discussions/953

<sup>Originally posted by **infnibor** January 26, 2025</sup>
In addition to CPU and Memory, it would be possible to add network consumption statistics as they appear in pterodactyl
image

image</div>

Boy132:
Closes #956

grafik


vinhais:
Why not add the suffixes to the values ​​of the other charts?

RAM would be 1.2GB (or GiB depending on the configuration)
CPU would add the %

infnibor:
In my opinion, showing the time under the chart only spoils the appearance and looks strange, and it is logical that every second there is a result. Relative to the rest of the charts, it deviates in appearance, and it would be better for the charts to have a fairly standardized appearance, so that it would also blend in nicely

notAreYouScared:
Somehow, these are broken? or they're not using the data from the websocket??? the bytes never change therefore there is not "network activity" 



image
image

lancepioch (approved):

notAreYouScared (changes requested):
See previous comment.

notAreYouScared (approved):

notAreYouScared (approved):

rmartinoscar (approved):

PR #1292, opened 2025-04-22
Remove open in new tab for "Open in Admin" option

PalmarHealer:
Removing the open in new tab since both are on filament now. And the tenant: null was function default so not needed aswell

### DIFF
--- a/app/Providers/Filament/ServerPanelProvider.php
+++ b/app/Providers/Filament/ServerPanelProvider.php
@@ -63,7 +63,7 @@ class ServerPanelProvider extends PanelProvider
             ])
             ->navigationItems([
                 NavigationItem::make('Open in Admin')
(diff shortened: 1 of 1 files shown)


notAreYouScared:
I personally would rather it open in a new tab, the reason of them both not being Filament was never the reason it wasn't opened in the same tab

PalmarHealer:
But why is it only there?

When I go to the server from admin it uses the same tab. Same when I use the top right menu. This is the only occourance where it is like that.

It doesn't feel consistent. Personally it does annoys me that I have a new tab each time...

lancepioch (approved):

lancepioch:
Thanks! I agree that a person can middle or right click to open a new tab by their own choice!

Issue #1211, opened 2025-04-03
File browser breaking due to special character in file name


nzxl101:
### Current Behavior

When uploading a file through URL that contains special characters e.g. a quotation mark, the whole file browser will stop working until the file has been deleted through SFTP.

### Expected Behavior

Be able to upload files with special characters by properly escaping/or removing them.

### Steps to Reproduce

1. Go to Files
2. Upload
3. Upload from URL/or local
4. URL input: https://mediafilez.forgecdn.net/files/5317/51/Sneak%27s%20RPG%20Pack%20SERVER.zip

### Panel Version

1.0.0-beta18

### Wings Version

1.0.0-beta10

### Error Logs

```bash

```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue.
- [x] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

rmartinoscar:
Hey please provide more logs, i just tried Upload > From URL and i get

![Image](https://github.com/user-attachments/assets/fca03e85-ef73-4898-872e-4ffe42c48e47)

```
DEBUG: [Apr  3 15:30:53.577] writing remote file to disk path=/Sneak's RPG Pack SERVER.zip server=ed4f088e-18d6-4b36-a691-0f145399d1e5
DEBUG: [Apr  3 15:30:57.126] GET /api/system/utilization client_ip=127.0.0.1 latency=1.232498ms request_id=18a37ae0-152b-44a2-ba12-bf4d0c458a05 status=200
INFO: [Apr  3 15:30:59.822] completed pull of remote file download_id=c4e4dfbb-f53a-4d59-8b04-360a74baecc2 server=ed4f088e-18d6-4b36-a691-0f145399d1e5
DEBUG: [Apr  3 15:31:03.296] GET /api/system/utilization client_ip=127.0.0.1 latency=1.192242ms request_id=e65eded6-473b-4165-a21b-c70529b61a70 status=200
DEBUG: [Apr  3 15:31:03.700] GET /api/servers/ed4f088e-18d6-4b36-a691-0f145399d1e5/files/list-directory?directory=%2F client_ip=127.0.0.1 latency=264.776µs request_id=d75c4ae9-9f09-40db-bc7e-91f3ba916714 status=200
```

nzxl101:
Hi, 
[...]

PR #1295, opened 2025-04-24
Fix file browser breaking with special characters in filenames

Scvtt:
## Summary
Fixed an issue where the file browser would become unusable after uploading or handling files with special characters (like quotes) in their names. The issue was caused by unescaped special characters breaking JavaScript execution in the frontend.

## Technical Changes
1. Modified the File model to handle special characters safely:
   - Added base64 encoding for filenames in internal representation
   - Added separate `display_name` field for UI display
   - Updated schema to support the new data structure

2. Updated ListFiles implementation:
   - Changed table column to use `display_name` for showing filenames
   - Added base64 decode/encode handling for all file operations
   - Updated all file actions to use decoded filenames

3. File operations now properly handle special characters:
   - Rename
   - Move
   - Copy
   - Delete
   - Archive/Unarchive
   - Permissions changes
   - Bulk operations

## Impact
- Users can now upload and manage files with special characters in their names
- File browser remains functional after handling such files
- Original filenames are preserved and displayed correctly
- All file operations work properly with special characters
- Navigation and UI interactions remain smooth

## Testing
Verified working with filenames containing:
- Single quotes
- Double quotes
- Other special characters

## Related Issues
Fixes [Issue #1211](https://github.com/pelican-dev/panel/issues/1211)

### DIFF
--- a/app/Filament/Server/Resources/FileResource/Pages/ListFiles.php
+++ b/app/Filament/Server/Resources/FileResource/Pages/ListFiles.php
@@ -225,7 +225,6 @@ class ListFiles extends ListRecords
                             $decodedName = base64_decode($file->name);
                             $location = rtrim($data['location'], '/');
                             $files = [['to' => join_paths($location, $decodedName), 'from' => $decodedName]];
-
                             $this->getDaemonFileRepository()->renameFiles($this->path, $files);
 
                             $oldLocation = join_paths($this->path, $decodedName);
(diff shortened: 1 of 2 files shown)


Scvtt:

lancepioch:
Thank you @Scvtt for tackling this!

I just want to understand why the urls with UTF-8 aren't working, do you know if it's just because Filament (or Livewire) can't handle them? I think it'd be ideal to try to keep the exact paths in the url because it looks and feels better.

Scvtt:
This happened because Livewire's JavaScript router uses string literals for route parameters, and special characters weren't being properly escaped, causing JavaScript syntax errors.

lancepioch:
@Scvtt I believe we can retain the filenames in the url by using the `id` of the files as the primary key. Can you check this for me?

rmartinoscar:
Superseded by #1323

PR #1296, opened 2025-04-24
Add tzdata to docker image

Sebclem:
This PR simply add tzdata used to change the timezone in the container with the TZ env var:
https://stackoverflow.com/a/68996528


Sebclem:

lancepioch:
Do NOT change the time zone in the container! The machine's time zone should absolutely only ever be UTC.

The main reason we still support setting the time zone is simply for backwards compatibility.

Each user can set their own time zone in their profile settings!

PR #1300, opened 2025-04-25
generalize sponge installation

LetterN:
now uses sponge dl-api (not documented), their download site uses this
**NOTE!!!** when selecting a version, it will pull the latest, not the recommended for that version


rmartinoscar:
Does this include changes that have possibly been made since ?
https://github.com/pelican-eggs/minecraft/tree/main/java/spongevanilla

PR #1308, opened 2025-04-26
Configure docker PHP for production use

JustTNE:
As *strongly* recommended on https://hub.docker.com/_/php, configures PHP for production use. Additionally, the upload file size limit is removed entirely to allow for livewire uploads larger than the default 2 MB that PHP ships with.

This pull request does not build because #1251 introduced a build error, but works perfectly fine if the build error is fixed and is ready to merge as-is.


JustTNE:

lancepioch:
Thanks for the contribution, but having an unlimited post and file upload size on php's side is a security vulnerability.

Issue #1077, opened 2025-03-09
Customizable `user` & `group` id file ownership


kungfoome:
When mounting volumes to pelican, I want to be able to manage the files on there using a particular user. It doesn't seem like I can set the UID or GID for the container. I also have this volume mounted over NFS, which defaults to the user I want. For these reasons, the container won't start since it can't really write or fix the permissions and it's set to uid 86 for www-data it seems like.

I would like the ability to set the UID, GID, and timezone in the image to match the host. linuxserver.io images do a really good job of this and it would be something similar.

parkervcp:
Wings is what manages the users in the containers and you would set that in the config via 

```yaml
system:
  user:
    uid: 999
    gid: 995
```

lancepioch:
Looks like this issue is already resolved.

PR #1309, opened 2025-04-27
Example docker image that can update UID/GID

kungfoome:
I had recently opened an issue to update the UID/GID of the container https://github.com/pelican-dev/panel/issues/1077. This is so that it's easier to mount persistent files and have the proper permissions both within the container and on the host.

While this is not final by any means, it's just an example to show how this could possibly be done. The container does need to be run as root to update permissions, but if that was taken out, you could run the image as a normal user without root.

I also used franken since it included caddy already, just to speed things up a bit.

Nothing runs as root except to change permissions, after which is runs as a normal user. This includes any composer commands as well. The key is also generated using the artisan command and a default admin is created just to get the panel up and running out of the box.

Hopefully we can incorporate some changes to help support changing the uid and gid if you wanted to.

### DIFF
--- a/Dockerfile.franken
+++ b/Dockerfile.franken
@@ -1,0 +1,163 @@
+# syntax=docker/dockerfile:1
+# Pelican Production Dockerfile using FrankenPHP
+
+# ================================
+# Stage 1-1: Composer Install
+# ================================
+FROM php:8.3-alpine AS composer
+
+WORKDIR /build
+
+# Install composer
+COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
+
+# Install required PHP extensions for Composer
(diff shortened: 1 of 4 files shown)


kungfoome:

parkervcp:
The container runs as www-data which is a standard user for web hosting.

kungfoome:
> The container runs as www-data which is a standard user for web hosting.

The user name doesn't matter so much as the ids. The user can be anything really. The main thing is being able to update the ids to match.

lancepioch:
Thanks for the contribution. At this moment, we're not interested in officially supporting FrankenPHP, let alone another Docker configuration.

But definitely feel free to PR any changes necessary to the existing config that will let people adjust the UID/GID

PR #1311, opened 2025-04-27
Rework subuser permission loading

PalmarHealer:
While not strictly necessary, this makes future expansions easier.

This system allows for the automatic registration of tabs. If extensions are introduced in the future, it will be possible to register tabs via functions.


PalmarHealer:
There even could be a centralised location where the array is defined and you only need to add it at one location, but that goes over my knowledge.
Another idea I had was to make the tab registration self-contained, so that each tab handles its own registration. I'm not sure if this makes sense, but essentially, each site would manage its own tab registration independently.

PalmarHealer:
> I don't know if this is necessary I just took it from the "old" version. But it is not used in app/Filament/Server/Resources/UserResource/Pages/ListUsers.php

I think this I why the test are failing. I haven't touched the $user at all...

rmartinoscar:
Hey great work, could you make a single function in the UserResource and reuse it in UserList instead of redefining it ?
Tests are failing cause `pint` is mad, run `/vendor/bin/pint` before pushing your changes you can even use [git-hooks](https://laraveldaily.com/post/laravel-pint-pre-commit-hooks-github-actions)

Boy132:
Thanks, the frontend for subusers permissions is indeed very messy.
There actually already is a [list with all permissions on the Permission model](https://github.com/pelican-dev/panel/blob/8ae3c88c913e5766816b8de555b3d2a627c59cee/app/Models/Permission.php#L116-L235). Could you change to pr to use this instead?

For the failing tests: Pint can be easily fixed by running `\vendor\bin\pint`, for phpstan you can run `\vendor\bin\phpstan` to check for errors. (also see our [contributing guid](https://github.com/pelican-dev/panel/blob/main/contributing.md)e)

PalmarHealer:
I know it has something to do with the changes I made... Wonders.

I would just keep the old array and move on, but it should work the way I intended. Maybe one of you can take a look and point me in the right direction? I’m not really sure how to approach this.

PalmarHealer:
At least PHPStan is happy now...

PalmarHealer:
I'm bit out of ideas. Could it be that
`protected static array $permissions`
is used somewhere and it is failing since I removed it.

PalmarHealer:
I completely forgot that there is a 'user' entry. I saw 'websocket' because it was on top of the array and 'user' is inside. I think i got all and compared my output with the old `$permissions` array to be sure.

Now it's just Pint how is complaining...

PalmarHealer:
Okay, Pint and I were a bit stubborn, but now it's done. Gotta love those green checkmarks ^-^

I hope this PR is still understandable and didn’t get too messy. Sorry for all the pings.

app/Models/Permission.php
```diff
@@ -115,125 +115,332 @@ class Permission extends Model implements Validatable
...
+                        ],
+                        [
+                            'name' => 'Read Content',
+                            'label' => 'read-content',
+                            'description' => trans('server/users.permissions.file_read_content'),
+                        ],
+                        [
+                            'name' => 'Create',
+                            'label' => 'create',
+                            'description' => trans('server/users.permissions.backup_create'),
```
Boy132, line 192:
```suggestion
                            'description' => trans('server/users.permissions.file_create'),
```

app/Models/Permission.php
```diff
@@ -115,125 +115,332 @@ class Permission extends Model implements Validatable
...
+    public static function permissionTabs(): array
+    {
+        return [
+            [
+                'name' => 'Console',
+                'description' => trans('server/users.permissions.control_desc'),
+                'icon' => 'tabler-terminal-2',
+                'checkboxList' => [
+                    'name' => 'control',
+                    'columns' => 2,
```
Boy132, line 146: `columns` is always 2 so this can be removed.

app/Models/Permission.php
```diff
@@ -115,125 +115,332 @@ class Permission extends Model implements Validatable
...
+                    'options' => [
+                        [
+                            'name' => 'Console',
+                            'label' => 'console',
```
Boy132, line 150: Remove `label`. These two fields are basically the same just with different case. (or see other review below)

app/Models/Permission.php
```diff
@@ -247,6 +454,33 @@ protected function casts(): array
...
+            ],
+            'user' => [
+                'description' => 'Permissions that allow a user to manage other subusers on a server. They will never be able to edit their own account, or assign permissions they do not have themselves.',
+                'keys' => [
+                    'create' => 'Allows a user to create new subusers for the server.',
+                    'read' => 'Allows the user to view subusers and their permissions for the server.',
+                    'update' => 'Allows a user to modify other subusers.',
+                    'delete' => 'Allows a user to delete a subuser from the server.',
+                ],
+            ],
```
Boy132, line 481: Is there a reason these two are extra here and not in the other array?

Boy132, line 481: `User` tab is missing on frontend.

Old:
grafik

Now:
grafik

app/Filament/Server/Resources/UserResource.php
```diff
@@ -83,6 +83,67 @@ public static function table(Table $table): Table
...
+                                    ->bulkToggleable()
+                                    ->label('')
+                                    ->columns($tab['checkboxList']['columns'])
+                                    ->options($options)
+                                    ->descriptions($descriptions),
+                            ]),
+                    ]);
+
+                continue;
+            }
```
Boy132, line 130: Why is there special handling just for `control`?

app/Models/Permission.php
```diff
@@ -115,125 +115,332 @@ class Permission extends Model implements Validatable
...
+                'description' => trans('server/users.permissions.control_desc'),
+                'icon' => 'tabler-terminal-2',
+                'checkboxList' => [
+                    'name' => 'control',
+                    'columns' => 2,
+                    'options' => [
+                        [
+                            'name' => 'Console',
+                            'label' => 'console',
+                            'description' => trans('server/users.permissions.control_console'),
```
Boy132, line 151: The `description` could be dynamically set, right?
The trans key is always `server/users.permissions.` + the group (e.g. `control`) + `_` + permission name (e.g. `console`).

Also if we do that we don't need `name` and `label` but use a simple string array for the options.
Example:
```php
'options' => ['console', 'start', 'restart', 'stop'],
```
Display name would be something like `Str::headline($key)`, and description `trans('server/users.permissions.' . $group . '_' . $key)`.

Boy132 (changes requested):
Looks good, just a few ideas to make everything a bit simpler.

PalmarHealer:
Any update? I made all changes from your suggestions. It should be ready to go.

Boy132 (approved):

notAreYouScared (approved):

PR #1337, opened 2025-05-04
move redis only command to if statement in kernel

notAreYouScared:
Kernel will run this command no matter the setting in the env, per docs, this only applies to cache when it's using redis as the driver, added if statement, no point in having an extra command to run when it will just error out.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -31,8 +31,11 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule): void
     {
-        // https://laravel.com/docs/10.x/upgrade#redis-cache-tags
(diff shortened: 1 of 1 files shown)


rmartinoscar:
image
I don't get any error using file or database cache driver.
https://github.com/laravel/framework/blob/12.x/src/Illuminate/Cache/Console/PruneStaleTagsCommand.php#L41
Has a fail safe, yes the `ConsoleCommand` will error out in CLI if not using redis but not the `ScheduleEvent` returned by `$schedule->command()`

notAreYouScared:
I look at it as a "one less command that needs to run"

Boy132 (approved):
There is no harm in adding the additional check.

PR #1341, opened 2025-05-05
Cleanup files mount and fix path for global search

Boy132:
- Overwriting `mount` to get the path & search term in `ListFiles` & `SearchFiles` isn't needed.
- `path` in `SearchFiles` was never correctly filled because of missing `Url` attribute.

### DIFF
--- a/app/Filament/Server/Resources/FileResource/Pages/ListFiles.php
+++ b/app/Filament/Server/Resources/FileResource/Pages/ListFiles.php
@@ -46,16 +46,9 @@ class ListFiles extends ListRecords
     protected static string $resource = FileResource::class;
 
     #[Locked]
-    public string $path;
+    public string $path = '/';
 
     private DaemonFileRepository $fileRepository;
-
-    public function mount(?string $path = null): void
-    {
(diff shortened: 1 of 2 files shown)


notAreYouScared (approved):

app/Filament/Server/Resources/FileResource/Pages/ListFiles.php
```diff
@@ -46,17 +46,10 @@ class ListFiles extends ListRecords
     protected static string $resource = FileResource::class;
 
     #[Locked]
-    public string $path;
+    public string $path = '/';
```
rmartinoscar, line 49: Why is this one a Locked when the other is a Url ?

Boy132: On ListFiles the path in included in the url (`/files/{path}`).
On SearchFiles only the search term is included in the url and the path is an additional parameter. (`/files/search/{searchTerm}?path={path}`)

rmartinoscar (approved):

PR #1348, opened 2025-05-08
Hide `ChartWidgets` when `Server` isInConflictState or Offline

### LINKED ISSUE #1346: Console graphs should be hidden if server is not online
CPU/Memory/Network graphs should be hidden when the server is not starting/installing/running.

rmartinoscar:
Closes #1346

### DIFF
--- a/app/Filament/Server/Pages/Console.php
+++ b/app/Filament/Server/Pages/Console.php
@@ -125,19 +125,7 @@ class Console extends Page
      */
     public function getVisibleWidgets(): array
     {
-        /** @var Server $server */
-        $server = Filament::getTenant();
-
-        $widgets = collect($this->filterVisibleWidgets($this->getWidgets()));
-        if ($server->isInConflictState() || $server->retrieveStatus()->isOffline()) {
-            $widgets = $widgets->filter(fn ($widget) => !in_array($widget, [
-                ServerCpuChart::class,
-                ServerMemoryChart::class,
-                ServerNetworkChart::class,
-            ]));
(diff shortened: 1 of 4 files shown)


Boy132 (changes requested):
Overwrite `canView` inside the Widgets instead.

```php
public static function canView(): bool
{
	/** @var Server $server */
	
	return !$server->isInConflictState() && !$server->retrieveStatus()->isOffline();
}
```

rmartinoscar:
> Overwrite `canView` inside the Widgets instead.
> 
> ```
> public static function canView(): bool
> {
> 	/** @var Server $server */
> 	$server = Filament::getTenant();
> 	
> 	return !$server->isInConflictState() && !$server->retrieveStatus()->isOffline();
> }
> ```

It felt like more trouble then worth since we already redefine `getVisibleWidgets()` but i'll change that if you insist.

Boy132:
> > Overwrite `canView` inside the Widgets instead.
> > ```
> > public static function canView(): bool
> > {
> > 	/** @var Server $server */
> > 	$server = Filament::getTenant();
> > 	
> > 	return !$server->isInConflictState() && !$server->retrieveStatus()->isOffline();
> > }
> > ```
> 
> It felt like more trouble then worth since we already redefine `getVisibleWidgets()` but i'll change that if you insist.

I think it's cleaner this way and also less hard coding. (Also `getVisibleWidgets()` is only there because the Console is a custom page.)

Boy132 (approved):

PR #1349, opened 2025-05-08
Use `except` so `Tags` work in `StoreNodeRequest`

rmartinoscar:
When you make a request with something that isn't in `Model::getRules()` its omitted

### DIFF
--- a/app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php
+++ b/app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php
@@ -18,26 +18,7 @@ class StoreNodeRequest extends ApplicationApiRequest
      */
     public function rules(?array $rules = null): array
     {
-        return collect($rules ?? Node::getRules())->only([
-            'public',
-            'name',
-            'description',
-            'fqdn',
-            'scheme',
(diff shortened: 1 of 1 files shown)


app/Http/Requests/Api/Application/Nodes/StoreNodeRequest.php
```diff
@@ -18,26 +18,7 @@ class StoreNodeRequest extends ApplicationApiRequest
...
-            'disk_overallocate',
-            'cpu',
-            'cpu_overallocate',
-            'upload_size',
-            'daemon_listen',
-            'daemon_sftp',
-            'daemon_sftp_alias',
-            'daemon_base',
-        ])->mapWithKeys(function ($value, $key) {
+        return collect($rules ?? Node::getRules())->except('maintenance_mode')->mapWithKeys(function ($value, $key) {
```
Boy132, line 21: Why except `maintenance_mode`?

Boy132: There is no harm in allowing to set it. It's just odd that only `maintenance_mode` is excepted. :D

Boy132 (approved):

PR #1350, opened 2025-05-08
Enable ipv6 on frontend

Boy132:
Ipv6 allocations already worked fine with the backend.
This makes the frontend accept ipv6 and display them correctly.


lancepioch (approved):

notAreYouScared (approved):
Personally I dont hae ipv6 so I say we merge it and let the people test it.

PR #1354, opened 2025-05-08
DiscordWebhooks

JoanFo1456:
Something I worked with Martin, it works with actual canary, i made sure of it..
I had to update faker, if you don't update it, the preview won't work.
I think this is good enough to be first version, but i'm open to suggestions/changes.


JoanFo1456:
Let me do some changes, seems that my git is not working correctly.. Some files are missing.

JoanFo1456:
Will open another pr once i have all repaired, seems that git undid most of the changes on this branch, i have to redo some of them.

PR #1378, opened 2025-05-14
Use `supervisorctl` instead of `systemctl` when running in docker

rmartinoscar:


Boy132:
`p:environment:queue-service` shouldn't be used when running the panel inside docker?

rmartinoscar:
> `p:environment:queue-service` shouldn't be used when running the panel inside docker? It's only used for creating the queue worker service on a normal install.

Its used whenever you switch to redis

Boy132:
> > `p:environment:queue-service` shouldn't be used when running the panel inside docker? It's only used for creating the queue worker service on a normal install.
> 
> Its used whenever you switch to redis

Yeah, for the non dockerized install. Your changes will only restart the service which is pointless and has nothing to do with the command.

rmartinoscar:
Then how are you supposed to change drivers once its installed beside manually editing `.env` or adding those to the `compose` file ?
It makes sense for me to allow all standalone commands to be ran in Docker as well

PR #1396, opened 2025-05-22
Trait-ify resources and add customizable options

Boy132:
Adds new traits to most of our resources to make it possible to customize pages, relations, tables and forms. This can later be used by plugins to extend (and change) our default resources.

I also added a trait that handles the access when the server is in a conflict state and a trait for displaying a count/ limit on the navigation badge.


notAreYouScared (approved):
LGTM, nothing appears to be broken

PR #1413, opened 2025-05-28
Lazy load server entries (`Grid` only)

Boy132:
Animation


notAreYouScared (approved):

Boy132:
Marking this as draft for now because sometimes entries are not loaded correctly (either blank or missing "indicator bar")
grafik

My guess is this is because of the tabs and/ or SPA.

PR #1430, opened 2025-06-03
Refactor `EnvironmentTrait` to use Env Facade

rmartinoscar:
Also remove `columnSpanFull`
image


app/Traits/EnvironmentWriterTrait.php
```diff
@@ -2,54 +2,20 @@
...
-
-            if (preg_match_all('/^' . $key . '=(.*)$/m', $saveContents) < 1) {
-                $saveContents = $saveContents . PHP_EOL . $saveValue;
-            } else {
-                $saveContents = preg_replace('/^' . $key . '=(.*)$/m', $saveValue, $saveContents);
-            }
-        });
-
-        file_put_contents($path, $saveContents);
+        Env::writeVariables($values, base_path('.env'), true);
```
Boy132, line 19: At this point we can also just get rid of the trait and use `Env::writeVariables` directly where needed, no?

rmartinoscar: I rather keep it so we don't have to specify the path & override everytime but we could, that's why i asked your review too

Boy132 (approved):

PR #1435, opened 2025-06-06
Fix hostname env variable name in rust egg

Boy132:
`HOSTNAME` is reserved, so rename it to `SERVER_HOSTNAME`.
This is done automatically when importing eggs but when updating from older versions this can cause sql errors.


database/Seeders/eggs/rust/egg-rust.json
```diff
@@ -20,7 +20,7 @@
         "ghcr.io\/parkervcp\/games:rust": "ghcr.io\/parkervcp\/games:rust"
```
QuintenQVD0, line 20: This is not exported from a panel as the timestamp did not update

Boy132: No idea what you are talking about :x

PR #1448, opened 2025-06-12
Delete old csgo egg

Boy132:


rmartinoscar:
Do we not replace it with cs2 instead of removing it entirely ?

Boy132:
> Do we not replace it with cs2 instead of removing it entirely ?

IMO no. Eggs should be downloaded from the eggs repo if needed.

rmartinoscar:
> > Do we not replace it with cs2 instead of removing it entirely ?
> 
> IMO no. Eggs should be downloaded from the eggs repo if needed.

I'd still keep it until we have @notAreYouScared's egg downloader builtin to the panel

lancepioch (approved):

rmartinoscar (approved):

PR #1451, opened 2025-06-17
Add "egg index" and dropdown to egg importer

Boy132:
Adds a new command that checks [pelican.json](https://github.com/pelican-eggs/pelican-eggs.github.io/blob/main/content/pelican.json) to get a list of all available eggs.
This eggs can be added to the egg url importer via dropdown.

grafik

### DIFF
--- a/app/Filament/Components/Actions/ImportEggAction.php
+++ b/app/Filament/Components/Actions/ImportEggAction.php
@@ -2,6 +2,7 @@
 
 namespace App\Filament\Components\Actions;
 
+use App\Console\Commands\Egg\UpdateEggIndexCommand;
 use App\Models\Egg;
 use App\Services\Eggs\Sharing\EggImporterService;
 use Closure;
@@ -17,6 +18,7 @@ use Filament\Forms\Get;
 use Filament\Forms\Set;
 use Filament\Notifications\Notification;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Str;
 use Livewire\Features\SupportFileUploads\TemporaryUploadedFile;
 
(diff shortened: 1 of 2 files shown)


notAreYouScared (approved):

app/Console/Kernel.php
```diff
@@ -41,7 +42,9 @@ protected function schedule(Schedule $schedule): void
 
         $schedule->command(CleanServiceBackupFilesCommand::class)->daily();
         $schedule->command(PruneImagesCommand::class)->daily();
-        $schedule->command(CheckEggUpdatesCommand::class)->hourly();
+
+        $schedule->command(CheckEggUpdatesCommand::class)->daily();
+        $schedule->command(UpdateEggIndexCommand::class)->daily();
```
rmartinoscar, line 47: `daily()` only runs at 00:00 so unless you manually run `php artisan p:egg:update-index` you won't see any of them ~~maybe we should add it to one of these ?~~
https://github.com/pelican-dev/panel/blob/0156ac15092f67c49542f5fb895988fa74eb3fa1/composer.json#L83-L88

Other then that seems good to me.

rmartinoscar (approved):

app/Console/Kernel.php
notAreYouScared: I suggested, in our internal chat, adding a manual hint action that would pull it in demand.

rmartinoscar: That works better sorry i didn't see

app/Console/Commands/Egg/UpdateEggIndexCommand.php
```diff
@@ -0,0 +1,46 @@
...
+    public function handle(): int
+    {
+        try {
+            $data = file_get_contents('https://raw.githubusercontent.com/pelican-eggs/pelican-eggs.github.io/refs/heads/main/content/pelican.json');
+            $data = json_decode($data, true, 512, JSON_THROW_ON_ERROR);
+        } catch (Exception $exception) {
+            $this->error($exception->getMessage());
+
+            return 1;
+        }
```
lancepioch, line 21: I think you need this?

```suggestion
        } catch (Exception $exception) {
            $this->error($exception->getMessage());
            
            report($exception);

            return 1;
        }
```

PR #1461, opened 2025-06-21
Fix: Search for tags in correct variable

wbvanderbroek:
Previously creating a server through the api on a specific node wouldn't work with something like this in the payload:

"deploy" => [
            "dedicated_ip" => false,
            "tags" => ["yournodetag"],
            "port_range" => []
        ],
Because the tag would be completely erased by the validator. And would only be stored in the deployment variable.
This will fix that by using the deployment variable instead of the data variable

### DIFF
--- a/app/Services/Servers/ServerCreationService.php
+++ b/app/Services/Servers/ServerCreationService.php
@@ -128,7 +128,7 @@ class ServerCreationService
             Arr::get($data, 'memory', 0),
             Arr::get($data, 'disk', 0),
             Arr::get($data, 'cpu', 0),
-            Arr::get($data, 'tags', []),
+            $deployment->getTags(),
         );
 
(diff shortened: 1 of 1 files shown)


wbvanderbroek:
> I have read the CLA Document and I hereby sign the CLA

wbvanderbroek:
recheck

wbvanderbroek:

wbvanderbroek:
recheck

wbvanderbroek:
I just noticed some more strange behaviour with creating servers through the api. When NOT using the deployment it will use the first allocation available, but when you do use the deployment it simply gets a random allocation that is available instead.

I guess this is fine though since the allocation specified will then place it in the correct node but it does feel a bit off.

lancepioch (approved):

lancepioch:
Thank you!

PR #1470, opened 2025-06-26
back port power actions from v4 branch

notAreYouScared:
New power actions that are under the "heartbeat"

https://github.com/user-attachments/assets/7a623c79-47ca-4574-ad40-de7fae368d51


rmartinoscar (approved):
Since we don't use ContextMenu anymore we should get rid of the dispatch in powerActions, thought we could do it in this PR other then that looks good.

PR #1478, opened 2025-07-03
Add SSH Keys to Profile

### LINKED ISSUE #1201: Re-Add SSH Keys for SFTP Auth
SSH-Keys still need to be re-added to the new profile page, along with the supporting backend.

Boy132:
Closes #1201
Also fixes activity log for deleting api keys.

grafik
_Yes, the fingerprint is overflowing and there is this weird gap. That's because of the (non-exsiting) label and I don't know to fix it. :(_


rmartinoscar:
I feel like SSHKeyController, StoreSSHKeyRequest & tests should be refactored to use the new service, other then type casting looks good i'll give it a try when home.

notAreYouScared (approved):

PR #1479, opened 2025-07-03
Replace `gethostbynamel` with `dns_get_record`

### LINKED ISSUE #1477: Panel lists node domain as invaid if only AAAA record is avalible
### Current Behavior

Currently the Panel lists only A-Records as valid in the Panel. But if a host is reachable only via IPv6 and therefore the domain only has an AAAA-Record, the panel ignores it.

### Expected Behavior

I would expect that the Panel will also check if AAAA-Records are available. 

As an example googles IPv4 and IPv6 only domains:

Image

Image

As you can see only A-Records are accepted.

### Steps to Reproduce

Create a DNS Entry that only points to a AAAA-Record. Try to add it as node in Pelican. It will show as invalid.

I think I found the code that is responsible for this:
The Panel uses the `gethostbynamel` function. This PHP function is known for only supporting IPv4, which is criticized for years. The exact same issue was part of Pterodactyl a few years ago: https://github.com/pterodactyl/panel/pull/4398

### Panel Version

1.0.0-beta21

### Wings Version

1.0.0-beta13

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

Boy132:
Closes #1477

grafik
grafik

### DIFF
--- a/app/Filament/Admin/Resources/NodeResource/Pages/CreateNode.php
+++ b/app/Filament/Admin/Resources/NodeResource/Pages/CreateNode.php
@@ -89,11 +89,20 @@ class CreateNode extends CreateRecord
                                         return;
                                     }
 
-                                    $validRecords = gethostbynamel($state);
-                                    if ($validRecords) {
+                                    $validARecords = @dns_get_record($state, DNS_A);
+                                    if ($validARecords) {
                                         $set('dns', true);
 
-                                        $set('ip', collect($validRecords)->first());
+                                        $set('ip', collect($validARecords)->first()['ip']);
+
+                                        return;
+                                    }
+
+                                    $validAAAARecords = @dns_get_record($state, DNS_AAAA);
+                                    if ($validAAAARecords) {
+                                        $set('dns', true);
+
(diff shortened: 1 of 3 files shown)


rmartinoscar:
`dns_get_record()` doesn't work with `/etc/hosts` overrides.

marpisco:
> `dns_get_record()` doesn't work with `/etc/hosts` overrides.

Shouldn't be supposed to. It's meant to check if a domain is _actually valid_ for proper reach from a browser.

rmartinoscar:
> Shouldn't be supposed to. It's meant to check if a domain is _actually valid_ for proper reach from a browser.

You are right but its quite handy when debugging, this way you know what the panel resolves to and if you wanna know what the browser do you can just use a simple ping, while now you'll have to enter the container.

marpisco:
> its quite handy when debugging, this way you know what the panel resolves to

Debugging tools like that wouldn't be a terrible idea to add to the panel tho. Both useful for server admins and you, developers of this panel.

lancepioch (approved):

PR #1489, opened 2025-07-06
Add migration to match node ports

Boy132:
Make sure the node ports matche if not `behind_proxy`.

### DIFF
--- a/database/migrations/2025_07_06_213447_match-node-port.php
+++ b/database/migrations/2025_07_06_213447_match-node-port.php
@@ -1,5 +1,6 @@
 <?php
 
+use App\Models\Node;
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Support\Facades\DB;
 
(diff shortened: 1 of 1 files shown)


database/migrations/2025_07_06_213447_match-node-port.php
```diff
@@ -0,0 +1,25 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::table('nodes')
+            ->where('behind_proxy', false)
+            ->update(['daemon_connect' => DB::raw('`daemon_listen`')]);
```
rmartinoscar:
```suggestion
        $nodes = Node::where('behind_proxy', false);
        foreach ($nodes as $node) { 
            /** @var Node $node */
            $node->update(['daemon_connect' => $node->daemon_listen]));
        }
```

rmartinoscar (dismissed):
Fix postgres

notAreYouScared (approved):

PR #1496, opened 2025-07-07
Filter out already used ips with the same port

### LINKED ISSUE #1495: 500 when changing allocation's ip using Select
Image
Image

https://github.com/pelican-dev/panel/blob/514d961c24616c1e8960574eb9848c5e3e8becac/app/Filament/Admin/Resources/NodeResource/RelationManagers/AllocationsRelationManager.php#L65
We should filter out available ips for the given port.

rmartinoscar:
Closes #1495


mkaulfers:
Related to this? https://github.com/pelican-dev/panel/pull/1508

PR #1498, opened 2025-07-08
Use filaments password broker for reset link token when creating subuser

Boy132:
~~Otherwise you sometimes get `403 invalid signature` errors.~~ Just a small refactor.


notAreYouScared (approved):
I don't use the emails, but if this fixes it 👍

Boy132:
So, turns out the 403 errors are a problem with url encoding. While this PR does _not_ fix that it's still a nice little refactor.

PR #1501, opened 2025-07-08
Improve turnstile error handling (+ cleanup)

Boy132:


app/Extensions/Captcha/Schemas/Turnstile/Rule.php
```diff
@@ -4,17 +4,21 @@
...
 use Illuminate\Support\Facades\App;
 
 class Rule implements ValidationRule
 {
     public function validate(string $attribute, mixed $value, Closure $fail): void
     {
-        $response = App::call(fn (CaptchaService $service) => $service->getActiveSchema()->validateResponse($value));
+        try {
+            $service = App::make(CaptchaService::class);
+            $service->get('turnstile')->validateResponse($value);
```
rmartinoscar:
```suggestion
            App::call(fn (CaptchaService $service) => $service->get('turnstile')->validateResponse($value));
```
There's no use for a variable and we should use `App::call` not `App::make` since this is a singleton.

PR #1508, opened 2025-07-10
Adds same-network check for port allocations.

mkaulfers:
Adds a new validation check for same network check on nodes and port allocations. If two separate nodes share the same subnet, it will throw an invalid port exception whenever a duplicate is detected.

### DIFF
--- a/app/Exceptions/Service/Allocation/PortConflictOnSameNetworkException.php
+++ b/app/Exceptions/Service/Allocation/PortConflictOnSameNetworkException.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace App\Exceptions\Service\Allocation;
+
+use App\Exceptions\DisplayException;
+
+class PortConflictOnSameNetworkException extends DisplayException
+{
+    /**
+     * PortConflictOnSameNetworkException constructor.
+     */
+    public function __construct(int $port, string $ip, int $conflictingNodeId)
+    {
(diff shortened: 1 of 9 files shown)


mkaulfers:

JoanFo1456:
You need to fix the issues the tests are pointing out, you can help yourself with phpstan on your own computer or following the tests results, from there it's easier to check the branch...

mkaulfers:
@JoanFo1456 before I overcommit, do you think it's a worthwhile change adding in cross-node port validation on the same subnet?

JoanFo1456:
I'm not sure, as I'm not a pelican, nor part from this project, I suggest you fix it and them will say if it's useful or not.

Boy132:
Could you please carify the benefits of these changes?
In what cases would two nodes share the same subnet? And how could there be a port conflict?

parkervcp:
This could potentially break any hosts that may use Tunneling/VPNs to proxy their connections. 

As an example. I have 2 game hosts on the same wireguard instance with `10.210.66.x` and upstream proxies with different external ip addresses that forward to the hosts on the other end. This would break that.

rmartinoscar:
This would also break multiple nodes on the same public subnet.
I'v had multiple setups in a 123.234.123.XX/28 where 123.234.123.2 is the panel & 123.234.123.3 & 123.234.123.4 are nodes.

PR #1514, opened 2025-07-11
Improve schedule form

Boy132:
Small cleanups and fixes for the schedule form.

Before:
old create page
new view page
new edit page

After:
new create page
new view page
new edit page

### DIFF
--- a/app/Filament/Components/Forms/Actions/CronPresetAction.php
+++ b/app/Filament/Components/Forms/Actions/CronPresetAction.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace App\Filament\Components\Forms\Actions;
+
+use Filament\Forms\Components\Actions\Action;
+use Filament\Forms\Get;
+use Filament\Forms\Set;
+
+class CronPresetAction extends Action
+{
+    protected string $minute = '0';
+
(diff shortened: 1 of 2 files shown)


notAreYouScared (approved):

app/Filament/Components/Forms/Actions/CronPresetAction.php
```diff
@@ -0,0 +1,53 @@
...
+        parent::setUp();
+
+        $this->disabled(fn (string $operation) => $operation === 'view');
+
+        $this->color(fn (Get $get) => $get('cron_minute') == $this->minute &&
+                                    $get('cron_hour') == $this->hour &&
+                                    $get('cron_day_of_month') == $this->dayOfMonth &&
+                                    $get('cron_month') == $this->month &&
+                                    $get('cron_day_of_week') == $this->dayOfWeek
+            ? 'success' : 'primary');
```
rmartinoscar, line 32: Is there no better way then having 5 comparison every time it checks for color ?

Boy132: I don't think so. If you find a better way let me know.

PR #1515, opened 2025-07-11
WebHook headers fix

JoanFo1456:
This branch fixes that when you don't have custom headers is not sending the predefined one.


JoanFo1456:
Once I get home I’ll fix the lint, made it quickly.

PR #1539, opened 2025-07-20
api(client): use server condition value for status

AidanConn:
api(client): use condition property for server status

Fixes client api, api:client:server.view, by replacing the server status field with the value from the server’s condition property, which was previously always null. Now returns “offline”, “running”, etc.


AidanConn:

notAreYouScared:
Thats for installing/install failed/suspended status

You should be calling the /resources endpoint for status

AidanConn:

Boy132:
Like Charles said this is correct, the status refers to the _panel status_ (e.g. `installing` or `transferring`) and not the _container status_ (e.g. `running`).

PR #1540, opened 2025-07-20
Switch the sub-server File Manager to display more files per page

memesalot:
### Added
- Larger file-manager pagination options (100, 150, 200).

For server that have a large amount of plugins the default values of 25/50 is very annoying to use. No reason to not support larger pagination as the allocation page for the node options supports this.


Lovinoes:
They're limited to 50 due to performance issues with filament afaik

image

image

notAreYouScared:
This can be revisited once we move to filament v4. A lot of performance improvements have been made compared to v3

Issue #1545, opened 2025-07-21
Allow for backups to be renamed


Awhikax:
As a panel user,
I want to rename backups from the panel,
So that I can keep them organized

AndyIsHereBoi:
This is something I really wish pterodactyl let you do because sometimes I forget to set one before creating it

PR #1546, opened 2025-07-21
Allow for backups to be renamed

Awhikax:
Addresses issue #1545

### DIFF
--- a/app/Filament/Server/Resources/BackupResource.php
+++ b/app/Filament/Server/Resources/BackupResource.php
@@ -116,6 +116,28 @@ class BackupResource extends Resource
             ])
             ->actions([
                 ActionGroup::make([
+                    Action::make('rename')
+                        ->icon('tabler-pencil')
+                        ->authorize(fn () => auth()->user()->can(Permission::ACTION_BACKUP_DELETE, $server))
+                        ->label('Rename')
+                        ->form([
+                            TextInput::make('name')
+                                ->label('Backup Name')
+                                ->required()
+                                ->maxLength(255)
+                                ->default(fn (Backup $backup) => $backup->name),
+                        ])
+                        ->action(function (Backup $backup, $data, BackupController $backupController, Request $request) use ($server) {
+                            $request->merge(['name' => $data['name']]);
+                            $backupController->rename($request, $server, $backup);
+
+                            Notification::make()
+                                ->title('Backup Renamed')
+                                ->body('The backup has been successfully renamed.')
+                                ->success()
+                                ->send();
+                        })
+                        ->visible(fn (Backup $backup) => $backup->status === BackupStatus::Successful),
                     Action::make('lock')
                         ->icon(fn (Backup $backup) => !$backup->is_locked ? 'tabler-lock' : 'tabler-lock-open')
                         ->authorize(fn () => auth()->user()->can(Permission::ACTION_BACKUP_DELETE, $server))
--- a/app/Filament/Server/Resources/BackupResource.php
+++ b/app/Filament/Server/Resources/BackupResource.php
@@ -127,9 +127,18 @@ class BackupResource extends Resource
                                 ->maxLength(255)
                                 ->default(fn (Backup $backup) => $backup->name),
                         ])
-                        ->action(function (Backup $backup, $data, BackupController $backupController, Request $request) use ($server) {
-                            $request->merge(['name' => $data['name']]);
-                            $backupController->rename($request, $server, $backup);
+                        ->action(function (Backup $backup, $data) {
+                            $oldName = $backup->name;
+                            $newName = $data['name'];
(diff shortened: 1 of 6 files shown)


JoanFo1456:
I think you should check if it has a permission, for example delete. (Made a comment since I can't make an edit to suggest a change)

Awhikax:
> I think you should check if it has a permission, for example delete. (Made a comment since I can't make an edit to suggest a change)

Isn't it already the case?

JoanFo1456:
> > I think you should check if it has a permission, for example delete. (Made a comment since I can't make an edit to suggest a change)
> 
> Isn't it already the case?

Seems so, Copilot was tripping, mb

Awhikax:

Awhikax:
recheck

Awhikax:
Side note: I wasn't sure if it should allow renaming when the backup is locked. I've made it so it's possible as this is not really a destructive action but it may be changed if needed.

JoanFo1456:
I think you shouldn't be able to do so. Like when locked is 100% locked... Not delete, so why rename. Or at least is what I think.

Awhikax:
Both works to me but I do have one issue with unlocking a backup to rename it: if your backup slots are full and that an automated backup (schedule) triggers exactly when you unlock your backup to rename it, it'll be lost. This is a rare case, yet this can happen. I personally see the lock more as a protection against deletion than a way to prevent the user to do anything related to the backup in this scenario.

lang/fr/activity.php
```diff
@@ -58,6 +58,7 @@
             'fail' => 'La sauvegarde <b>:name</b> a échoué',
             'lock' => 'La sauvegarde <b>:name</b> a été verrouillée',
             'unlock' => 'La sauvegarde <b>:name</b> a été déverrouillée',
+            'rename' => 'La sauvegarde a été renommée de "<b>:old_name</b>" à "<b>:new_name</b>"',
```
Boy132: Translations are handled via crowdin. Only update the english translations.

app/Filament/Server/Resources/BackupResource.php
```diff
@@ -116,6 +116,28 @@ public static function defaultTable(Table $table): Table
...
+                        ->form([
+                            TextInput::make('name')
+                                ->label('Backup Name')
+                                ->required()
+                                ->maxLength(255)
+                                ->default(fn (Backup $backup) => $backup->name),
+                        ])
+                        ->action(function (Backup $backup, $data, BackupController $backupController, Request $request) use ($server) {
+                            $request->merge(['name' => $data['name']]);
+                            $backupController->rename($request, $server, $backup);
```
Boy132: Do not use an api controller here. Either make a service or handle the rename here directly.

app/Http/Controllers/Api/Client/Servers/BackupController.php
```diff
@@ -195,6 +195,39 @@ public function download(Request $request, Server $server, Backup $backup): Json
...
+            'name' => 'required|string|max:255',
+        ]);
+
+        $oldName = $backup->name;
+        $backup->update(['name' => $request->input('name')]);
+
+        Activity::event('server:backup.rename')
+            ->subject($backup)
+            ->property(['old_name' => $oldName, 'new_name' => $backup->name])
+            ->log();
```
Boy132: Activity should only be logged if the name actually changed.

app/Http/Controllers/Api/Client/Servers/BackupController.php
```diff
@@ -195,6 +195,39 @@ public function download(Request $request, Server $server, Backup $backup): Json
...
+     */
+    public function rename(Request $request, Server $server, Backup $backup): array
+    {
+        if (!$request->user()->can(Permission::ACTION_BACKUP_DELETE, $server)) {
+            throw new AuthorizationException();
+        }
+
+        $request->validate([
+            'name' => 'required|string|max:255',
+        ]);
```
Boy132: The request should handle permissions and validation, see `StoreBackupRequest` for an example.

Boy132 (changes requested):
Thank you for your contribution. Just some small changes needed.
Also, renaming locked backups seems fine to me.

app/Filament/Server/Resources/BackupResource.php
rmartinoscar: Service seems like a better fit with existing code.

Boy132: Generally yes, but a whole service for basically just two or three lines doesn't seem worth it.

Awhikax: I can make it a service if needed but this looked a bit excessive for this specific case compared to the existing code.

lancepioch (approved):

PR #1552, opened 2025-07-22
use condition instead of status

hasyirin:
Server status is null most of the time, I think it should use server condition instead.
Honestly, I think it's better if we deprecate status field and add in a condition field instead, but other deprecated comments in the files are saying to use status field.. so...


notAreYouScared:
https://github.com/pelican-dev/panel/pull/1539

hasyirin:
Also using the value of `$server->condition` because Filament Resource EditServer.php is also using it to show "server status" on line 134.

PR #1557, opened 2025-07-25
Cleanup and update node packages

Boy132:
Remove unused packages, update packages, cleanup `app.css`/ `app.js` loading


package.json
```diff
@@ -6,16 +6,13 @@
         "dev": "vite"
     },
     "devDependencies": {
-        "@tailwindcss/forms": "^0.5.9",
-        "@tailwindcss/typography": "^0.5.15",
-        "autoprefixer": "^10.4.20",
-        "axios": "^1.7.4",
-        "concurrently": "^9.0.1",
```
rmartinoscar, line 13: This is required https://github.com/pelican-dev/panel/blob/88943563c7c15f13e689ca137c6d9dcda175baee/composer.json#L91

vite.config.js
```diff
@@ -7,8 +7,9 @@ export default defineConfig({
             input: [
                 'resources/css/app.css',
                 'resources/js/app.js',
-                'resources/js/console.js',
```
rmartinoscar, line 10: How about we use wildcards like for tailwind since load order doesn’t matter ?

rmartinoscar (changes requested):

package.json
Boy132: I don't think anyone uses `composer dev` either

rmartinoscar: True just letting you know where it was used, either way I don't feel like running concurrent tasks from npx in composer is best practice so we should replace or remove it.

rmartinoscar:
I think you need `**` for wildcards since they are nested folders.

Boy132:
> I think you need `**` for wildcards since they are nested folders.

They aren't nested folders.

notAreYouScared (approved):

notAreYouScared (approved):

PR #1559, opened 2025-07-26
remove maxlength rule from egg variable's default_value field

hasyirin:
The default_value field used to be string type in database, however it was changed to text.

[Rust autowipe egg](https://github.com/pelican-eggs/games-steamcmd/blob/main/rust/rust_autowipe/egg-rust-autowipe.json) has a variable which has pretty lengthy default value, it will always throw an error whenever we try to edit the egg due to default_value being limited to 255. 

Also since it's a text type in database, should the input for this field become textarea instead of normal text input?


Boy132 (approved):
Good catch, thank you for your contribution!

I think the `TextInput` is fine, `Textarea` only make sense when there are multiline values.

hasyirin:
Ah right, that makes sense. Thanks :)

PR #1566, opened 2025-07-29
Fixed email from name allowing ' breaking .env

### LINKED ISSUE #1561: Apostrophe in email "From Name" causes 500 Error
### Current Behavior

Adding an apostrophe into the "From Name" of an email in the SMTP settings causes the site to give out a 500 error. Inspecting the .env we can see that the apostrophe has been escaped using a backslash such as: \'
This causes the error to occur. When running the updating script it mentions that the line is unexpected; removing the backslash causes panel to run as expected. Further, fixing this manually in the .env means that any further changes in the settings tab require going back to the .env and fixing this again. There is no error apparent in the logs.

### Expected Behavior

Apostrophes should be either not escaped using a backslash, or the escaped character should be handled properly and not cause a crash.

### Steps to Reproduce

Adding an apostrophe into the "From Name" title in the SMTP settings of panel then saving. 

### Panel Version

1.0.0-beta23

### Wings Version

1.0.0-beta15

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

JoanFo1456:
It's a simple regex, I don't think any feedback is required, but welcome it is...
Closes #1561

### DIFF
--- a/app/Filament/Admin/Pages/Settings.php
+++ b/app/Filament/Admin/Pages/Settings.php
@@ -400,7 +400,8 @@ class Settings extends Page implements HasForms
                     TextInput::make('MAIL_FROM_NAME')
                         ->label(trans('admin/setting.mail.from_name'))
                         ->required()
-                        ->default(env('MAIL_FROM_NAME', config('mail.from.name'))),
+                        ->default(env('MAIL_FROM_NAME', config('mail.from.name')))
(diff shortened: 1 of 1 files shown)


Boy132:
Thanks, but I would rather fix the escaping in our env writer than restrict the input.

PR #1567, opened 2025-07-30
Fixed env parser allowing ' and breaking.

JoanFo1456:
I attempted with both " and ` seems that the only one that broke .env was '.
Let me know if you guys prefer to do it so I can stop trying...


Boy132:
This won't fix anything. Using `'` should not be prohibited.
As this is a laravel problem there is now a laravel github issue. A fix needs to be applied in laravel.

PR #1569, opened 2025-08-01
Fix server creation without deployment

Boy132:
When not using deployment, get the `node_id` from the allocation.

### DIFF
--- a/app/Services/Servers/ServerCreationService.php
+++ b/app/Services/Servers/ServerCreationService.php
@@ -39,6 +39,8 @@ class ServerCreationService
      * as possible given the input data. For example, if an allocation_id is passed with
      * no node_id the node_is will be picked from the allocation.
      *
+     * @param  array<mixed, mixed>  $data
+     *
      * @throws \Throwable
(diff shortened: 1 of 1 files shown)


app/Services/Servers/ServerCreationService.php
```diff
@@ -39,15 +39,7 @@ public function __construct(
...
-     * @param array{
-     *     node_id?: int,
-     *     oom_killer?: bool,
-     *     oom_disabled?: bool,
-     *     egg_id?: int,
-     *     image?: ?string,
-     *     startup?: ?string,
-     *     start_on_completion?: ?bool,
-     * } $data
+     * @param  array<mixed, mixed>  $data
```
rmartinoscar, line 42: Why did you replace proper heredoc with mixed ?
Other then that looks good
(Im unable to checkout atm)

Boy132: Because they were wrong.

rmartinoscar: Guess we can get to that whenever we bump phpstan level

PR #1570, opened 2025-08-02
Webhooks not working variables such actor.username

JoanFo1456:
This is something a friend found, he used ChatGPT and made a suggestion that works. He asked me to open the pull request for it, as he doesn't want. 
Probably the code needs to be changed, but I'm not sure how. I'll do the suggested changes myself from here.


JoanFo1456:
I'll close this until i find a way to fix both...

PR #1571, opened 2025-08-03
Update healthcheck

parkervcp:
Updates healthcheck to handle caddy and php fcgi for when caddy is disabled
Updates php upload limit env var

Fixes how trusted proxies are handled in caddy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,10 +93,11 @@ COPY docker/Caddyfile /etc/caddy/Caddyfile
 # Add Laravel scheduler to crontab
 COPY docker/crontab /etc/supercronic/crontab
 
-COPY docker/entrypoint.sh ./docker/entrypoint.sh
+COPY docker/entrypoint.sh /entrypoint.sh
+COPY docker/healthcheck.sh /healthcheck.sh
 
 HEALTHCHECK --interval=5m --timeout=10s --start-period=5s --retries=3 \
-  CMD curl -f http://localhost/up || exit 1
+  CMD bash /healthcheck.sh
 
 EXPOSE 80 443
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -93,16 +93,17 @@ COPY docker/Caddyfile /etc/caddy/Caddyfile
 # Add Laravel scheduler to crontab
 COPY docker/crontab /etc/supercronic/crontab
 
-COPY docker/entrypoint.sh ./docker/entrypoint.sh
+COPY docker/entrypoint.sh /entrypoint.sh
+COPY docker/healthcheck.sh /healthcheck.sh
 
 HEALTHCHECK --interval=5m --timeout=10s --start-period=5s --retries=3 \
-  CMD curl -f http://localhost/up || exit 1
+  CMD bash /healthcheck.sh
```
QuintenQVD0, line 101: Bash on alpine?

Should we not use ash?

QuintenQVD0 (changes requested):

docker/entrypoint.sh
```diff
@@ -41,6 +42,11 @@ php artisan filament:optimize
 
 export SUPERVISORD_CADDY=false
 
+if [[ ! -z ${TRUSTED_PROXIES} ]]; then
```
rmartinoscar: Should we move that to line 54 since this is only for CADDY ?

Dockerfile
parkervcp: missed that. The shell file is even set for ash.

docker/entrypoint.sh
parkervcp: moved

PR #1573, opened 2025-08-04
Auto create missing users when using oauth

Boy132:
Adds a new option to all providers to automatically create (& link) missing users. This is turned off by default.
These users will still recieve a "Account created" mail with a password reset link, if they wish to also setup a panel password.

grafik

### DIFF
--- a/app/Extensions/OAuth/OAuthSchemaInterface.php
+++ b/app/Extensions/OAuth/OAuthSchemaInterface.php
@@ -32,4 +32,6 @@ interface OAuthSchemaInterface
     public function getHexColor(): ?string;
 
     public function isEnabled(): bool;
+
+    public function shouldCreatingMissingUsers(): bool;
 }
--- a/app/Extensions/OAuth/OAuthSchemaInterface.php
+++ b/app/Extensions/OAuth/OAuthSchemaInterface.php
@@ -33,5 +33,5 @@ interface OAuthSchemaInterface
 
     public function isEnabled(): bool;
 
-    public function shouldCreatingMissingUsers(): bool;
+    public function shouldCreateMissingUsers(): bool;
 }
--- a/app/Http/Controllers/Auth/OAuthController.php
+++ b/app/Http/Controllers/Auth/OAuthController.php
@@ -12,7 +12,6 @@ use Filament\Notifications\Notification;
 use Illuminate\Auth\AuthManager;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
-use Illuminate\Support\Str;
 use Laravel\Socialite\Facades\Socialite;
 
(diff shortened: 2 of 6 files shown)


avaruus1:
Maybe the option could just be called "Perform automated provisioning" or something similar? The current name seems kind of oververbose.

jeffpeng3:
> Maybe the option could just be called "Perform automated provisioning" or something similar? The current name seems kind of oververbose.

Maybe it could be named “Allow user creation”?

lancepioch (approved):

app/Extensions/OAuth/OAuthSchemaInterface.php
```diff
@@ -32,4 +32,6 @@ public function getIcon(): ?string;
     public function getHexColor(): ?string;
 
     public function isEnabled(): bool;
+
+    public function shouldCreatingMissingUsers(): bool;
```
rmartinoscar:
```suggestion
    public function shouldCreateMissingUser(): bool;
```

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -59,33 +63,63 @@ public function callback(Request $request, string $driver): RedirectResponse
...
+                    ->send();
+
+                return redirect()->route('auth.login');
+            }
+
+            $username = $oauthUser->getNickname();
+            $email = $oauthUser->getEmail();
+
+            if (!$username && $email) {
+                $username = str($email)->before('@')->toString() . Str::random(3);
```
rmartinoscar: Wouldn't thid break wings sftp if the email includes a dot ?

app/Http/Controllers/Auth/OAuthController.php
```diff
@@ -59,33 +63,63 @@ public function callback(Request $request, string $driver): RedirectResponse
...
+
+            $data = [
+                'username' => $username,
+                'email' => $email,
+                'oauth' => [
+                    $driver->getId() => $oauthUser->getId(),
+                ],
+            ];
+
+            $user = $this->userCreation->handle($data);
```
rmartinoscar: How about we let the userCreationService handle the generated username based of the email ?

app/Services/Users/UserCreationService.php
```diff
@@ -42,6 +43,16 @@ public function handle(array $data): User
         $isRootAdmin = array_key_exists('root_admin', $data) && $data['root_admin'];
         unset($data['root_admin']);
 
+        if (empty($data['username'])) {
+            $username = str($data['email'])
```
rmartinoscar: The replace should always run not only if the username is empty in case the oauth gives out invalid username

rmartinoscar (changes requested):

PR #1574, opened 2025-08-04
Update egg-vanilla-minecraft.yaml

alwism:
Added ipv6 support by removing the ip lock to ipv4 anycast.

Removed the 0.0.0.0 server-ip entry

It is also recommended to be empty
https://minecraft.wiki/w/Server.properties

### DIFF
--- a/database/Seeders/eggs/minecraft/egg-vanilla-minecraft.yaml
+++ b/database/Seeders/eggs/minecraft/egg-vanilla-minecraft.yaml
@@ -1,7 +1,7 @@
 _comment: 'DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PANEL'
 meta:
   version: PLCN_v2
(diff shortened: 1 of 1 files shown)


alwism:

alwism:
recheck

Boy132:
Thanks, but this is intentionally not empty. If you want ipv6 as well the value would be `::`.

Also, when editing eggs please make sure to do the edits on the panel and then use the export button.

alwism:
Ok no hard feelings was more about awareness because i seached far and wide in my setup to find the issue being this because i have a fully configured dual stack server because of the way i proxy traffic through the system i was expecting something wrong with the proxying.

But now you made me curious i get why any thing is better then nothing but what exactly was the reason for the 0.0.0.0 instead for example using the allocation ?

Boy132 (approved):

PR #1584, opened 2025-08-10
Fix auto deploy docker command not including the container argument

IThundxr:
docker exec -it requires the container as the argument, which currently the command doesn't pass so it gets parsed as: docker exec configure ... in container wings, rather then docker exec wings configure ... in container wings.

### DIFF
--- a/app/Services/Nodes/NodeAutoDeployService.php
+++ b/app/Services/Nodes/NodeAutoDeployService.php
@@ -49,7 +49,7 @@ class NodeAutoDeployService
 
         return sprintf(
             '%s wings configure --panel-url %s --token %s --node %d%s',
-            $docker ? 'docker compose exec -it' : 'sudo',
+            $docker ? 'docker compose exec -it wings' : 'sudo',
             config('app.url'),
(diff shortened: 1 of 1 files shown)


IThundxr:

app/Services/Nodes/NodeAutoDeployService.php
```diff
@@ -49,7 +49,7 @@ public function handle(Request $request, Node $node, ?bool $docker = false): ?st
 
         return sprintf(
             '%s wings configure --panel-url %s --token %s --node %d%s',
-            $docker ? 'docker compose exec -it' : 'sudo',
+            $docker ? 'docker compose exec -it wings' : 'sudo',
```
rmartinoscar:
```suggestion
            $docker ? 'docker compose exec -it $(docker ps --filter "name=wings" --format "{{.Names}}")' : 'sudo',
```
This will try to find any docker container named wings instead of hardcoding `wings` which will always fail when using compose.

rmartinoscar (approved):

PR #1593, opened 2025-08-11
Fix panel version in dockerfile

NerdsCorp:
i believe this should fix the issue with the docker container to show the published version rather than canary version. It sets the app.php to what version specified in the Release. 

Please take a look and let me know if you think this would work. I'm not a PHP Dev but I do have some knowledge on the GitHub container side.


NerdsCorp:

rmartinoscar:
This shouldn't be needed the docker build workflow should run on release/v1.0.0-betaXX instead of main for latest tag.

PR #1603, opened 2025-08-12
Restic support for backups

IThundxr:
This is a set of PRs which implement support for [restic](https://restic.net/) in the current backup system. Restic basically allows for incremental and deduplicated backups which would take up much less space then the current wings/s3 backups.

Related Wings PR: https://github.com/pelican-dev/wings/pull/117

Tasks:
- [ ] Document Restic backups and explain how they work, and what is required to set them up

Tasks Panel:
- [x] Undo changes for local development
- [x] Look into if FilesystemAdapter is actually needed for backups, if not refactor it to use something else
- [x] Ensure all restic backups are deleted properly when a server is deleted
- [x] Restic Repository setting doesn't fill in with the current value

Tasks Wings:
- [x] Test Local restic backups
- [x] Test S3 restic backups
- [x] Either bundle restic in the docker image or document how to properly mount the restic binary into the container
- [x] Implement folder for [restic caching](https://restic.readthedocs.io/en/latest/manual_rest.html#caching) when in a container
- [x] Implement support for restic in RemoveAllServerBackups

@coderabbitai ignore


.dockerignore
```diff
@@ -27,3 +27,4 @@ storage/framework/testing
 storage/framework/views/*.php
 storage/logs/*.log
 vendor
+pelican-data
```
rmartinoscar:
```suggestion
```
The docker volume should never been in the same dir as your dev env

.gitignore
```diff
@@ -27,3 +27,4 @@ public/assets/manifest.json
 filament-monaco-editor/
 _ide_helper*
 /.phpstorm.meta.php
+pelican-data/
```
rmartinoscar:
```suggestion
```
The docker volume should never been in the same dir as your dev env

Dockerfile.dev
```diff
@@ -21,7 +21,7 @@ COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
 # Copy bare minimum to install Composer dependencies
 COPY composer.json composer.lock ./
 
-RUN composer install --no-dev --no-interaction --no-autoloader --no-scripts
```
rmartinoscar:
```suggestion
RUN composer install --no-dev --no-interaction --no-autoloader --no-scripts
```
Please keep --no-dev there's no reason to ship with dev deps

app/Extensions/Backups/BackupManager.php
```diff
@@ -75,7 +77,13 @@ protected function resolve(string $name): FilesystemAdapter
...
         $adapterMethod = 'create' . Str::studly($adapter) . 'Adapter';
         if (method_exists($this, $adapterMethod)) {
-            $instance = $this->{$adapterMethod}($config);
+            // Special handling for Restic adapter, which requires S3 configuration.
+            if ($name === Backup::ADAPTER_RESTIC) {
+                $s3Config = $this->getConfig(Backup::ADAPTER_AWS_S3);
+                $instance = $this->{$adapterMethod}($config, $s3Config);
+            } else {
+                $instance = $this->{$adapterMethod}($config);
+            }
```
rmartinoscar, line 86: Each drivers should have their own config, please don't rely on S3

app/Extensions/Filesystem/ResticFilesystem.php
```diff
@@ -0,0 +1,139 @@
+<?php
```
rmartinoscar: Maybe implementing FilesystemAdapter isn't the way to go about this if you just throw `Unsupported` on every functions

app/Extensions/Backups/BackupManager.php
```diff
@@ -75,7 +77,13 @@ protected function resolve(string $name): FilesystemAdapter
 
         $adapterMethod = 'create' . Str::studly($adapter) . 'Adapter';
         if (method_exists($this, $adapterMethod)) {
-            $instance = $this->{$adapterMethod}($config);
+            // Special handling for Restic adapter, which requires S3 configuration.
+            if ($name === Backup::ADAPTER_RESTIC) {
+                $s3Config = $this->getConfig(Backup::ADAPTER_AWS_S3);
```
rmartinoscar, line 82: What happens if restic->use_s3 is false ?

app/Http/Controllers/Api/Remote/Backups/BackupRemoteResticDetailsController.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
rmartinoscar: Think we should adapt the current Controller to choose between drivers instead of making a whole new one with one line changed

compose.yml
```diff
@@ -1,7 +1,7 @@
 x-common:
```
rmartinoscar: Please revert those changes.

rebuild.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
rmartinoscar: Please revert that as well, i'd recommend using .bashrc alias for that instead of a script.

app/Extensions/Filesystem/ResticFilesystem.php
IThundxr: I think it might be better to just drop the use of FilesystemAdapater for backup adapaters, the panel itself doesn't interface with the backups apart from when deleting s3 backups, so i don't really see why FilesystemAdapater is being used here, it's not like backups are ever stored on the machine the panel is running on anyway, There could probably be a base BackupAdapater Interface which adapaters are required to implement and can use to provide extra info later down the line (for example, s3 details or restic details)

app/Extensions/Backups/BackupManager.php
IThundxr: If use_s3 is false it'll skip providing s3 details to wings and wings will use the repository that should exist at the path provided in the config

app/Extensions/Filesystem/ResticFilesystem.php
IThundxr: I've gone ahead and made that change here, and from some quick testing everything still functions as expected https://github.com/IThundxr/pelican-panel/tree/feat/backup-adapters

app/Services/Servers/ServerDeletionService.php
```diff
@@ -34,21 +38,38 @@ public function withForce(bool $bool = true): self
...
+                    if (!$this->force) {
+                        throw $exception;
+                    }
+
+                    $backup->delete();
+
+                    logger()->warning($exception);
+                }
+            }
+        });
```
Boy132: Wings handles the deletion of backups when a server is deleted. This was intentional done because the task of deleting backups can take some time, depending on the count and size of backups.

IThundxr: Wings currently seems to miss out on s3 backups at the moment, and would require a bigger refactor (alongside bringing back the /restic route) to handle backup deletion on that side

lancepioch:
Thank you very much for you contributions @IThundxr 

However I am going to decline the pull request for the following reasons:

* Restic supports S3 which we already support (or so Google says)
* Not enough public support and/or being well known enough
* Too specific of software to for us to support it indefinitely
* I don't think anybody on our team uses this so we wouldn't be able to guarantee support anyways

PR #1615, opened 2025-08-15
Create custom startup variable field

### LINKED ISSUE #1602: Egg Varaibles are not validated on server creation until final submit
When creating a server, page 2, is egg variables. They used to validate when you clicked Next Step, but now they don't validate until you get to the last page and click Create Server, requiring the user to go back a page. They are also no longer highlighted when they're invalid. 

Image

Boy132:
Also closes #1602

### DIFF
--- a/app/Enums/StartupVariableType.php
+++ b/app/Enums/StartupVariableType.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace App\Enums;
+
+enum StartupVariableType: string
+{
+    case Text = 'text';
+    case Select = 'select';
+    case Toggle = 'toggle'; // TODO: add toggle to blade view
+}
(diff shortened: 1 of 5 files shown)


app/Enums/StartupVariableType.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Enums;
+
+enum StartupVariableType: string
+{
+    case Text = 'text';
+    case Select = 'select';
+    case Toggle = 'toggle'; // TODO: add toggle to blade view
```
rmartinoscar, line 9: Should we wait for v4 since they made it a blade component ?

notAreYouScared: Creating servers is also broken in v4, this probably fixes it in both

rmartinoscar:
I'd like the possibility of having a `multiple` rule and we should set this otherwise it will always use the first option when not selected instead of null.
```php
$this->selectablePlaceholder(fn (StartupVariable $component) => !$component->isRequired())
```

app/Enums/StartupVariableType.php
Boy132: > Creating servers is also broken in v4, this probably fixes it in both

Yes.

rmartinoscar (approved):

PR #1626, opened 2025-08-20
fix: Wrong conversion if decimal prefix selected

SaurFort:
When using the decimal prefix and setting 16 GB of RAM, the panel displayed 16.79 GB (without overhead) instead of 16 GB (without overhead).
This was due to the using of a binary multiplier (`2^20`), which is only correct for MiB/GiB.

So I've used the `config('panel.use_binary_prefix')` to check what multiplier should be used.

This is a visual fix, but I think it's better like that.

### DIFF
--- a/app/Filament/Server/Widgets/ServerOverview.php
+++ b/app/Filament/Server/Widgets/ServerOverview.php
@@ -68,7 +68,7 @@ class ServerOverview extends StatsOverviewWidget
         }
 
         $latestMemoryUsed = collect(cache()->get("servers.{$this->server->id}.memory_bytes"))->last(default: 0);
-        $totalMemory = $this->server->memory * (config('panel.use_binary_prefix') ? 2 ** 20 : 1000 * 1000);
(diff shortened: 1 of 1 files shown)


SaurFort:

SaurFort:
recheck

app/Filament/Server/Widgets/ServerOverview.php
```diff
@@ -68,7 +68,7 @@ public function memoryUsage(): string
         }
 
         $latestMemoryUsed = collect(cache()->get("servers.{$this->server->id}.memory_bytes"))->last(default: 0);
-        $totalMemory = $this->server->memory * 2 ** 20;
+        $totalMemory = $this->server->memory * (config('panel.use_binary_prefix') ? 2 ** 20 : 1000 * 1000);
```
Boy132: Looks good, just a nitpick for the sack of consistency.

```suggestion
        $totalMemory = $this->server->memory * (config('panel.use_binary_prefix') ? 1024 * 1024 : 1000 * 1000);
```

PR #1641, opened 2025-08-30
Fix activityLog permission name

### LINKED ISSUE #1637: Permission for seeing activity log ips is named wrong
https://github.com/pelican-dev/panel/blob/4aae2562ea65707439a7e036194de7dbd1e6371f/app/Models/Role.php#L47-L49

This should be `activityLog` instead of `activity`.
+ add migration to fix existing roles.

rmartinoscar:
Closes #1637

### DIFF
--- a/database/migrations/2025_08_30_111308_fix_activity_log_permission_name.php
+++ b/database/migrations/2025_08_30_111308_fix_activity_log_permission_name.php
@@ -1,0 +1,23 @@
+<?php
+
+use App\Models\Permission;
+use Illuminate\Database\Migrations\Migration;
+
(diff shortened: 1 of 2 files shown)


database/migrations/2025_08_30_111308_fix_activity_log_permission_name.php
Boy132, line 1: The `Permission` model is for subuser permissions, not admin roles. The spatie plugin handles role permissions.

Boy132 (changes requested):

Boy132 (approved):

PR #1650, opened 2025-09-03
Rename api key prefixes for better clarity

Boy132:
_This does not change existing api keys are any way._

### DIFF
--- a/app/Models/ApiKey.php
+++ b/app/Models/ApiKey.php
@@ -216,7 +216,7 @@ class ApiKey extends PersonalAccessToken
     {
         Assert::oneOf($type, [self::TYPE_ACCOUNT, self::TYPE_APPLICATION]);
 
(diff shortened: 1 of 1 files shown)


rmartinoscar:
I think it makes sense the way it already is

Boy132:
> I think it makes sense the way it already is

Both `plcn` and `peli` are just short forms for `pelican`. How would you know which one is which?

notAreYouScared:
I have to agree with Boy on this one... I like it.

rmartinoscar:
> > I think it makes sense the way it already is
> 
> Both `plcn` and `peli` are just short forms for `pelican`. How would you know which one is which?

cn sounds like client to me but ok
Btw this means we need to change
https://github.com/pelican-dev/wings/blob/9fcb7d98c952e5edfad400173ff3d35b2632b202/cmd/configure.go#L186

PR #1651, opened 2025-09-03
Filament v4 🎉 

notAreYouScared:
Well... This should be fun ❤️

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6eb38a9c85241824ed162fa2557e8ac4",
+    "content-hash": "09abbfbce5021c591de46868e3abfc3f",
     "packages": [
         {
             "name": "amphp/amp",


SirEndii:
long journey this was

Boy132:
@coderabbitai review

app/Filament/Server/Resources/Files/Pages/EditFiles.php
```diff
@@ -120,15 +123,38 @@ public function form(Form $form): Form
...
-                            ->options(EditorLanguages::class)
+                            ->options(Language::class)
                             ->selectablePlaceholder(false)
-                            ->afterStateUpdated(fn ($state) => $this->dispatch('setLanguage', lang: $state))
-                            ->default(fn () => EditorLanguages::fromWithAlias(pathinfo($this->path, PATHINFO_EXTENSION))),
-                        MonacoEditor::make('editor')
+                            ->default(fn () => match (pathinfo($this->path, PATHINFO_EXTENSION)) {
+                                'cc', 'hpp' => Language::Cpp,
+
+                                'css', 'scss' => Language::Css,
```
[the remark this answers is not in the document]
rmartinoscar: That was the use of the previous `EditorLanguages` Enum.

phpstan.neon
```diff
@@ -14,6 +14,8 @@ parameters:
     ignoreErrors:
         - identifier: argument.templateType
         - identifier: missingType.generics
+        - '#Call to an undefined method Filament\\Forms\\Components\\TextInput::hintCopy\(\)#'
+        - '#Call to an undefined method Filament\\Forms\\Components\\Textarea::hintCopy\(\)#'
```
lancepioch: Get all of them.

```suggestion
        - '#Call to an undefined method Filament\\Forms\\Components\\(.+)::hintCopy\(\)#'
```

app/Models/User.php
```diff
@@ -315,6 +315,12 @@ public function subServers(): BelongsToMany
         return $this->belongsToMany(Server::class, 'subusers');
     }
 
+    /** @return array<mixed> */
+    public function getCustomization(): array
+    {
+        return json_decode($this->customization, true) ?? [];
```
lancepioch: Shouldn't customization be a an array cast so you don't need `json_decode`?

app/Models/User.php
```diff
@@ -439,9 +445,44 @@ public function canAccessTenant(Model $tenant): bool
...
+    {
+        return $this->email;
+    }
+
+    /**
+     * @return array<string>|null
+     */
+    public function getAppAuthenticationRecoveryCodes(): ?array
+    {
+        return $this->mfa_app_recovery_codes;
```
lancepioch, line 475: Should be an array cast so you shouldn't need this.

app/Services/Subusers/SubuserCreationService.php
```diff
@@ -29,29 +31,29 @@ public function __construct(
...
+     * @throws DataValidationException
+     * @throws ServerSubuserExistsException
+     * @throws UserIsServerOwnerException
+     * @throws Throwable
      */
     public function handle(Server $server, string $email, array $permissions): Subuser
     {
         return $this->connection->transaction(function () use ($server, $email, $permissions) {
-            $user = User::query()->where('email', $email)->first();
+            $user = User::withoutGlobalScopes()->where('email', $email)->first();
```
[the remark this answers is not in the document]
rmartinoscar: Users can't be softDeleted ?

app/Listeners/Auth/TwoFactorListener.php
```diff
@@ -7,6 +7,7 @@
 
 class TwoFactorListener
 {
+    // TODO: add event to filament
```
lancepioch, line 10: Do we need to do this now?

lancepioch (approved):

app/Listeners/Auth/TwoFactorListener.php
Boy132: Yes-ish. If the event isn't called there is no activity logged.

app/Models/User.php
Boy132: Wdym? It is an array cast?
https://github.com/pelican-dev/panel/blob/948ea1aa04e2369b355a07a6c7df4155253cc624/app/Models/User.php#L176

Or are you talking about the function itself? It's needed from an interface.

app/Models/User.php
rmartinoscar: Yes and even better you should be able to specify the key as a function arg, if there's none you return the full array otherwise you return the value of that key so we don't have `getCustomization()['key']` but `getCustomization('key')`

Boy132: It throws an `App\Models\User::getCustomization(): Return value must be of type array, string returned` when not using `json_decode` for some reason.

PR #1659, opened 2025-09-04
Fix finish & add translation for Installer title

rmartinoscar:


app/Livewire/Installer/PanelInstaller.php
```diff
@@ -79,14 +84,18 @@ protected function getFormSchema(): array
                 SessionStep::make(),
             ])
                 ->persistStepInQueryString()
-                ->nextAction(fn (Action $action) => $action->keyBindings('enter'))
+                ->nextAction(function (Action $action) {
+                    $action
+                        ->label(trans('installer.next_step'))
```
Boy132, line 89: Is there no translation from filament?

notAreYouScared: https://github.com/filamentphp/filament/blob/21e01fe9710064d7feed198fd37a527ec9127072/packages/schemas/resources/lang/en/components.php#L14

Looks like its just next

rmartinoscar: There is but its `Next` not `Next step` and we override it in Create Node & Server so i thought we'd do it for Installer as well.
https://github.com/filamentphp/filament/blob/3.x/packages/forms/resources/lang/en/components.php#L512

Boy132: Hmm okay, sure why not

Boy132 (approved):

Issue #1526, opened 2025-07-17
Docker Image reports Panel-Version as canary


mristau:
I'm using Pelican with docker and in the panel it will always say Version canary instead ov 1.0.0-betaxx

<img width="1071" height="247" alt="Image" src="https://github.com/user-attachments/assets/6cb9064d-8847-452b-a208-c0814da9c3e6" />

I'm currently using this image `ghcr.io/pelican-dev/panel:v1.0.0-beta22`

It will not report as outdated if a new release is created

PR #1663, opened 2025-09-05
Fix Docker versioning 

NerdsCorp:
here is the workflow fix for the docker version of the panel.

you can use my image for testing the image and workflow ghcr.io/nerdscorp/panel:latest

this will use the Release name and update the version from canary to whatever the release name is. i set up the newest relese using this as well as a fake one seen below.

image

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -134,6 +134,16 @@ jobs:
           docker push localhost:5000/base-php:arm64
           rm base-php-arm64.tar base-php-amd64.tar
           
+      - name: Update version in config/app.php (tag)
+        if: "github.event_name == 'release' && github.event.action == 'published'"
+        run: |
+          sed -i "s/'version' => 'canary',/'version' => '${{ steps.build_info.outputs.version_tag }}',/" config/app.php
+          
+      - name: Update version in config/app.php (main)
+        if: "github.event_name == 'push' && contains(github.ref, 'main')"
+        run: |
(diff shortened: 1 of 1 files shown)


NerdsCorp:
This closes #1526

NerdsCorp:
image


image


This is what it now shows.

.github/workflows/docker-publish.yml
```diff
@@ -134,6 +134,16 @@ jobs:
...
+      - name: Update version in config/app.php (tag)
+        if: "github.event_name == 'release' && github.event.action == 'published'"
+        run: |
+          sed -i "s/'version' => 'canary',/'version' => '${{ steps.build_info.outputs.version_tag }}',/" config/app.php
+          
+      - name: Update version in config/app.php (main)
+        if: "github.event_name == 'push' && contains(github.ref, 'main')"
+        run: |
+          sed -i "s/'version' => 'canary',/'version' => 'dev-${{ steps.build_info.outputs.short_sha }}',/" config/app.php
+
```
Boy132: For `main` the version should remain `canary`.

rmartinoscar: I like the dev-{commit} better, in fact that's what i have for my wings make.

Boy132: Then we would need to A) update the panel because there are `canary` checks and B) also do that for the non-dockerized version.

NerdsCorp: Removed main as requested.

PR #1676, opened 2025-09-06
Replace `str_random` with `Str::random`

notAreYouScared:
rabbit suggested due to str_random being deprecated


rmartinoscar (commented):
Why did public got touched ?

PR #1684, opened 2025-09-09
Small v4 fixes

Boy132:
- Fix redirect after using power action on server list
- Fix `Create Schedule` page
- Fix totp migration by not migrating the old secret - **this means 2fa will be disabled for all users and they have to set it up again!**

### DIFF
--- a/database/migrations/2025_07_22_091435_update_users_totp.php
+++ b/database/migrations/2025_07_22_091435_update_users_totp.php
@@ -1,6 +1,5 @@
 <?php
 
-use App\Models\User;
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
@@ -16,16 +15,6 @@ return new class extends Migration
             $table->text('mfa_app_secret')->nullable();
             $table->text('mfa_app_recovery_codes')->nullable();
             $table->boolean('mfa_email_enabled')->default(false);
-        });
-
-        User::chunk(100, function ($users) {
-            foreach ($users as $user) {
-                $user->update([
-                    'mfa_app_secret' => $user->use_totp ? $user->totp_secret : null,
-                    'mfa_app_recovery_codes' => null,
-                    'mfa_email_enabled' => false,
-                ]);
-            }
         });
 
         Schema::table('users', function (Blueprint $table) {
--- a/app/Filament/App/Resources/Servers/Pages/ListServers.php
+++ b/app/Filament/App/Resources/Servers/Pages/ListServers.php
@@ -216,7 +216,7 @@ class ListServers extends ListRecords
 
             cache()->forget("servers.$server->uuid.status");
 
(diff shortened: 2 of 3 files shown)


database/migrations/2025_07_22_091435_update_users_totp.php
```diff
@@ -18,16 +17,6 @@ public function up(): void
             $table->boolean('mfa_email_enabled')->default(false);
         });
 
-        User::chunk(100, function ($users) {
-            foreach ($users as $user) {
-                $user->update([
-                    'mfa_app_secret' => $user->use_totp ? $user->totp_secret : null,
-                    'mfa_app_recovery_codes' => null,
-                    'mfa_email_enabled' => false,
```
rmartinoscar, line 26: https://github.com/pelican-dev/panel/blob/1900c04b7172961349c2461c1ecd5236ada3ef94/app/Models/User.php#L62
https://github.com/pelican-dev/panel/blob/1900c04b7172961349c2461c1ecd5236ada3ef94/app/Models/User.php#L486-L489
Shouldn't this be nullsafed now that this table doesn't exists ?

Boy132: The table still exists? And `mfa_email_enabled` defaults to `false`, it's not nullable.

app/Models/Schedule.php
```diff
@@ -107,6 +110,13 @@ protected function casts(): array
         ];
     }
 
+    protected function status(): Attribute
+    {
+        return Attribute::make(
+            get: fn () => !$this->is_active ? ScheduleStatus::Inactive : ($this->is_processing ? ScheduleStatus::Processing : ScheduleStatus::Active),
```
rmartinoscar, line 116: Why don't we fully replace is_active & is_processing with the status enum ?

Boy132: We can do that in a separate PR if we want to. I basically just copied this from the backup status.

lajczi:
You can also fix this translation

image

lancepioch (approved):

app/Filament/Server/Resources/Files/Pages/ListFiles.php
```diff
@@ -86,10 +86,11 @@ public function table(Table $table): Table
...
 
         return $table
-            ->paginated()
+            ->paginated(['25', '50', '100', '150', '200'])
             ->paginationMode(PaginationMode::Simple)
+            ->defaultPaginationPageOption(50)
+            ->deferLoading()
             ->searchable()
-            ->defaultPaginationPageOption(25)
             ->query(fn () => $files->orderByDesc('is_directory'))
```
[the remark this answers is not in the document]
rmartinoscar: Rabbit seems legit and `$files` isn't used anywhere else so i guess we can move it to query()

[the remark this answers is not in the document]
Boy132: This is bascially used as "caching". So even if the query is run multiple times it doesn't actually runs multiple wings requests.

lancepioch (approved):

PR #1700, opened 2025-09-10
Add RequireApplicationApiKey middleware & remove rootAdmin bypass

rmartinoscar:
Before (non root user)
image

Before (root user)
image

After (regardless of root)
image


rmartinoscar:
Tests for these are a mess.

PR #1702, opened 2025-09-11
Remove username rules and allow to change it in profile

Boy132:
There is no reason to restrict the username or to not allow users to edit theirs.

#FreeUsername


rmartinoscar:
Wings won't be able to parse them for sftp

JoanFo1456:
It would be nice as you freed username, to free passwords or to allow admins to select the security of them, as perhaps an admin doesn't care about passwords or he wants more robust ones...

Boy132:
> Wings won't be able to parse them for sftp

It works fine. :) Even something more crazy like `admin#@/-$` works.

grafik
grafik

Leon-JavaScript:
Will is be possible for admins to restrict users being able to change their own username? (like with the server description)
I think some people want to have clean username lists (or even specific username patterns) e.g. when using the panel with an external system.

lancepioch (approved):

PR #1703, opened 2025-09-12
Fix server owner permissions

danielkurek:
I have stumbled upon a possible bug Server owner cannot start/restart/stop their own server if they have assigned some admin role without server update permission. I would expect that being a server owner would allow you to do anything with the server.

I am not sure if this is just a bug or if it is intentional. I have not read most of the codebase so I do not know if this change has some unwanted behaviour.


danielkurek:

PR #1704, opened 2025-09-13
Fixed webhooks on v4 and nested values

JoanFo1456:
This pull request repairs the variables such as {{allocation.id}} not working, and editing the old webhook example to match the exact one being sent on updated: Server.


app/Models/WebhookConfiguration.php
```diff
@@ -191,45 +190,209 @@ function ($matches) use ($replacement) {
...
             'node_id' => 1,
-            'name' => 'Eagle',
-            'description' => 'This is an example server description.',
-            'skip_scripts' => false,
+            'name' => 'Example Server',
             'owner_id' => 1,
-            'memory' => 2048,
-            'swap' => 128,
-            'disk' => 10240,
+            'memory' => 6411,
```
rmartinoscar:
```suggestion
            'memory' => 6144,
```
6Gb ?

resources/views/filament/components/webhooksection.blade.php
```diff
@@ -5,9 +5,6 @@
     :compact="$isCompact()"
     :content-before="$isFormBefore()"
     :description="$getDescription()"
-    :footer-actions="$getFooterActions()"
```
rmartinoscar, line 8: We should update the blade rather then just remove random stuff.

JoanFo1456: It does nothing to the blade removing it...
It's literally the same one as before

rmartinoscar: Ye but its not up to date with filament v4 section's blade

JoanFo1456: Okay, i'll update it then when i have some time... Thanks Martin

rmartinoscar (approved):

PR #1710, opened 2025-09-15
Remove table row icons

notAreYouScared:
Removes the table row icons that are in every row; we have table headers, and that should be sufficient. It would be nice to have the icon in the header, but that does not seem like an option at the moment.

### DIFF
--- a/app/Filament/Admin/Resources/ApiKeys/ApiKeyResource.php
+++ b/app/Filament/Admin/Resources/ApiKeys/ApiKeyResource.php
@@ -77,6 +77,7 @@ class ApiKeyResource extends Resource
             ->columns([
                 TextColumn::make('key')
                     ->label(trans('admin/apikey.table.key'))
+                    ->icon('tabler-clipboard-text')
                     ->state(fn (ApiKey $key) => $key->identifier . $key->token)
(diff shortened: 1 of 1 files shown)


app/Filament/Admin/Resources/ApiKeys/ApiKeyResource.php
```diff
@@ -77,7 +77,6 @@ public static function defaultTable(Table $table): Table
             ->columns([
                 TextColumn::make('key')
                     ->label(trans('admin/apikey.table.key'))
-                    ->icon('tabler-clipboard-text')
```
Boy132: I think in this case the icon should stay so it's clear you can click to copy. Thoughts?

notAreYouScared: I would agree with that.

notAreYouScared:
@coderabbitai generate docstrings

lancepioch:
But I think I like the icons in every row 🤔

notAreYouScared:
It's just so repetitive. 

Example, 

List Eggs - Each egg name has an egg icon before it... why? 
We know they're eggs... The same with server count, and any other thing that's a count.

PR #1722, opened 2025-09-19
Make models strict and enable auto eager loading

Boy132:
- Prevents lazy loading, silently discarding attributes, and accessing missing attributes
- Enables automatic eager loaded when accessing model relationshoips

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -33,6 +33,7 @@ use Filament\Support\Facades\FilamentColor;
 use Filament\Support\Facades\FilamentView;
 use Filament\View\PanelsRenderHook;
 use Illuminate\Config\Repository;
+use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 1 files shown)


Boy132:
@coderabbitai review

app/Models/EggVariable.php
```diff
@@ -52,6 +52,8 @@ class EggVariable extends Model implements Validatable
      */
     protected $guarded = ['id', 'created_at', 'updated_at'];
 
+    public ?string $server_value = null;
```
rmartinoscar, line 55: What is this for ?

Boy132: `variables` relation on the server returns the egg variables + the server value.
With strict models the property needs to exist in the model. Similar to the sum properties in the node model.

https://github.com/pelican-dev/panel/blob/c10280af4b9763e965e9e590980f68d29c094359/app/Models/Server.php#L310-L321

PR #1765, opened 2025-09-30
Create shift workflow

rmartinoscar:
This is a draft cause its not updating Laravel as it should see https://github.com/pelican-dev/panel/pull/1790
It also doesn't `git add .` before pushing so filament assets aren't published.


lancepioch (approved):

.github/workflows/shift.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - cron: "0 0 * * 5"
+
+jobs:
+  shift:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Shift
+        run: |
+          curl -X POST -s -retry 5 -m 60 --fail-with-body \
```
rmartinoscar, line 15:
```suggestion
          curl -X POST -s --retry 5 -m 60 --fail-with-body \
```

PR #1771, opened 2025-10-02
Task Offset

JoanFo1456:
When you make a task for the first time, it shouldn't have an offset, seems that it broke with the update.
This pr what it does is to fix first action to not show the offset field on creation and make sure it will always be 0 when reordered or deleted a task.
I did this at 2am, so I'll be grateful on feedback.


JoanFo1456:
I'm not doing Rabbit suggestion, as when I did it, I got from phpstan that the ? is unnecessary from both new things.

Boy132:
I don't understand the purpose of the pr? The `time_offset` will be ignored on the first task anyways? So all this does is "cleaning up" the database value?

JoanFo3123:
> I don't understand the purpose of the pr? The `time_offset` will be ignored on the first task anyways? So all this does is "cleaning up" the database value?

It’s fixing that the first task won’t show the field to add offset and in case you move tasks, to remove it from the first one, but I suppose if it’s being ignored we can just discard this…

JoanFo1456:
> > I don't understand the purpose of the pr? The `time_offset` will be ignored on the first task anyways? So all this does is "cleaning up" the database value?
> 
> It’s fixing that the first task won’t show the field to add offset and in case you move tasks, to remove it from the first one, but I suppose if it’s being ignored we can just discard this…

This is me, I just used the other account as it was the only available on my phone ^

PR #1773, opened 2025-10-02
Add sudo to crontab command

Boy132:


app/Livewire/Installer/Steps/QueueStep.php
```diff
@@ -48,7 +48,7 @@ public static function make(PanelInstaller $installer): Step
                     ->label(new HtmlString(trans('installer.queue.fields.crontab')))
                     ->disabled()
                     ->hintCopy()
-                    ->default('(crontab -l -u www-data 2>/dev/null; echo "* * * * * php ' . base_path() . '/artisan schedule:run >> /dev/null 2>&1") | crontab -u www-data -')
+                    ->default('(sudo crontab -l -u www-data 2>/dev/null; echo "* * * * * php ' . base_path() . '/artisan schedule:run >> /dev/null 2>&1") | sudo crontab -u www-data -')
```
QuintenQVD0, line 51: sudo might not always be installed.

Debian for example doesn't have it installed by default

JoanFo1456: > sudo might not always be installed.
> 
> Debian for example doesn't have it installed by default

But for installing pelican you need a bit of brain, if it tells you "sudo not found" you have eyes to remove it from the command...

rmartinoscar: We already use sudo
https://github.com/pelican-dev/panel/blob/19103b16b865f7ea213721317ed9a1f8536c7b59/app/Livewire/Installer/Steps/QueueStep.php#L58

https://github.com/pelican-dev/panel/blob/19103b16b865f7ea213721317ed9a1f8536c7b59/app/Services/Nodes/NodeAutoDeployService.php#L53

rmartinoscar (approved):

PR #1783, opened 2025-10-08
Remove delete action on last allocation for users

### LINKED ISSUE #1740: Prevent users from deleting "default" allocations
### Discussed in https://github.com/pelican-dev/panel/discussions/1616

<sup>Originally posted by **AV3RG** August 15, 2025</sup>
Users can delete primary allocations and other allocations for the server that have been assigned to them, it becomes an issue when they accidentally delete these allocations and then open up support tickets which have to be managed, allocations that were assigned to them were not meant to be deleted in the first place as they are necessary for the server for function for some eggs.

I understand that a counter argument can be to let them self allocate ports but then that also creates issues if you have like a script that handles reverse proxy creation for each server from billing panel and also if you use different network subnets for different clients to let them privately network in their bungee and stuff. Letting them self allocate will also allow them to allocate themselves to someone else's network</div>

notAreYouScared:
Closes #1740

### DIFF
--- a/app/Filament/Server/Resources/Allocations/AllocationResource.php
+++ b/app/Filament/Server/Resources/Allocations/AllocationResource.php
@@ -76,8 +76,9 @@ class AllocationResource extends Resource
             ])
             ->recordActions([
                 DetachAction::make()
+                    ->visible(fn () => user()?->isRootAdmin() || $server->allocations()->count() !== 1)
                     ->authorize(fn () => user()?->can(Permission::ACTION_ALLOCATION_DELETE, $server))
-                    ->label(trans('server/network.delete'))
+                    ->label(trans('filament-actions::delete.single.modal.actions.delete.label'))
                     ->icon('tabler-trash')
                     ->action(function (Allocation $allocation) {
                         Allocation::query()->where('id', $allocation->id)->update([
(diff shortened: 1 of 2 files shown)


JoanFo3123:
I think adding an option is best, not making users directly not able to do it. Like on settings to select to allow or not the deleting.
> When I did this with Quinten we thought of this, never did it, but it’s a better solution

notAreYouScared:
While an option woukd be nice, I dont wabt to have 40 options to allow a user to do this, that, this other thing, etc. The server either gets an allocation or multiple, or gets none.

JoanFo3123:
> While an option woukd be nice, I dont wabt to have 40 options to allow a user to do this, that, this other thing, etc. The server either gets an allocation or multiple, or gets none.

Simple toggle, allow removing primary allocation on admin side 
> Like an admin setting on settings

JoanFo3123:
> > While an option woukd be nice, I dont wabt to have 40 options to allow a user to do this, that, this other thing, etc. The server either gets an allocation or multiple, or gets none.
> 
> Simple toggle, allow removing primary allocation on admin side

> like a general setting

notAreYouScared:
And where would you like said setting saved? It has to be stored somewhere.

JoanFo3123:
> And where would you like said setting saved? It has to be stored somewhere.

Like a normal setting, you guys have some of them, .env or migration, but on the settings tab.
If needed I can do it when I get home.

notAreYouScared:
No thanks. I feel its fine how it is.

PR #1789, opened 2025-10-09
Topbar removed if using sidebar

JoanFo1456:
Seems that on Filament 4.1 they introduced this option, so you can have a fully topbar-sidebar thing, not like the actual hybrid.
New sidebar when topbar is gone:
image
image


JoanFo1456:
I guess rabbit is sniffing something, as topbar is completely new .-.
> Just saw what he meant, let me check...

Issue #1744, opened 2025-09-24
Add a 'What's new' button on the admin dashboard when an update is available


Boy132:
### Discussed in https://github.com/pelican-dev/panel/discussions/830

<div type='discussions-op-text'>

<sup>Originally posted by **JustRed23** December 13, 2024</sup>
As a dev myself, I want to know what changed when the panel gets an update, I need to go to this github and look through the releases section to do that now. Would it be a nice idea to add a button that compares the current version to the newest version and makes a github link like [this](https://github.com/pelican-dev/panel/compare/v1.0.0-beta14...v1.0.0-beta16)?</div>

PR #1792, opened 2025-10-10
Add changelog preview when a new update is available

LetterN:
implements #1744

image


LetterN:
how about this? 
image

changelog is collapsed by default

lancepioch:
This looks great @LetterN - can you quick add the translation key and it's good to go!

PR #1810, opened 2025-10-14
Added Korean Translation

zzunipark:


zzunipark:
Note: The duplicate 'flags' entry also exists in the original English source file.
I kept it as-is to stay consistent with the original translation, so this duplication is intentional and not an error in my translation.

Also, updated the translation of "dead" from '죽음' to '중단됨' for a more natural and context-appropriate meaning.

Boy132:
Thanks, but translations are managed via crowdin: https://crowdin.com/project/pelican-dev

notAreYouScared:
I'd also like to thank you for the translations. I have imported the language files into Crowdin for you. :)

PR #1817, opened 2025-10-17
Remove `hintCopy` from EditNode autosize configuration

PalmarHealer:
It has been removed since it’s not available, and unfortunately, there is no good replacement for it.

`Method 'hintCopy' not found in \Filament\Forms\Components\Textarea`

It would be possible to use the `TextInput`, but that didn’t work for me either. Therefore, I suggest removing it to avoid confusion, as it’s not needed at the moment.

Maybe a `// TODO: Add copy button` comment could be added so it can be revisited later.

### DIFF
--- a/app/Filament/Admin/Resources/Nodes/Pages/EditNode.php
+++ b/app/Filament/Admin/Resources/Nodes/Pages/EditNode.php
@@ -591,7 +591,6 @@ class EditNode extends EditRecord
                                                     ->label(trans('admin/node.auto_command'))
                                                     ->readOnly()
                                                     ->autosize()
-                                                    ->hintCopy()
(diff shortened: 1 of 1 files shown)


JoanFo1456:
LGTM on demo, I don't know why are you facing that but it works on demo.

PalmarHealer:
Okay, its also working for me in prod. so maybe I did a thing with my dev env

rmartinoscar:
Its a macro we made not something from Filament itself.
https://github.com/pelican-dev/panel/blob/1a4fa5e67a7709ade51e968d046c6748e34d5589/app/Providers/AppServiceProvider.php#L153-L156

PR #1822, opened 2025-10-21
Fixed console duplicating

### LINKED ISSUE #1241: Server console text duplication on navigate back
### Current Behavior

When navigating back to the server console page from any other page using the "back" button it duplicates the console window and leaves 1 of them in a broken state.

### Expected Behavior

When navigating back to the server console page from any other page using the back button it should not duplicate the console window.

### Steps to Reproduce

When you have a server running and see content in the console and you click on a sidebar tab (for example network) and navigate to that page and then navigate back to the server console page (using the back button in the top left of the browser next to the URL).  the console window will be duplicated. 1 of them will work but the other  one will be in a broken state unable to scroll through either.

### Panel Version

1.0.0-beta19

### Wings Version

1.0.0-beta11

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

JoanFo1456:
Closes #1241


Boy132 (changes requested):
Rabbit is right, in general.

JoanFo1456:
> Rabbit is right, in general.

Doesn't look like it, tested with both demo and the panel before and after, doesn't have any memory leak.
Martin was with me when testing
> Also, for example is not duplicating messages, so I guess Rabbit is not right

JoanFo1456:
#1990 Fixes the same as this pr

PR #1825, opened 2025-10-23
Add toggle for externally managed users

### LINKED ISSUE #1728: Readonly user profile
### Discussed in https://github.com/pelican-dev/panel/discussions/1475

<sup>Originally posted by **InvalidJoker** June 30, 2025</sup>
Add a setting for admins to choose whether users can edit their own profiles or only admins can do so (via the admin UI). This can be useful for syncing Pelican with systems like WHMCS, removing the need to constantly check if both are in sync.</div>

Boy132:
Closes #1728

grafik
grafik


app/Filament/Pages/Auth/EditProfile.php
```diff
@@ -94,19 +94,22 @@ public function form(Schema $schema): Schema
                             ->icon('tabler-user-cog')
                             ->schema([
                                 TextInput::make('username')
+                                    ->disabled(fn (User $user) => $user->is_managed_externally)
```
rmartinoscar, line 96:
```suggestion
                                    ->readOnly(fn (User $user) => $user->is_managed_externally)
```
I think this would be better since you can still select it but you might need a [`dehydrated(false)`](https://filamentphp.com/docs/4.x/forms/text-input#making-the-field-read-only) with that.

app/Models/User.php
```diff
@@ -174,6 +178,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     protected function casts(): array
     {
         return [
+            'is_managed_externally' => 'boolean',
```
rmartinoscar, line 182:
```suggestion
```
This is unnecessary

database/Factories/UserFactory.php
```diff
@@ -26,6 +26,7 @@ public function definition(): array
 
         return [
             'external_id' => null,
+            'is_managed_externally' => false,
```
rmartinoscar, line 29:
```suggestion
```
This is unnecessary

app/Models/User.php
Boy132: It's not. This makes sure the value is `true`/ `false` instead of `1`/`0`. So you don't have to manually cast to a bool every time.

database/Factories/UserFactory.php
Boy132: We also specify other values that have a default value. For the sake of consitency I added it.

app/Filament/Pages/Auth/EditProfile.php
Boy132: You can also select a disabled field. Also, the disabled makes it visually clear that you can't edit it.

notAreYouScared: I would prefer the disabled over the readOnly

app/Http/Requests/Api/Client/Account/UpdateEmailRequest.php
rmartinoscar, line 11: How about ? https://github.com/pelican-dev/panel/blob/611b8649e057e2dd6d451b45cbd8735478c7e084/app/Http/Requests/Api/Client/Account/UpdateUsernameRequest.php#L11

rmartinoscar:
I agree with Lance if we are ever gonna add timestamp, might aswell do it now.

notAreYouScared:
> I agree with Lance if we are ever gonna add timestamp, might aswell do it now.

I still see no reason to have a timestamp... It's either on or off.

PR #1828, opened 2025-10-23
Add archive extension selection

notAreYouScared:
Requires [wings/#123](https://github.com/pelican-dev/wings/pull/123)

Allows the user to pick an archive type, 

image

### DIFF
--- a/app/Http/Requests/Api/Client/Servers/Files/CompressFilesRequest.php
+++ b/app/Http/Requests/Api/Client/Servers/Files/CompressFilesRequest.php
@@ -22,7 +22,7 @@ class CompressFilesRequest extends ClientApiRequest
             'files' => 'required|array',
             'files.*' => 'string',
             'name' => 'sometimes|nullable|string',
-            'extension' => 'sometimes|in:zip,tgz,tbz2,txz,"tar.xz","tar.bz2","tar.gz"',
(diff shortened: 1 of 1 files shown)


app/Http/Requests/Api/Client/Servers/Files/CompressFilesRequest.php
```diff
@@ -22,6 +22,7 @@ public function rules(): array
             'files' => 'required|array',
             'files.*' => 'string',
             'name' => 'sometimes|nullable|string',
+            'extension' => 'sometimes|in:zip,tgz,tbz2,txz,"tar.xz","tar.bz2","tar.gz"',
```
rmartinoscar:
```suggestion
            'extension' => 'sometimes|in:zip,tgz,tar.gz,txz,tar.xz,tbz2,tar.bz2',
```

rmartinoscar (changes requested):

app/Http/Requests/Api/Client/Servers/Files/CompressFilesRequest.php
notAreYouScared: Im 98.99% sure the order does not make a difference.

Boy132 (approved):

Issue #1829, opened 2025-10-25
Startup boolean value error


rurueuh:
### Current Behavior

Only on "client" server view, boolean value always fail to update (only true to false)

<img width="400" height="86" alt="Image" src="https://github.com/user-attachments/assets/493bf24c-e638-4545-84c4-48c8025c9797" />

### Expected Behavior

Should be able to change value true or false without any error

### Steps to Reproduce

- simple install
- start gmod server
- try to disable lua refresh

### Panel Version

1.0.0-beta27

### Wings Version

1.0.0-beta18

### Games and/or Eggs Affected

any with boolean value

### Error Logs

```bash

```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue.
- [x] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

rmartinoscar:
<img width="1645" height="218" alt="Image" src="https://github.com/user-attachments/assets/1e6427c3-0254-42b2-ba92-ee5c40119c8f" />
Toggle true works but not false

Boy132:
This is not a bug, that's expected behaviour. If something is `required` it can't be false, null or empty.
If anything, the gmod eggs need to be updated and the `required` removed from the variable.

rurueuh:
That looks not only the GMod egg but a lot of other eggs (Avorion, Ark, GMod, Brickadia, Assetto and many others, I checked only 7-8 eggs).
Should we fix them one by one, or find a code solution?
And, if a code solution, would you prefer to force "True"/"False" rather than 1 or 0? And I think before it was '1'/'0' as a string (need to confirm this information)

And a required value can be false but not empty or null in many languages and logic.

Update:
on Ark egg for exemple this condition if we remove required can't work anymore
( [  \"$SERVER_PVE\" == \"0\" ] || printf %s '?ServerPVE=True' )

<img width="404" height="197" alt="Image" src="https://github.com/user-attachments/assets/cc81058f-6ddd-4683-b438-e706029f63de" />

Boy132:
[...]

PR #1830, opened 2025-10-25
Fix boolean validator on startup value

rurueuh:
fixing https://github.com/pelican-dev/panel/issues/1829


rurueuh:

Boy132:
Thanks but forcing a boolean to always be `0` or `1` will break some eggs when they expect `false` and `true`.
See my comment under the issue: https://github.com/pelican-dev/panel/issues/1829#issuecomment-3446894761

PR #1837, opened 2025-10-27
Use `Policies` rather then overriding `can*()` functions

### LINKED ISSUE #1831: Move to Policies
> Overriding the `can*()` authorization methods on a `Resource`, `RelationManager` or `ManageRelatedRecords` class</span>
> 
> Although these methods, such as `canCreate()`, `canViewAny()` and `canDelete()`weren’t documented, if you’re overriding those to provide custom authorization logic in v3, you should be aware that they aren’t always called in v4. The authorization logic has been improved to properly support [policy response objects](https://laravel.com/docs/authorization#policy-responses), and these methods were too simple as they are just able to return booleans.
> 
> If you can make the authorization customization inside the policy of the model instead, you should do that. If you need to customize the authorization logic in the resource or relation manager class, you should override the `get*AuthorizationResponse()` methods instead, such as `getCreateAuthorizationResponse()`, `getViewAnyAuthorizationResponse()` and `getDeleteAuthorizationResponse()`. These methods are called when the authorization logic is executed, and they return a [policy response object](https://laravel.com/docs/authorization#policy-responses). If you remove the override for the `can*()` methods, the `get*AuthorizationResponse()` methods will be used to determine the authorization response boolean, so you don't have to maintain the logic twice.

[Source](https://filamentphp.com/docs/4.x/upgrade-guide#low-impact-changes)

rmartinoscar:
Closes #1831

Made this a Draft cause Filament doesn't seem to call them everywhere atm.

### DIFF
--- a/app/Policies/Admin/ServerPolicy.php
+++ b/app/Policies/Admin/ServerPolicy.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace App\Policies\Admin;
+
+class ServerPolicy
+{
+    use DefaultPolicies;
+
+    protected string $modelName = 'server';
+}
--- a/app/Policies/Server/ServerPolicy.php
+++ b/app/Policies/Server/ServerPolicy.php
@@ -9,18 +9,6 @@ use App\Models\User;
 class ServerPolicy
 {
     /**
-     * This is a horrendous hack to avoid Laravel's "smart" behavior that does
-     * not call the before() function if there isn't a function matching the
-     * policy permission.
-     *
-     * @param  array<string, mixed>  $arguments
-     */
-    public function __call(string $name, array $arguments): void
-    {
-        // do nothing
-    }
-
-    /**
      * Runs before any of the functions are called. Used to determine if the (sub-)user has permissions.
      */
     public function before(User $user, string $ability, string|Server $server): ?bool
@@ -50,5 +38,17 @@ class ServerPolicy
 
         // Return null to let default policies take over
         return null;
+    }
+
+    /**
+     * This is a horrendous hack to avoid Laravel's "smart" behavior that does
+     * not call the before() function if there isn't a function matching the
+     * policy permission.
+     *
+     * @param  array<string, mixed>  $arguments
+     */
+    public function __call(string $name, array $arguments): void
+    {
+        // do nothing
     }
 }


app/Policies/Server/ActivityLogPolicy.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class ActivityLogPolicy
+{
+    protected string $modelName = 'activityLog';
+
+    public function viewAny(): bool
+    {
+        return user()?->can(Permission::ACTION_ACTIVITY_READ, Filament::getTenant());
+    }
```
Boy132: The user should be passed as param, like the other (admin) policies.
This is not always the logged-in user.

https://github.com/pelican-dev/panel/blob/d1a808a746e531cc5531b22e248c6aee9cfe476c/app/Policies/DefaultPolicies.php#L10-L13

app/Policies/Server/ActivityLogPolicy.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace App\Policies\Server;
+
+use App\Models\Permission;
+use Filament\Facades\Filament;
+use Illuminate\Database\Eloquent\Model;
+
+class ActivityLogPolicy
+{
+    protected string $modelName = 'activityLog';
```
Boy132: I don't think the modelName is needed?

app/Policies/Server/DefaultPolicies.php
```diff
@@ -0,0 +1,18 @@
...
+     * This is a horrendous hack to avoid Laravel's "smart" behavior that does
+     * not call the before() function if there isn't a function matching the
+     * policy permission.
+     *
+     * @param  array<string, mixed>  $arguments
+     */
+    public function __call(string $name, array $arguments): void
+    {
+        // do nothing
+    }
```
Boy132: Why was this moved to it's own class/ trait?

app/Providers/AppServiceProvider.php
```diff
@@ -169,8 +170,21 @@ public function boot(
...
+            if ($panelId === 'App') {
+                return;
+            }
+
+            $modelName = class_basename($modelClass);
+            $class = "App\\Policies\\{$panelId}\\{$modelName}Policy";
+
+            if (class_exists($class)) {
+                return $class;
+            }
```
Boy132, line 124: Isn't this missing a `return null;` or something at the end?

Boy132:
@coderabbitai review

app/Filament/Server/Resources/Schedules/ScheduleResource.php
```diff
@@ -357,7 +334,8 @@ public static function defaultTable(Table $table): Table
                     ->state(fn (Schedule $schedule) => $schedule->status === ScheduleStatus::Active ? $schedule->next_run_at : null),
             ])
             ->recordActions([
-                ViewAction::make(),
+                ViewAction::make()
+                    ->hidden(fn ($record) => static::getEditAuthorizationResponse($record)->allowed()),
```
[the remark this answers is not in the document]
rmartinoscar: This is expected we don't want to show the read only ViewAction if the user has Edit privilege.

Boy132 (approved):

PR #1841, opened 2025-10-27
Fix 500 "No route found" when creating db host

m41denx:
This fixes existing issue with application api when creating Database Host.

Pelican API was throwing error 500 "No route found" (yes, 500 and no route), yet still creating DB host entry.


m41denx:

PR #1842, opened 2025-10-27
Fix weird postgres behavior when selecting mounts

m41denx:
Basically I use Pelican with Postgres (I know it's not recommended, but still) and on New Mount page in admin area I get following error:
```
SQLSTATE[42883]: Undefined function: 7 ERROR: could not identify an equality operator for type json 
LINE 1: select distinct "eggs".* from "eggs" left join "mountables" ... ^ 
(Connection: pgsql, SQL: select distinct "eggs".* from "eggs" left join "mountables" on "eggs"."id" = "mountables"."mountable_id" order by "eggs"."name" asc limit 50) 
(View: /var/www/html/vendor/filament/forms/resources/views/components/select.blade.php) 
(View: /var/www/html/vendor/filament/forms/resources/views/components/select.blade.php) 
(View: /var/www/html/vendor/filament/forms/resources/views/components/select.blade.php) 
(View: /var/www/html/vendor/filament/forms/resources/views/components/select.blade.php) 
(View: /var/www/html/vendor/filament/forms/resources/views/components/select.blade.php)
```

I'm not sure if this the correct solution, but I opted for selecting only distinct id and name to prevent Postgres from comparing JSON columns

### DIFF
--- a/app/Filament/Admin/Resources/Mounts/MountResource.php
+++ b/app/Filament/Admin/Resources/Mounts/MountResource.php
@@ -162,12 +162,8 @@ class MountResource extends Resource
                     Section::make()->schema([
                         Select::make('eggs')->multiple()
                             ->label(trans('admin/mount.eggs'))
-                            ->relationship('eggs', 'name', function (Builder $query) {
-                                // Selecting only non-json fields to prevent Postgres from choking on DISTINCT JSON columns
-                                return $query
-                                    ->select(['eggs.id', 'eggs.name'])
-                                    ->orderBy('eggs.name');
-                            })
+                            // Selecting only non-json fields to prevent Postgres from choking on DISTINCT JSON columns
+                            ->relationship('eggs', 'name', fn (Builder $query) => $query->select(['eggs.id', 'eggs.name']))
                             ->preload(),
                         Select::make('nodes')->multiple()
                             ->label(trans('admin/mount.nodes'))


m41denx:
I've just deployed this fix to my server and now I can create mounts, BUT NOT EDIT THEM
image

I'll investigate this issue further

m41denx:
Yes, that fixed the issue. I couldn't find any more issues with postgres and mounts, all CRUD operations seem to work

app/Filament/Admin/Resources/Mounts/MountResource.php
```diff
@@ -162,7 +162,12 @@ public static function defaultForm(Schema $schema): Schema
                     Section::make()->schema([
                         Select::make('eggs')->multiple()
                             ->label(trans('admin/mount.eggs'))
-                            ->relationship('eggs', 'name')
+                            ->relationship('eggs', 'name', function (Builder $query) {
+                                // Selecting only non-json fields to prevent Postgres from choking on DISTINCT JSON columns
+                                return $query
+                                    ->select(['eggs.id', 'eggs.name'])
+                                    ->orderBy('eggs.name');
+                            })
```
Boy132: I don't think we need to sort by name?
```suggestion
                            // Selecting only non-json fields to prevent Postgres from choking on DISTINCT JSON columns
                            ->relationship('eggs', 'name', fn (Builder $query) => $query->select(['eggs.id', 'eggs.name']))
```

_Not pinted_

m41denx: Touché

rmartinoscar (approved):

PR #1850, opened 2025-10-30
feat: Add mixed navigation type with admin-configurable defaults

PalmarHealer:
Adds a new "mixed" navigation type that combines sidebar navigation with topbar functionality (like it was before #1789), and allows admins to set the default navigation type for new users.

## Changes
- **Navigation Types**: Extends navigation options from boolean (sidebar/topbar) to three distinct types:
  - `sidebar`: Traditional sidebar navigation (default)
  - `topbar`: Full top navigation layout
  - `mixed`: Sidebar navigation with topbar enabled (best of both worlds)

- **Admin Configuration**:
  - Added `FILAMENT_DEFAULT_NAVIGATION` setting in admin settings panel
  - New users will inherit the admin-configured default navigation type
  - Setting stored in `config/panel.php` as `filament.default-navigation`

- **User Customization**:
  - Updated user profile customization to include all three navigation options
  - Added backward compatibility for existing boolean values (auto-converts to string types)
  - **Panel Providers**:
  - Updated `PanelProvider.php` to handle string-based navigation types
  - Updated `AppPanelProvider.php` to support mixed mode
  - Maintains backward compatibility with boolean values (`true` → `topbar`, `false` → `sidebar`)

### DIFF
--- a/app/Providers/Filament/AdminPanelProvider.php
+++ b/app/Providers/Filament/AdminPanelProvider.php
@@ -17,11 +17,6 @@ class AdminPanelProvider extends PanelProvider
             ->path('admin')
             ->homeUrl('/')
             ->breadcrumbs(false)
-            ->topbar(function () {
-                $navigationType = user()?->getCustomization(\App\Enums\CustomizationKey::TopNavigation);
-
-                return $navigationType === 'mixed' || $navigationType === 'topbar' || $navigationType === true;
-            })
             ->sidebarCollapsibleOnDesktop(fn () => !$panel->hasTopNavigation())
             ->userMenuItems([
                 Action::make('exit_admin')
--- a/app/Providers/Filament/ServerPanelProvider.php
+++ b/app/Providers/Filament/ServerPanelProvider.php
@@ -20,11 +20,6 @@ class ServerPanelProvider extends PanelProvider
             ->path('server')
             ->homeUrl(fn () => Filament::getPanel('app')->getUrl())
             ->tenant(Server::class, 'uuid_short')
-            ->topbar(function () {
-                $navigationType = user()?->getCustomization(\App\Enums\CustomizationKey::TopNavigation);
-
-                return $navigationType === 'mixed' || $navigationType === 'topbar' || $navigationType === true;
-            })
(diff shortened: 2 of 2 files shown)


PalmarHealer:
Does this add more complexity or should I implement it as suggested?

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -17,6 +17,11 @@ public function panel(Panel $panel): Panel
             ->path('admin')
             ->homeUrl('/')
             ->breadcrumbs(false)
+            ->topbar(function () {
+                $navigationType = user()?->getCustomization(\App\Enums\CustomizationKey::TopNavigation);
+
+                return $navigationType === 'mixed' || $navigationType === 'topbar' || $navigationType === true;
+            })
```
Boy132: Can be removed, the admin panel extends our `PanelProvider` which already has the `topNavigation` and `topbar` settings.

app/Providers/Filament/ServerPanelProvider.php
```diff
@@ -20,6 +20,11 @@ public function panel(Panel $panel): Panel
             ->path('server')
             ->homeUrl(fn () => Filament::getPanel('app')->getUrl())
             ->tenant(Server::class, 'uuid_short')
+            ->topbar(function () {
+                $navigationType = user()?->getCustomization(\App\Enums\CustomizationKey::TopNavigation);
+
+                return $navigationType === 'mixed' || $navigationType === 'topbar' || $navigationType === true;
+            })
```
Boy132: Can be removed, the server panel extends our `PanelProvider` which already has the `topNavigation` and `topbar` settings.

Boy132 (changes requested):

app/Providers/Filament/AdminPanelProvider.php
PalmarHealer: didn't will change

app/Providers/Filament/ServerPanelProvider.php
PalmarHealer: didn't will change

PR #1851, opened 2025-10-30
add server description to grid view too

mristau:
table view already has description, just adding to the grid view too
+ added .ddev to gitignore to keep it from adding all development settings

### DIFF
--- a/resources/views/livewire/server-entry.blade.php
+++ b/resources/views/livewire/server-entry.blade.php
@@ -22,7 +22,6 @@
         "></div>
         @endif
 
-        <div class="flex items-center mb-5 gap-2">
         <div @class([
             'flex items-center gap-2',
             'mb-5' => !$server->description,


mristau:

PalmarHealer:
Good idea. I was always wondering where that would be used for because I can't call any place to name where it is being displayed except in the settings...

If you add it you could also add it to the admin panel server list in https://github.com/pelican-dev/panel/blob/main/app/Filament/Admin/Resources/Servers/Pages/ListServers.php line 48
```php
->description(fn (Server $server) => $server->description)
```

resources/views/livewire/server-entry.blade.php
```diff
@@ -11,7 +11,7 @@ class="relative cursor-pointer"
     </div>
 
     <div class="flex-1 dark:bg-gray-800 dark:text-white rounded-lg overflow-hidden p-3">
-        <div class="flex items-center mb-5 gap-2">
+        <div class="flex items-center gap-2{{$server->description ? '' : ' mb-5' }}">
```
rmartinoscar:
```suggestion
        <div @class([
            'flex items-center gap-2',
            'mb-5' => !$server->description,
        ])>
```
Please use a @class Attribute merge instead of raw ternary

rmartinoscar (changes requested):

Boy132:
I like the general idea, but it looks a bit weird:
grafik
grafik

mristau:
image
it is already there for table and it even creates a scrollbar with such long descriptions, it could get capped to like 40 characters though, but should then be for both

rmartinoscar (approved):

Boy132:
> image it is already there for table and it even creates a scrollbar with such long descriptions, it could get capped to like 40 characters though, but should then be for both

I didn't mean the length but the general position. IMO it looks out of place and squeezed in.

mristau:
i had moved it in, to be inline with the title, but there are not enough margin-classes for the whole icon, so had to add padding
image

PR #1852, opened 2025-10-30
fix: Automatic allocation creation

PalmarHealer:
Fixes that new allocations are being created because ```$server->node->allocations()``` is scoped. Also adds the option to disable the creation for new allocations and limits it to the unassigned allocations already available.


JoanFo1456:
I don't think this is a good idea.
What if you don't want to give X allocations?
It is random and good already, if you need X allocation ask the admin.
> Also this looks very AI to me, comments make it look like it, since they aren't like the ones they add

PalmarHealer:
> I don't think this is a good idea. What if you don't want to give X allocations? It is random and good already, if you need X allocation ask the admin.

I think there's a bit of a misunderstanding here.
This doesn't change the default behaviour - it's an optional mode admins can enable. The problem I'm solving: I have 2 nodes with different available port ranges. Node A might have ports 25565-25600 forwarded, and Node B has 30000-30050. The current system only lets you configure one global range, which doesn't work for this. With "random from existing" mode, I can manually add the specific available ports for each node, and the system picks from those instead of trying to create new ones from a range that might not work on that node.

This is just giving admins more control for setups where not all ports are available or where different nodes have different port configurations. The default "range" mode stays exactly the same. I can't see why this would be a bad idea

PalmarHealer:
> > Also this looks very AI to me, comments make it look like it, since they aren't like the ones they add

I used AI as a tool to save time, but I read over everything myself. My goal was to solve this problem, not creating some AI crap it thought would add to the repo.

JoanFo1456:
Not gonna say anything else, but I think it's a bit stupid the idea of having this of only giving the allocations you made or the range one. You could've made a setting for the node itself for the range instead.
> Making what you want, different ranges per each node, and it would make things a lot easier.

PalmarHealer:
> Not gonna say anything else, but I think it's a bit stupid the idea of having this of only giving the allocations you made or the range one. You could've made a setting for the node itself for the range instead.
> 
> > Making what you want, different ranges per each node, and it would make things a lot easier.

I'm sorry but I can't see your argument. This is an option you can simply not use and this would be great for me. I know a per-node basis could work as well but comes with drawbacks. The current implementation would allow only assigning every other port (if you would want that) and a per-node basis with a start and end port wouldn't. You could implement a whole separate list for each node to enable that functionality but it feels a lot like shooting with cannons at sparrows.

JoanFo1456:
> > Not gonna say anything else, but I think it's a bit stupid the idea of having this of only giving the allocations you made or the range one. You could've made a setting for the node itself for the range instead.
> > > Making what you want, different ranges per each node, and it would make things a lot easier.
> 
> I'm sorry but I can't see your argument. This is an option you can simply not use and this would be great for me. I know a per-node basis could work as well but comes with drawbacks. The current implementation would allow only assigning every other port (if you would want that) and a per-node basis with a start and end port wouldn't. You could implement a whole separate list for each node to enable that functionality but it feels a lot like shooting with cannons at sparrows.

What I can see the code does is to if it's on random when not finding any allocation creates a new one, otherwise it will say that there aren't any allocations.
The problem you had is the range, two nodes different ranges, instead of locking to only existing allocations, you can create the range on each node, and when calling factory to check the node range.

PalmarHealer:
I don't understand what you're trying to say. I'm sorry. But thank you for testing me.

There was a bug in the original code that ```$server->node->allocations()``` only returns already assigned allocations. The query is:
```select * from `allocations` where `node_id` = ? and `server_id` is null and `allocations`.`server_id` in (?)``` and ```and `allocations`.`server_id` in (?)``` makes it impossible to find any unassigned allocations making the function create new allocations regardless if there where unused allocations or not. I've also changed the visual presentation so its a bit easier to understand what it does.

grafik

**Comment by JoanFo1456** [comment]
> I don't understand what you're trying to say. I'm sorry. But thank you for testing me.
> 
> There was a bug in the original code that `$server->node->allocations()` only returns already assigned allocations. The query is: `` select * from `allocations` where `node_id` = ? and `server_id` is null and `allocations`.`server_id` in (?) `` and `` and `allocations`.`server_id` in (?) `` makes it impossible to find any unassigned allocations making the function create new allocations regardless if there where unused allocations or not. I've also changed the visual presentation so its a bit easier to understand what it does.
> 


image
Something like this, to make the node say which port range will be used.
Node1 can have 25565-35565 and node2 4453-10000 and will check this to do the allocations.

> This was made fast, the idea is to add two TextInputs, like the old Port range

**Comment by PalmarHealer** [comment]
> Something like this, to make the node say which port range will be used. Node1 can have 25565-35565 and node2 4453-10000 and will check this to do the allocations. 
> 
> > This was made fast, the idea is to add two TextInputs, like the old Port range

I think you're only arguing with me just for the sake of it.

I feel like your talking me down just because you don’t like my Idea. At the start you said this is a "bit stupid" and now you suggest to make it more advanced.

What do you want?

**Comment by JoanFo1456** [comment]
> I think you're only arguing with me just for the sake of it.
> 
> I feel like your talking me down just because you don’t like my Idea. At the start you said this is a "bit stupid" and now you suggest to make it more advanced.
> 
> What do you want?

At first, the idea of making the edit was dumb the way it was first made. Edit: I mean like the random allowing creation or not.
Once I understood the problem, it's better to fix it.
This suggestion I made is to fix the problem you have requiring different range ports on a Node, without overcomplicating.

**Comment by PalmarHealer** [comment]
The per-node solution is more complicated and the assignments where random before, this was there before my change.

**Comment by JoanFo1456** [comment]
> The per-node solution is more complicated and the assignments where random before, this was there before my change.

> If you don't mind, speak with me on discord, so it's not flooding this… My username is .joanfo (you can also mention me on general)

Edit: Nevermind, do it whatever you want, as you seem to know what you want

**Comment by Boy132** [comment]
Why would you want to disable the allocation creation if no allocation is available?

**Comment by PalmarHealer** [comment]
Because I have 2 nodes and they have different ranges of what ports can be forwarded and the second one is behind a firewall. 

I have to request every port so I forwarded "only" 1000 and these are the only ports that should be used. And I could set the range to the 1000 ports of the second node but the first node has a different range.

I know that a per-node basis would work for me as well but that is way easier and does the job just fine. That way I don't have to keep track of the allocation ranges. I just create all the ports that can be used on that node and done.

File: app/Filament/Admin/Pages/Settings.php [hunk 1]
```diff
@@ -613,21 +613,33 @@ private function miscSettings(): array
...
+                        ->stateCast(new BooleanStateCast(false))
+                        ->default(env('PANEL_CLIENT_ALLOCATIONS_CREATE_NEW', config('panel.client_features.allocations.create_new'))),
                     TextInput::make('PANEL_CLIENT_ALLOCATIONS_RANGE_START')
                         ->label(trans('admin/setting.misc.auto_allocation.start'))
                         ->required()
                         ->numeric()
                         ->minValue(1024)
                         ->maxValue(65535)
-                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED'))
+                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED') && $get('PANEL_CLIENT_ALLOCATIONS_CREATE_NEW'))
```
Boy132, line 643: The start and end range needs to be visible even when `PANEL_CLIENT_ALLOCATIONS_CREATE_NEW` is false.
```suggestion
                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED'))
```

app/Filament/Admin/Pages/Settings.php
```diff
@@ -613,21 +613,33 @@ private function miscSettings(): array
...
                         ->default(env('PANEL_CLIENT_ALLOCATIONS_RANGE_START')),
                     TextInput::make('PANEL_CLIENT_ALLOCATIONS_RANGE_END')
                         ->label(trans('admin/setting.misc.auto_allocation.end'))
                         ->required()
                         ->numeric()
                         ->minValue(1024)
                         ->maxValue(65535)
-                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED'))
+                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED') && $get('PANEL_CLIENT_ALLOCATIONS_CREATE_NEW'))
```
Boy132, line 651: The start and end range needs to be visible even when `PANEL_CLIENT_ALLOCATIONS_CREATE_NEW` is false.
```suggestion
                        ->visible(fn (Get $get) => $get('PANEL_CLIENT_ALLOCATIONS_ENABLED'))
```

app/Services/Allocations/FindAssignableAllocationService.php
```diff
@@ -37,18 +39,31 @@ public function handle(Server $server): Allocation
...
         /** @var Allocation|null $allocation */
-        $allocation = $server->node->allocations()
+        $allocation = Allocation::withoutGlobalScopes()
+            ->where('node_id', $server->node_id)
             ->when($server->allocation, function ($query) use ($server) {
                 $query->where('ip', $server->allocation->ip);
             })
             ->whereNull('server_id')
             ->inRandomOrder()
             ->first();
```
Boy132, line 59: This should also check for the start and end range, really.
I'm honestly suprised it uses a random one and completely ignores the range.

PalmarHealer: The range is only taken into account when no free allocation is found (and since `$server->node->allocations()` is scoped it is currently always creating a new one)

I think this not bad how it is now. That it will look for any unassigned allocation regardless of the given range and create a new allocation when no is available and only then use the range. Thats why i made the range inputs inisible because I thought this was intentional. (I also explained the behaviour in the translations so there is no confusion)

Can you give me a thumbs up on this if I should change it to your suggestion?

Boy132: I think it's very confusing if you provide a range and then it will just ignore it and choose a random one. Why should it only apply the range when creating a new allocation and not when selecting?

mrhid6:
When `$server->allocation` is null, e.g., when there are no allocations on the server, creating a new allocation or associating an allocation fails because it cannot obtain the IP.

I have been creating my servers using the API and leaving the network allocation null, so it can be added in the console.
This was working correctly in 1.0.0-beta25 pre Filament 4

Boy132:
I would say we split this PR in 3:
1. Fix the scoping
2. Fix the range not being used for creation
3. Add new toggle to disable automatic creation

Thoughts? @pelican-dev/team

lancepioch:
> I would say we split this PR in 3:
> 
> 1. Fix the scoping
> 2. Fix the range not being used for creation
> 3. Add new toggle to disable automatic creation
> 
> Thoughts? @pelican-dev/team

I think that's a good and fair ask.

PalmarHealer:
Superseded by referenced PRs

PR #1860, opened 2025-11-04
Allow to "embed" server list

Boy132:
Isn't really doing anything on it's own, but allows plugins to "embed" the server list as normal page when expanding the `app` panel.

### DIFF
--- a/app/Filament/App/Resources/Servers/ServerResource.php
+++ b/app/Filament/App/Resources/Servers/ServerResource.php
@@ -33,7 +33,7 @@ class ServerResource extends Resource
         ];
     }
 
-    public static function embedServerList(bool $condition = true)
+    public static function embedServerList(bool $condition = true): void
     {
         static::$slug = $condition ? null : '/';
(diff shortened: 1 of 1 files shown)


app/Filament/App/Resources/Servers/ServerResource.php
```diff
@@ -10,10 +10,17 @@ class ServerResource extends Resource
...
 
+    protected static string|\BackedEnum|null $navigationIcon = 'tabler-brand-docker';
+
     protected static ?string $slug = '/';
 
     protected static bool $shouldRegisterNavigation = false;
 
+    public static function getNavigationBadge(): ?string
+    {
+        return (string) user()?->directAccessibleServers()->where('owner_id', user()?->id)->count();
```
rmartinoscar, line 21: Shouldn't we count every servers accessible by the user (including subs & not owned) ?

Boy132: Since the default tab is "my servers" I felt like it makes more sense when only counting owned servers.

PR #1869, opened 2025-11-07
Add session key to use last used node, instead of latest created node

### LINKED ISSUE #1234: When creating a new Server, try to use the last used Node instead
When creating a new Server, the newest Node is always selected.

Let's try storing the last used Node in the session and default to the newest node if not available!
What is the last used Node? If you create/edit a Node, save that as the last used. If a new Server is created, store its Node as the last used. Etc.

https://github.com/pelican-dev/panel/blob/d721af6be10ab9169b8c36363c46df2a21b2b0ea/app/Filament/Admin/Resources/ServerResource/Pages/CreateServer.php#L112

notAreYouScared:
Closes #1234 
-
User experience improvements:

* The form now attempts to pre-select the last used node by retrieving it from the session, provided the user still has access to that node. If not, it falls back to the latest accessible node.
* After a server is created, the selected node is saved to the session as `lastUsedNode`, ensuring future server creations default to the user's previous choice.

### DIFF
--- a/app/Filament/Admin/Resources/Servers/Pages/CreateServer.php
+++ b/app/Filament/Admin/Resources/Servers/Pages/CreateServer.php
@@ -116,7 +116,7 @@ class CreateServer extends CreateRecord
                                 ->prefixIcon('tabler-server-2')
                                 ->selectablePlaceholder(false)
                                 ->default(function () {
-                                    $lastUsedNode = session()->get('lastUsedNode');
+                                    $lastUsedNode = session()->get('last_utilized_node');
 
                                     if ($lastUsedNode && user()?->accessibleNodes()->where('id', $lastUsedNode)->exists()) {
                                         $this->node = Node::find($lastUsedNode);


app/Filament/Admin/Resources/Servers/Pages/CreateServer.php
```diff
@@ -116,6 +116,14 @@ public function form(Schema $schema): Schema
                                 ->prefixIcon('tabler-server-2')
                                 ->selectablePlaceholder(false)
                                 ->default(function () {
+                                    $lastUsedNode = session()->get('lastUsedNode');
```
lancepioch:
```suggestion
                                    $lastUsedNode = session()->get('last_utilized_node');
```

app/Filament/Admin/Resources/Servers/Pages/CreateServer.php
```diff
@@ -829,6 +837,8 @@ protected function handleRecordCreation(array $data): Model
             $data['allocation_additional'] = collect($allocation_additional)->filter()->all();
         }
 
+        session()->put('lastUsedNode', $data['node_id']);
```
lancepioch:
```suggestion
        session()->put('last_utilized_node', $data['node_id']);
```

PR #1870, opened 2025-11-07
Fix Docker configuration issues

gOOvER:
- Add flexible webserver configuration (SKIP_CADDY, SKIP_WEBSERVER)
- Harmonize upload limits between PHP and Caddy (UPLOAD_LIMIT)
- Improve entrypoint.sh with better webserver mode handling
- Update Caddyfile with proper upload limits and timeouts

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -22,10 +22,21 @@ else
   echo -e "APP_INSTALLED=false" >> /pelican-data/.env
 fi
 
-sed -i "s/^upload_max_filesize = .*/upload_max_filesize = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
-sed -i "s/^post_max_size = .*/post_max_size = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
-sed -i "s/^max_execution_time = .*/max_execution_time = 300/" /usr/local/etc/php/php.ini-production
-sed -i "s/^memory_limit = .*/memory_limit = 512M/" /usr/local/etc/php/php.ini-production
+## Configure PHP settings
+UPLOAD_LIMIT_VAL=${UPLOAD_LIMIT:-100}
+
+# Calculate max_execution_time based on upload limit (larger files need more time)
+# Base time: 300s, additional 3s per MB above 100MB
+if [ "$UPLOAD_LIMIT_VAL" -gt 100 ]; then
+  CALCULATED_EXECUTION_TIME=$((300 + (UPLOAD_LIMIT_VAL - 100) * 3))
+else
+  CALCULATED_EXECUTION_TIME=300
+fi
+
+sed -i "s/^upload_max_filesize = .*/upload_max_filesize = ${UPLOAD_LIMIT_VAL}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/^post_max_size = .*/post_max_size = ${UPLOAD_LIMIT_VAL}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/^max_execution_time = .*/max_execution_time = ${PHP_MAX_EXECUTION_TIME:-$CALCULATED_EXECUTION_TIME}/" /usr/local/etc/php/php.ini-production
+sed -i "s/^memory_limit = .*/memory_limit = ${PHP_MEMORY_LIMIT:-512M}/" /usr/local/etc/php/php.ini-production
 mkdir -p /pelican-data/database /pelican-data/storage/avatars /pelican-data/storage/fonts /var/www/html/storage/logs/supervisord 2>/dev/null
 
 if ! grep -q "APP_KEY=" .env || grep -q "APP_KEY=$" .env; then


docker/entrypoint.sh
```diff
@@ -22,7 +22,10 @@ else
   echo -e "APP_INSTALLED=false" >> /pelican-data/.env
 fi
 
-sed -i "s/upload_max_filesize = 2M/upload_max_filesize = ${UPLOAD_LIMIT}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/upload_max_filesize = 2M/upload_max_filesize = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
```
rmartinoscar:
```suggestion
sed -Ei "s/(upload_max_filesize|post_max_size) = [0-9]+M/\1 = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
```

docker/entrypoint.sh
```diff
@@ -40,20 +43,27 @@ php artisan migrate --force
...
+  # Only PHP-FPM will run, no internal webserver
+elif [ "${SKIP_CADDY:-}" = "true" ]; then
+  echo "Starting PHP-FPM only (Caddy disabled)"
+  # Only PHP-FPM will run, but for internal use without webserver
 else
-  echo "Starting PHP-FPM and Caddy"
+  echo "Starting PHP-FPM and Caddy webserver"
   # enable caddy
   export SUPERVISORD_CADDY=true
+  export SUPERVISORD_WEBSERVER=true
```
rmartinoscar: Im really confused as to what SUPERVISORD_WEBSERVER is supposed to do since you didn't edit [supervisor.conf](https://github.com/pelican-dev/panel/blob/main/docker/supervisord.conf)

docker/entrypoint.sh
```diff
@@ -22,7 +22,10 @@ else
   echo -e "APP_INSTALLED=false" >> /pelican-data/.env
 fi
 
-sed -i "s/upload_max_filesize = 2M/upload_max_filesize = ${UPLOAD_LIMIT}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/^upload_max_filesize = .*/upload_max_filesize = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/^post_max_size = .*/post_max_size = ${UPLOAD_LIMIT:-100}M/" /usr/local/etc/php/php.ini-production
+sed -i "s/^max_execution_time = .*/max_execution_time = 300/" /usr/local/etc/php/php.ini-production
+sed -i "s/^memory_limit = .*/memory_limit = 512M/" /usr/local/etc/php/php.ini-production
```
rmartinoscar: These should have their own env variable as well at this point.
Maybe calculate max_execution_time based on $UPLOAD_LIMIT ?

docker/entrypoint.sh
gOOvER: does alpine sed support -E? Not every version of SED supports E

docker/entrypoint.sh
```diff
@@ -43,17 +45,22 @@ php artisan filament:optimize
...
-  echo "Starting PHP-FPM and Caddy"
+  echo "Starting PHP-FPM and Caddy webserver"
   # enable caddy
   export SUPERVISORD_CADDY=true
 
   # handle trusted proxies for caddy
-  if [[ ! -z ${TRUSTED_PROXIES} ]]; then
-    export CADDY_TRUSTED_PROXIES=$(echo "trusted_proxies static ${TRUSTED_PROXIES}" | sed 's/,/ /g')
+  if [ -n "${TRUSTED_PROXIES:-}" ]; then
+    CADDY_TRUSTED_PROXIES_VALUE=$(echo "trusted_proxies static ${TRUSTED_PROXIES}" | sed 's/,/ /g')
```
rmartinoscar, line 73: Why ?

gOOvER: Deepl Translation:

We had to replace  with  to make the script compatible with Alpine Linux. The container uses  as its shell (see ), and  is Bash-specific — it throws syntax errors in  or .
The new syntax is POSIX-compliant and works across all common shells. Using  ensures we don’t hit “unbound variable” errors if the variable isn’t set, and quoting it avoids issues with word splitting (e.g. if the value contains spaces). This also keeps Shellcheck happy and avoids false positives around command substitution in exports.

docker/entrypoint.sh
rmartinoscar: Sure does in our image.
image

notAreYouScared:
Uploads will be client side when the drag and drop pr is merged. So setting the client time and post size seems kinda useless  

The only upload will be egg / egg images / avatar / schedule andddd I think thats it? None of those should be over 2MB

docker/entrypoint.sh
gOOvER: ChatGTP Translation: (looks better than Deepl :D)

This change doesn’t make sense for a couple of reasons:
1. Shell compatibility issue
Alpine Linux uses  and busybox , not GNU . The  combination doesn’t work here — busybox  doesn’t support , and combining it with  throws an error. So this would break in Alpine-based containers.
2. Regex is too narrow
The pattern  only matches values like , but fails for valid alternatives like  or . That makes it fragile and format-dependent.
3. Current solution is more robust

gOOvER:
> Uploads will be client side when the drag and drop pr is merged. So setting the client time and post size seems kinda useless
> 
> The only upload will be egg / egg images / avatar / schedule andddd I think thats it? None of those should be over 2MB

So the pinned message from fish is also outdated. Lets wait fot the full release of pelican and then update the image

PR #1872, opened 2025-11-08
Add avatar option to use app logo

notAreYouScared:


rmartinoscar:
https://github.com/pelican-dev/panel/blob/cec141889a0e77bf2e732ff12a17198ec6d04cb8/app/Extensions/Avatar/AvatarService.php#L28-L39
How about we set the default if it couldn't find users's uploaded avatar ?
*Not worth adding an AvatarSchema for a default imo*

notAreYouScared:
The users uploaded avatar still overrides this, assuming its enabled in the env

lancepioch (approved):
I feel like this is a lot of code, but nothing looks wrong lol

app/Providers/Filament/PanelProvider.php
```diff
@@ -34,6 +35,7 @@ public function panel(Panel $panel): Panel
             ->brandLogo(config('app.logo'))
             ->brandLogoHeight('2rem')
             ->favicon(config('app.favicon', '/pelican.ico'))
+            ->defaultAvatarProvider(PelicanAvatarProvider::class)
```
Boy132, line 38: Overwriting the default provider will break the uiavatars provider because we assume uiavatars is the default.
https://github.com/pelican-dev/panel/blob/a30c45fbbecb44b13d5a3dd2bd01d7cf9766ceed/app/Extensions/Avatar/Schemas/UiAvatarsSchema.php#L20-L24

notAreYouScared: I honestly don't care, its other people that care.

PR #1873, opened 2025-11-08
Replaced UI Avatars

JoanFo1456:
Filament depends on an external service to make server images and user ones, this pr wants to make the panel able to make the same icons but locally, to avoid more problems like not showing an image at all.
(If you see some useless commits that I could do at once, my Git is on some substance)

### DIFF
--- a/app/Extensions/Avatar/Schemas/Local/LocalAvatarProvider.php
+++ b/app/Extensions/Avatar/Schemas/Local/LocalAvatarProvider.php
@@ -2,18 +2,15 @@
 
 namespace App\Extensions\Avatar\Schemas\Local;
 
-use App\Models\User;
 use Filament\AvatarProviders\Contracts\AvatarProvider;
 use Filament\Facades\Filament;
-use Filament\Models\Contracts\HasAvatar;
-use Filament\Models\Contracts\HasName;
 use Illuminate\Database\Eloquent\Model;
 
 class LocalAvatarProvider implements AvatarProvider
-{   
+{
     private string $name;
 
-    private ?string $initials;
+    private string $initials;
 
     private string $backgroundColor = '111827';
 
@@ -21,17 +18,16 @@ class LocalAvatarProvider implements AvatarProvider
 
     private float $size = 128;
 
-    private float $fontSize = 128;
-
     private float $fontSizeMultiplier = 0.4;
 
     public function get(Model $record): string
     {
         $name = Filament::getNameForDefaultAvatar($record);
-        
+
         return $this->generateDataUri($name);
-        
+
     }
+
     private function generateDataUri(string $name): string
     {
         $this->name = $name;
@@ -49,7 +45,8 @@ class LocalAvatarProvider implements AvatarProvider
         });
     }
 
-    public function generateSvgAvatar(): string {
+    public function generateSvgAvatar(): string
+    {
         $fontSize = $this->size * $this->fontSizeMultiplier;
 
         $svg = <<<SVG
@@ -77,7 +74,6 @@ SVG;
     {
         $hash = md5($this->initials);
         $hue = hexdec(substr($hash, 0, 2)) / 255 * 360;
-
 
         return $this->hslToHex($hue);
     }
--- a/app/Extensions/Avatar/Schemas/Local/LocalAvatarSchema.php
+++ b/app/Extensions/Avatar/Schemas/Local/LocalAvatarSchema.php
@@ -4,7 +4,6 @@ namespace App\Extensions\Avatar\Schemas\Local;
 
 use App\Extensions\Avatar\AvatarSchemaInterface;
 use App\Models\User;
-use Filament\Facades\Filament;
 
 class LocalAvatarSchema implements AvatarSchemaInterface
 {
@@ -21,6 +20,7 @@ class LocalAvatarSchema implements AvatarSchemaInterface
     public function get(User $user): string
     {
         $provider = new LocalAvatarProvider();
+
         return $provider->get($user);
     }
 }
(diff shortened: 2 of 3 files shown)


JoanFo1456:
He's right, I'll fix it after finishing my homework, thanks Rabbit.

app/Providers/Extensions/AvatarServiceProvider.php
```diff
@@ -16,7 +16,7 @@ public function register(): void
 
             // Default Avatar providers
             $service->register(new GravatarSchema());
-            $service->register(new UiAvatarsSchema());
+            $service->register($app->make(LocalAvatarSchema::class));
```
rmartinoscar:
```suggestion
            $service->register(new LocalAvatarSchema());
```

app/Services/Avatars/LocalAvatarService.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $hash = md5($name);
+        $hue = hexdec(substr($hash, 0, 2)) / 255 * 360;
+        $saturation = 60;
+        $lightness = 45;
+
+        return $this->hslToHex($hue, $saturation, $lightness);
+    }
+
+    protected function hslToHex(float $h, float $s, float $l): string
```
rmartinoscar: Please use descriptive variable names.

JoanFo1456: Alright, I'll change it in a sec. Thanks

app/Providers/Filament/PanelProvider.php
```diff
@@ -50,6 +51,7 @@ public function panel(Panel $panel): Panel
                 'profile' => fn (Action $action) => $action
                     ->url(fn () => EditProfile::getUrl(panel: 'app')),
             ])
+            ->defaultAvatarProvider(LocalAvatarProvider::class)
```
rmartinoscar, line 54: https://github.com/pelican-dev/panel/pull/1872/files#r2507327188
Overwriting the default provider will break the uiavatars provider because we assume uiavatars is the default.
https://github.com/pelican-dev/panel/blob/a30c45fbbecb44b13d5a3dd2bd01d7cf9766ceed/app/Extensions/Avatar/Schemas/UiAvatarsSchema.php#L20-L24

app/Services/Avatars/LocalAvatarService.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public function generateDataUri(
+        string $name,
+        string $backgroundColor = '111827',
+        string $textColor = 'FFFFFF',
+        int $size = 128
+    ): string {
+        $backgroundColor = ltrim($backgroundColor, '#');
+        $cacheKey = "avatar:{$name}:{$backgroundColor}:{$textColor}:{$size}";
```
rmartinoscar: We should use $initials rather then $name since different names can lead to the same initials.

app/Services/Avatars/LocalAvatarService.php
```diff
@@ -0,0 +1,105 @@
...
+        return strtoupper((string) $initials);
+    }
+
+    public function generateDataUri(
+        string $name,
+        string $backgroundColor = '111827',
+        string $textColor = 'FFFFFF',
+        int $size = 128
+    ): string {
+        $backgroundColor = ltrim($backgroundColor, '#');
```
rmartinoscar: Why do it for $backgroundColor and not $textColor aswell ?

app/Services/Avatars/LocalAvatarService.php
```diff
@@ -0,0 +1,105 @@
...
+class LocalAvatarService
+{
+    public function generateSvgAvatar(
+        string $name,
+        string $backgroundColor = '111827',
+        string $textColor = 'FFFFFF',
+        int $size = 128
+    ): string {
+        $initials = $this->getInitials($name);
+        $fontSize = $size * 0.4;
```
rmartinoscar: Where does `0.4` come from ?

app/Services/Avatars/LocalAvatarService.php
rmartinoscar:
```suggestion
    public function generateColorFromName(string $name): string
    {
        $hash = md5($name);
        $hue = hexdec(substr($hash, 0, 2)) / 255 * 360;

        return $this->hslToHex($hue);
    }

    protected function hslToHex(float $hue, ?float $saturation = 60, float $lightness = 45): string
```
Please use descriptive variable names.

app/Services/Avatars/LocalAvatarService.php
```diff
@@ -0,0 +1,105 @@
...
+
+            $r = $this->hueToRgb($p, $q, $h + 1 / 3);
+            $g = $this->hueToRgb($p, $q, $h);
+            $b = $this->hueToRgb($p, $q, $h - 1 / 3);
+        }
+
+        return sprintf('%02x%02x%02x', round($r * 255), round($g * 255), round($b * 255));
+    }
+
+    protected function hueToRgb(float $p, float $q, float $t): float
```
rmartinoscar: Same here.

Boy132:
But why? Servers will most likely get proper icons at some point anyways.

JoanFo1456:
> But why? Servers will most likely get proper icons at some point anyways.

So we don't depend on external providers for this, I must say that feels useless, but it's better than the alternative to show a pelican icon.
I'll be pushing changes I made with Martin soon I have 10 minutes for myself.

Boy132:
> > But why? Servers will most likely get proper icons at some point anyways.
> 
> So we don't depend on external providers for this, I must say that feels useless, but it's better than the alternative to show a pelican icon. I'll be pushing changes I made with Martin soon I have 10 minutes for myself.

What's the problem with the external provider? Why should we reinvent the wheel?

JoanFo1456:
> What's the problem with the external provider? Why should we reinvent the wheel?

True, I'm going to close this as it feels useless now that some weeks passed.

PR #1883, opened 2025-11-09
fix: bypass tenant scoping in allocation queries

PalmarHealer:
Use `withoutGlobalScopes()` in allocation queries to ensure unassigned allocations can be found when called from the Server panel context. 

Supersedes #1852


PalmarHealer:
@coderabbitai review

PR #1884, opened 2025-11-09
feat: Add toggle for automatic allocation creation in panel settings

PalmarHealer:
- Introduced `create_new` configuration to control automatic allocation behavior.
- Updated allocation service logic to respect the `create_new` toggle.
- Enhanced panel settings UI to include the new `create_new` toggle option.
- Added helper text for better clarity on configuration options.

Supersedes #1852


PalmarHealer:
@coderabbitai review

rmartinoscar:
Doesn't seem worth adding, if an allocation is created by an admin it will be locked by default already.

PalmarHealer:
But if I provide a large port range and want to exclude certain ports from being assigned, this comes in very handy because I can create the assignable ports beforehand and then use this option.

<details>
  <summary>see more</summary>
  Or, in my use case, I have two nodes and can have different port ranges for each one by creating the ports beforehand and then giving a range that includes all and using this option to prevent ports I don't want from being used.


   I know it's a small change and can be seen as not needed, but I would benefit greatly from this, and I think others can appreciate it as well.
</details>

rmartinoscar:
@PalmarHealer Please resolve conflicts.

PR #1887, opened 2025-11-10
don't try to bulk update if egg doesn't even have a url

mristau:
just checking if update_url is set, before doing bulk update
keeps it from spamming error messages from update_url=null

image

### DIFF
--- a/app/Filament/Components/Actions/UpdateEggBulkAction.php
+++ b/app/Filament/Components/Actions/UpdateEggBulkAction.php
@@ -55,6 +55,7 @@ class UpdateEggBulkAction extends BulkAction
             foreach ($records as $egg) {
                 if ($egg->update_url === null) {
                     $skipped++;
+
                     continue;
                 }
                 try {


notAreYouScared:
I'd prefer the change to be in the actual action.

UpdateEggBulkAction.php

mristau:
did that now, i was thinking, to filter out all records with no url before doing the action

app/Filament/Components/Actions/UpdateEggBulkAction.php
```diff
@@ -49,25 +49,35 @@ protected function setUp(): void
...
+                        cache()->forget("eggs.$egg->uuid.update");
+                    } catch (Exception $exception) {
+                        $failed++;
 
-                    report($exception);
+                        report($exception);
+                    }
+                } else {
+                    $skipped++;
                 }
```
rmartinoscar:
```suggestion
                if ($egg->update_url === null) {
                    $skipped++;
                    continue;
                }
                try {
                    $eggImporterService->fromUrl($egg->update_url, $egg);

                    $success++;

                    cache()->forget("eggs.$egg->uuid.update");
                } catch (Exception $exception) {
                    $failed++;

                    report($exception);
                }
```

mristau: i can't read the full error message from pint now, but the code is like you suggested

PR #1897, opened 2025-11-13
Add back 2fa requirement middleware

### LINKED ISSUE #1718: Forced 2FA Requirement no longer works
### Current Behavior

Currently forcing a 2FA Requirement does nothing, unlike how it previously made the user setup 2FA before letting them use the panel.

### Expected Behavior

The expected behaviour here is some sort of a block preventing users from accessing the panel before having set up 2FA if the requirement is there.

### Steps to Reproduce

1. Enable 2FA Requirement
2. Create a new user account
3. Log in and observe that there is nothing asking said user to enable 2fa

### Panel Version

ce393af7a649189d9ae6f78c4f73e870e4110b34

### Wings Version

N/A

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

Boy132:
Closes #1718

### DIFF
--- a/app/Http/Middleware/RequireTwoFactorAuthentication.php
+++ b/app/Http/Middleware/RequireTwoFactorAuthentication.php
@@ -30,11 +30,8 @@ class RequireTwoFactorAuthentication
         /** @var ?User $user */
         $user = $request->user();
 
-        $uri = rtrim($request->getRequestUri(), '/') . '/';
-        $route = $request->route()->getName();
-
         // Auth and profile endpoints should always be available
-        if (!$user || Str::startsWith($uri, ['/auth/', '/profile']) || Str::startsWith($route, ['auth.', 'account.', 'filament.app.auth.'])) {
+        if (!$user || $request->routeIs('*auth.*')) {
             return $next($request);
         }
 
@@ -52,7 +49,7 @@ class RequireTwoFactorAuthentication
         }
 
         // For API calls return an exception which gets rendered nicely in the API response...
-        if ($request->isJson() || Str::startsWith($uri, '/api/')) {
+        if ($request->isJson() || Str::startsWith($request->path(), '/api')) {
             throw new TwoFactorAuthRequiredException();
         }
 


app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,67 @@
...
+    public function handle(Request $request, \Closure $next): mixed
+    {
+        /** @var ?User $user */
+        $user = $request->user();
+
+        $uri = rtrim($request->getRequestUri(), '/') . '/';
+        $route = $request->route()->getName();
+
+        // Auth and profile endpoints should always be available
+        if (!$user || Str::startsWith($uri, ['/auth/', '/profile']) || Str::startsWith($route, ['auth.', 'account.', 'filament.app.auth.'])) {
```
rmartinoscar: There's no route starting with `account.`.
> Also profile is registered under `filament.*.auth.profile` so there's no need for $uri

`routeIs` seems like a better fit
```suggestion
        // Auth and profile endpoints should always be available
        if (!$user || $request->routeIs('*auth.*')) {
```
Very permitive

```suggestion
        // Auth and profile endpoints should always be available
        if (!$user || $request->routeIs('*auth.*', 'filament.*.auth.*')) {
```
Only matches routes under `auth.*` & `filament.*.auth.*` so if you have `plugin.auth` it won't be granted.

app/Http/Middleware/RequireTwoFactorAuthentication.php
```diff
@@ -0,0 +1,67 @@
...
+            return $next($request);
+        }
+
+        if ($level === self::LEVEL_ADMIN && !$user->isAdmin()) {
+            // If the level is set as admin and the user is not an admin, pass them through as well.
+            return $next($request);
+        }
+
+        // For API calls return an exception which gets rendered nicely in the API response...
+        if ($request->isJson() || Str::startsWith($uri, '/api/')) {
```
rmartinoscar:
```suggestion
        // For API calls return an exception which gets rendered nicely in the API response...
        if ($request->isJson() || Str::startsWith($request->path(), '/api')) {
```
If we use `routeIs()` $uri should only be defined when its of use.

> BTW why do you right trim / but proceed to add it back ?

Boy132: > BTW why do you right trim / but proceed to add it back ?

Because most of the middleware code is from upstream. We only did small changes to make it work with filament v3, then removed it with v4. Now it's coming back. :D

rmartinoscar (approved):

PR #1902, opened 2025-11-15
Remove 'required' rule from egg-garrys-mod.yaml

Luke1505:
Removed 'required' rule from the configuration.

Fixes part of : #1829 as discussed in the Issue if a Boolean is required it expects it to have a value if the toggle is not active the value right now is empty which results in an error. 

Since this is a standard egg this was changed here but for the other eggs mentioned in the issue we need to do that in the Steamcmd-eggs repo


Luke1505:

Luke1505:
recheck

Luke1505:
recheck

PR #1904, opened 2025-11-16
Add server description to loading placeholder

notAreYouScared:
image

Currently, it just appears once it's loaded.

### DIFF
--- a/app/Livewire/ServerEntry.php
+++ b/app/Livewire/ServerEntry.php
@@ -37,7 +37,12 @@ class ServerEntry extends Component
                     "></div>
                 @endif
 
-                <div class="flex items-center mb-5 gap-2">
+
+                @if ($server->description)
+                    <div class="flex items-center gap-2">
+                @else
+                    <div class="flex items-center mb-5 gap-2">
+                @endif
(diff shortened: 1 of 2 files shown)


app/Livewire/ServerEntry.php
```diff
@@ -37,7 +37,12 @@ class="absolute left-0 top-1 bottom-0 w-1 rounded-lg"
                     "></div>
                 @endif
 
-                <div class="flex items-center mb-5 gap-2">
+
+                @if ($server->description)
+                    <div class="flex items-center gap-2">
+                @else
+                    <div class="flex items-center mb-5 gap-2">
+                @endif
```
rmartinoscar, line 45:
```suggestion
                <div @class([
                    'flex items-center gap-2',
                    'mb-5' => !$server->description,
                ])>
```
Please use a @class Attribute merge instead.

Just like https://github.com/pelican-dev/panel/blob/886836c60a8d8c531b83d711b464ebaadcef58ec/resources/views/livewire/server-entry.blade.php#L25-L28

rmartinoscar (dismissed):

PR #1911, opened 2025-11-18
Refactor schedule tasks

Boy132:
Same principle as oauth, captcha and avatars. Allows to register custom schedule tasks.

### DIFF
--- a/app/Filament/Server/Resources/Schedules/RelationManagers/TasksRelationManager.php
+++ b/app/Filament/Server/Resources/Schedules/RelationManagers/TasksRelationManager.php
@@ -51,7 +51,7 @@ class TasksRelationManager extends RelationManager
             Group::make(fn (Get $get) => array_key_exists($get('action'), $tasks) ? $tasks[$get('action')]->getPayloadForm() : []),
             TextInput::make('time_offset')
                 ->label(trans('server/schedule.tasks.time_offset'))
-                ->hidden(fn (Get $get) => config('queue.default') === 'sync' || $get('sequence_id') === 1 || $schedule->tasks->isEmpty())
+                ->hidden(fn (Get $get, ?Task $task) => config('queue.default') === 'sync' || $schedule->tasks->isEmpty() || $task?->isFirst())
                 ->default(0)
                 ->numeric()
                 ->minValue(0)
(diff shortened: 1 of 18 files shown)


app/Extensions/Tasks/Schemas/CreateBackupSchema.php
```diff
@@ -0,0 +1,32 @@
...
+use App\Services\Backups\InitiateBackupService;
+
+final class CreateBackupSchema extends TaskSchema
+{
+    public function __construct(private InitiateBackupService $backupService) {}
+
+    public function getId(): string
+    {
+        return Task::ACTION_BACKUP;
+    }
```
rmartinoscar, line 16: We should get rid of all those const.

app/Filament/Server/Resources/Schedules/RelationManagers/TasksRelationManager.php
```diff
@@ -2,74 +2,51 @@
...
-                ->label(trans('server/schedule.tasks.actions.power.action'))
-                ->required()
-                ->options([
-                    'start' => trans('server/schedule.tasks.actions.power.start'),
-                    'restart' => trans('server/schedule.tasks.actions.power.restart'),
-                    'stop' => trans('server/schedule.tasks.actions.power.stop'),
-                    'kill' => trans('server/schedule.tasks.actions.power.kill'),
-                ])
+                ->disableOptionWhen(fn (string $value) => !$tasks[$value]->canCreate($schedule))
+                ->options(Arr::mapWithKeys($tasks, fn (TaskSchemaInterface $task) => [$task->getId() => $task->getName()]))
```
rmartinoscar: Move to TaskService ?
https://github.com/pelican-dev/panel/blob/d78c977d7574097e14a8f5c54fbbc7cea61d9218/app/Extensions/Avatar/AvatarService.php#L50-L54

app/Extensions/Tasks/TaskSchemaInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
rmartinoscar, line 1: Add this to the Task model aswell so we can do `$task->getName()` instead of that `$tasks[$task->action]->getName()`

Boy132: "this"?

rmartinoscar: I meant the methods implemented by this interface

rmartinoscar:
image
When deleting last task then you can't create new ones after.

app/Filament/Server/Resources/Schedules/RelationManagers/TasksRelationManager.php
```diff
@@ -26,50 +27,28 @@ class TasksRelationManager extends RelationManager
...
-                    'start' => trans('server/schedule.tasks.actions.power.start'),
-                    'restart' => trans('server/schedule.tasks.actions.power.restart'),
-                    'stop' => trans('server/schedule.tasks.actions.power.stop'),
-                    'kill' => trans('server/schedule.tasks.actions.power.kill'),
-                ])
-                ->selectablePlaceholder(false)
-                ->default('restart'),
+                ->default(array_key_first($tasks))
+                ->afterStateUpdated(fn ($state, Set $set) => $set('payload', $tasks[$state]->getDefaultPayload())),
+            Group::make(fn (Get $get) => array_key_exists($get('action'), $tasks) ? $tasks[$get('action')]->getPayloadForm() : []),
```
rmartinoscar, line 51: Is it even worth adding the group if the task schema doesn't provide payloadForm ?

Boy132: Yes, there always is a payload form. The array_key_exists is only necessary because of a little race condition: the group schema is loaded before the default value, so for a short time `$get('action')` will return an empty string instead of the default task type.

rmartinoscar (approved):

PR #1939, opened 2025-11-27
feat: add support for changing docker labels via application api

m41denx:
This PR adds support for changing docker labels when creating server or modifying its details

Reasons for implementation:
- Pelican already supports editing docker labels via filament
- Support for container labels is one of Pelican's killer features that allows to use containers created by wings with sidecars (traefik, calico, node exporters and log collectors) so labels just have to have an API support

### DIFF
--- a/app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
+++ b/app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
@@ -33,7 +33,7 @@ class StoreServerRequest extends ApplicationApiRequest
             'egg' => $rules['egg_id'],
             'docker_image' => 'sometimes|string',
             'docker_labels' => ['sometimes', 'array', new DockerLabel()],
-            'docker_labels.*' => 'string',
+            'docker_labels.*' => 'required|string',
             'startup' => 'sometimes|string',
             'environment' => 'present|array',
             'skip_scripts' => 'sometimes|boolean',
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -185,7 +185,7 @@ class Server extends Model implements HasAvatar, Validatable
         'image' => ['required', 'string', 'max:255'],
         'icon' => ['sometimes', 'nullable', 'string'],
         'docker_labels' => ['array'],
-        'docker_labels.*' => ['string'],
+        'docker_labels.*' => ['required', 'string'],
         'database_limit' => ['present', 'nullable', 'integer', 'min:0'],
         'allocation_limit' => ['sometimes', 'nullable', 'integer', 'min:0'],
         'backup_limit' => ['present', 'nullable', 'integer', 'min:0'],


app/Services/Servers/DetailsModificationService.php
```diff
@@ -36,12 +37,18 @@ public function handle(Server $server, array $data): Server
...
                 'external_id' => Arr::get($data, 'external_id'),
                 'owner_id' => Arr::get($data, 'owner_id'),
                 'name' => Arr::get($data, 'name'),
                 'description' => Arr::get($data, 'description') ?? '',
-            ])->saveOrFail();
+            ];
+
+            if (array_key_exists('docker_labels', $data)) {
+                $attributes['docker_labels'] = Arr::get($data, 'docker_labels');
+            }
```
rmartinoscar:
```suggestion
            if ($labels = Arr::get($data, 'docker_labels')) {
                $attributes['docker_labels'] = $labels;
            }
```

app/Http/Requests/Api/Application/Servers/UpdateServerDetailsRequest.php
```diff
@@ -18,22 +18,30 @@ public function rules(): array
...
     /**
      * Convert the posted data into the correct format that is expected by the application.
      *
-     * @return array<array-key, string>
+     * @return array<string, mixed>
      */
     public function validated($key = null, $default = null): array
     {
-        return [
+        $payload = [
```
rmartinoscar:
```suggestion
        $attributes = [
```

app/Http/Requests/Api/Application/Servers/UpdateServerDetailsRequest.php
```diff
@@ -18,22 +18,30 @@ public function rules(): array
...
             'name' => $this->input('name'),
             'owner_id' => $this->input('user'),
             'description' => $this->input('description'),
         ];
+
+        if ($this->has('docker_labels')) {
+            $payload['docker_labels'] = $this->input('docker_labels');
+        }
+
+        return $payload;
```
rmartinoscar:
```suggestion
        if ($this->has('docker_labels')) {
            $attributes['docker_labels'] = $this->input('docker_labels');
        }

        return $attributes;
```

app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
```diff
@@ -29,6 +29,8 @@ public function rules(): array
             'user' => $rules['owner_id'],
             'egg' => $rules['egg_id'],
             'docker_image' => 'sometimes|string',
+            'docker_labels' => 'sometimes|array',
+            'docker_labels.*' => 'string',
```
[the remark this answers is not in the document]
m41denx: This is just an adapter, json specification doesnt allow keys to be non-strings

app/Models/Server.php
```diff
@@ -184,6 +184,8 @@ class Server extends Model implements HasAvatar, Validatable
         'skip_scripts' => ['sometimes', 'boolean'],
         'image' => ['required', 'string', 'max:255'],
         'icon' => ['sometimes', 'nullable', 'string'],
+        'docker_labels' => ['array'],
+        'docker_labels.*' => ['string'],
```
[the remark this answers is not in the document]
m41denx: this is an adapter, json input cannot have keys as non-strings

app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
rmartinoscar:
```suggestion
            'docker_labels.*' => 'required|string',
```
If you give `docker_labels` it shouldn't be an empty array.

app/Models/Server.php
rmartinoscar:
```suggestion
        'docker_labels.*' => ['required', 'string'],
```
Same here

app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
m41denx: Why not? Isn't absence of tags a valid state?

rmartinoscar (changes requested):
Thanks for your contribution great work so far !
I think we should enforce [labels restrictions](https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations) to make sure you don't brick the server.

m41denx:
> Thanks for your contribution great work so far ! I think we should enforce [labels restrictions](https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations) to make sure you don't brick the server.

Should we create a separate validator or simple inline regex would be enough?

rmartinoscar:
image
image
image

rmartinoscar:
image
image

rmartinoscar:
> Should we create a separate validator or simple inline regex would be enough?

simple regex is fine.

m41denx:
I'll post regex here while I read docs for laravel validator

```
^(?!(?:com\.docker\.|io\.docker\.|org\.dockerproject\.))(?=.*[a-z]$)[a-z](?:[a-z0-9]|(?<!\.)\.(?!\.)|(?<!-)-(?!-))*$
``` 
Regex101: https://regex101.com/r/FiYrwo/1
Validators:
- `(?!(?:com\.docker\.|io\.docker\.|org\.dockerproject\.))` → The `com.docker.*`, `io.docker.*`, and `org.dockerproject.*` namespaces are reserved by Docker for internal use.
- `(?=.*[a-z]$)` → Positive lookahead. Label keys should ... end with a lower-case letter ...
- `[a-z]` → Label keys should begin ... with a lower-case letter ...
- `(?:[a-z0-9]|(?<!\.)\.(?!\.)|(?<!-)-(?!-))*` → Non-capturing (+ performance), `*` - any amount of symbols as first and optional last symbols are validated by expressions before
  - `[a-z0-9]` → Label keys ... should only contain lower-case alphanumeric characters, ...
  - `(?<!\.)\.(?!\.)` → NLA + NLB for dots to check for consecutive dots (before or after current dot)
  - `(?<!-)-(?!-)`→ NLA + NLB for hyphens to check for consecutive hyphens (before or after current hyphen)

app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
```diff
@@ -29,6 +29,10 @@ public function rules(): array
             'user' => $rules['owner_id'],
             'egg' => $rules['egg_id'],
             'docker_image' => 'sometimes|string',
+            'docker_labels' => 'sometimes|array',
+            // Docker labels are validated via https://regex101.com/r/FiYrwo/1 following Docker key format
+            // recommendations: https://docs.docker.com/engine/manage-resources/labels/
+            'docker_labels.*' => 'required|regex:/^(?!(?:com\.docker\.|io\.docker\.|org\.dockerproject\.))(?=.*[a-z]$)[a-z](?:[a-z0-9]|(?<!\.)\.(?!\.)|(?<!-)-(?!-))*$/',
```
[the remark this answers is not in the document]
m41denx: uhh, I'll wait for maintainers response, I have little idea how laravel works besides basic http contracts

app/Http/Requests/Api/Application/Servers/StoreServerRequest.php
```diff
@@ -22,13 +23,15 @@
...
~
~        return [
~            'external_id' => $rules['external_id'],
~            'name' => $rules['name'],
~            'description' => array_merge(['nullable'], $rules['description']),
             'user' => $rules['owner_id'],
             'egg' => $rules['egg_id'],
             'docker_image' => 'sometimes|string',
+            'docker_labels' => ['sometimes', 'array', new DockerLabel],
+            'docker_labels.*' => 'required|string',
```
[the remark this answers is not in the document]
m41denx: indeed, but doesnt allowing empty values result in laravel allowing null values and other form of labels corruption lke accepting flat array instead of kv one?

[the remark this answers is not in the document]
m41denx: can you update tests added by this pr to accomodate for scenarios we discussed?

m41denx:
I actually found out docker allows weird stuff like here: https://regex101.com/r/FiYrwo/2

app/Models/Server.php
```diff
@@ -184,6 +184,8 @@ class Server extends Model implements HasAvatar, Validatable
         'skip_scripts' => ['sometimes', 'boolean'],
         'image' => ['required', 'string', 'max:255'],
         'icon' => ['sometimes', 'nullable', 'string'],
+        'docker_labels' => ['array'],
+        'docker_labels.*' => ['required', 'string'],
```
rmartinoscar, line 188: Add `use HasValidation { getRules as getValidationRules; }`

```php
    /**
     * @return array<string|string[]>
     */
    public static function getRules(): array
    {
        $rules = self::getValidationRules();

        $rules['docker_labels.*'][] = new DockerLabel();

        return $rules;
    }
```

Boy132:
> [...] I think we should enforce [labels restrictions](https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations) to make sure you don't brick the server.

They aren't enforced by docker and won't brick a server. We also don't have any validation on the frontend. There is no reason to make this overly complex.

app/Rules/DockerLabel.php
```diff
@@ -0,0 +1,27 @@
...
+
+class DockerLabel implements ValidationRule
+{
+    /**
+     * Run the validation rule.
+     *
+     * @param  \Closure(string, ?string=): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     */
+    public function validate(string $attribute, mixed $value, Closure $fail): void
+    {
```
lancepioch, line 16: Check if it's an array

```suggestion
    {
        if (!is_array($value)) {
            $fail("{$attribute} must be an array.");
            return;
        }

```

app/Rules/DockerLabel.php
```diff
@@ -0,0 +1,27 @@
...
+        foreach (array_keys($value) as $key) {
+            // Docker labels are validated via https://regex101.com/r/FiYrwo/2 following Docker key format
+            // recommendations: https://docs.docker.com/engine/manage-resources/labels/
+            if (!preg_match('/^(?!(?:com\.docker\.|io\.docker\.|org\.dockerproject\.))(?=.*[a-z]$)[a-z](?:[a-z0-9]|(?<!\.)\.(?!\.)|(?<!-)-(?!-)|\/|_)*$/', $key)) {
```
lancepioch, line 20: Please make this a constant with a description:

Keep those same comments with the constant.

@Boy132 we can piggyback off this later with real validation.

lancepioch (changes requested):

notAreYouScared:
Closed due to inactivity.

PR #1943, opened 2025-11-28
fix: child folder detection in ZIP for plugin extraction

UndercoverNL:
`locateName` does not support directory detection, fixed with new changes

Plugins can now correctly be uploaded as 1 zipped directory.

### DIFF
--- a/app/Services/Helpers/PluginService.php
+++ b/app/Services/Helpers/PluginService.php
@@ -338,7 +338,7 @@ class PluginService
             File::deleteDirectory(plugin_path($pluginName));
         }
 
-        $extractPath = $zip->locateName($pluginName . '/') !== false ? base_path('plugins') : plugin_path($pluginName);
+        $extractPath = $zip->locateName($pluginName . '/plugin.json') !== false ? base_path('plugins') : plugin_path($pluginName);
 
         if (!$zip->extractTo($extractPath)) {
             $zip->close();


UndercoverNL:

Boy132:
Thanks, but I don't understand the problem with `locateName`? It does detect directories?

I tried with a zip that directly contains the plugin data and a zip that has a folder with the plugin data and both imported fine.
grafik
grafik

UndercoverNL:
That's strange, on me it does not, with the current state of the branch when I upload: `my-plugin.zip/my-plugin/plugin.json` it creates `/root/plugins/my-plugin/my-plugin/plugin.json`, since when I dd `$zip->locateName($pluginName . '/')` it gives false even tho the zip structure is `my-plugin.zip/my-plugin/plugin.json`

UndercoverNL:
aaa I found the issue, so the issue is not `locateName` not detecting directories, but different ZIP tools not creating directory entries, for example, a ZIP tool which does not create a directory entry would only contain:
```
my-plugin/plugin.json
my-plugin/src/MyPluginPlugin.php
```
While a ZIP tool which does create a  directory entry would actually contain 
```
my-plugin/
my-plugin/plugin.json
my-plugin/src/MyPluginPlugin.php
```
Which would indeed work with the current code, so a more reliable way to check for this is by checking for the plugin.json file instead of the root directory it self:
```php
$extractPath = $zip->locateName($pluginName . '/plugin.json') !== false ? base_path('plugins') : plugin_path($pluginName);
```

PR #1968, opened 2025-12-08
Disable Captcha & Oauth Settings actions when read only

### LINKED ISSUE #1967: Admins with View Settings Permission can use Captcha & OAuth actions
Image
Image
Image

This is just a QoL not an actual security concern since they won't be able to save.

rmartinoscar:
Closes #1967

I used `disabled()` instead of `authorize()` cause i thought the section would look weird if it was completely empty.


app/Filament/Admin/Pages/Settings.php
```diff
@@ -297,11 +297,13 @@ private function captchaSettings(): array
                     Actions::make([
```
Boy132, line 297: I would add `visible(fn () => user()?->can('update settings'))` to the `Actions` instead.

Boy132: Nevermind, I see now what you meant by "empty section looking weird" :D

PR #1990, opened 2025-12-12
Fix console duplicating with spa

notAreYouScared:
_Should_ Fix https://github.com/pelican-dev/panel/issues/1241

This should be tested across multiple systems / multiple people should test this.


rmartinoscar:
It does fix the spa issue where the console would get duplicated however the websocket still gets duped.

notAreYouScared:
> It does fix the spa issue where the console would get duplicated however the websocket still gets duped.

How so? Enable Preserve Log in dev tools, and look at the old websockets; they don't have any data coming in, so how is it duplicated?

PR #1994, opened 2025-12-13
simplify development setup

LetterN:
i don't want to run each command individually when updating the repo (or restarting the dev server)

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -16,7 +16,7 @@ To start contributing to Pelican Panel, you need to have a basic understanding o
 2. Clone your Fork
 3. Install Dependencies (PHP modules, composer, node & yarn)
 4. Run development setup `composer run dev-setup`
-5. Create a use if you havent `php artisan p:user:make`
+5. Create a user if you havent `php artisan p:user:make`
 6. Run `composer run dev`
 
 As IDE we recommend [Visual Studio](https://visualstudio.microsoft.com)/ [Visual Studio Code](https://code.visualstudio.com) (free) or [PhpStorm](https://www.jetbrains.com/phpstorm) (paid).


composer.json
```diff
@@ -82,6 +82,16 @@
...
+        ],
+        "dev-setup": [
+            "yarn install",
+            "composer install",
+            "php artisan p:environment:setup",
+            "php artisan migrate --force -n"
+        ],
+        "dev": [
+            "Composer\\Config::disableProcessTimeout",
+            "npx concurrently -k -c \"#93c5fd,#c4b5fd,#d4d4d8,#fdba74\" \"php artisan serve\" \"php artisan queue:listen --tries=1\" \"php artisan pail\" \"yarn run dev\" --names=server,queue,logs,vite"
```
rmartinoscar, line 94: This was removed in https://github.com/pelican-dev/panel/pull/1557 i don't see why adding it back.

Boy132:
You don't need to run all the setup commands again when updating?

Also, an external webserver is highly recommended, using `php artisan serve` is unreliable. That's why we recommend using Laravel Herd - it comes with everything you need.

LetterN:
> You don't need to run all the setup commands again when updating?

oh yeah you do oops
 
> Also, an external webserver is highly recommended, using `php artisan serve` is unreliable. That's why we recommend using Laravel Herd - it comes with everything you need.

i'm using vscode and on linux so this isnt available either and the webserver works decently well (runtime reloading and such)

Boy132:
> i'm using vscode and on linux so this isnt available either and the webserver works decently well (runtime reloading and such)

Herd isn't available on Linux because installing packages is very trivial there.

Overall, I don't see the benefit of this PR.

PR #1998, opened 2025-12-13
View pages

JoanFo1456:
Continues with Boy's work (#583)
This adds the missing view pages, making permissions actually useful.
I followed what I thought it would be view, if that's not the real case, let me know and I'll change it.


JoanFo1456:
I only made Rabbit suggestions on things I actually messed up, everything else is already as it was before, so unless someone asks to change it, will stay like this.

lancepioch:
@Boy132 I'll let you be the judge!

lancepioch:
@JoanFo1456 Mark as ready once the conflicts are resolved, tyvm!

notAreYouScared:
Closed due to inactivity.

Issue #1976, opened 2025-12-10
Cannot see other servers with admin permission


PaulaBras:
### Current Behavior

I use OAuth to create and manage my users. Sadly, I cannot assign the Admin role to my admin user using OAuth. So I added a group with all permissions.
With the latest update, I cannot see other users servers without them manually adding me, despite having all the permissions.

### Expected Behavior

Assign the Admin Group to any user.
See all servers.

### Steps to Reproduce

OAuth User with seperate admin Group with all permissions selected.

### Panel Version

1.0.0-beta29

### Wings Version

1.0.0-beta20

### Error Logs

```bash

```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue.
- [x] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

mitchell-thompson-CS:
Also having this issue, hopefully the linked PR solves this issue!

PR #2002, opened 2025-12-14
Fix role-based server access for users with node-restricted permissions

memesalot:
## Description

Fixes a bug where users assigned to roles with server permissions and node restrictions could not see or access servers on their assigned nodes. Users could only see servers they directly owned, even when their role granted them broader access.

### `app/Models/Node.php`
- Changed the `roles()` relationship from `HasManyThrough` to `BelongsToMany` to correctly handle the many-to-many relationship through the `node_role` pivot table.

### `app/Models/User.php`
- Added `canViewServers()` method that directly checks for the `viewList server` permission using Spatie's `hasPermissionTo()`, bypassing the fragile policy chain.
- Updated `accessibleServers()` to use the new `canViewServers()` method instead of `canned('viewAny', Server::class)`.
- Updated `canAccessTenant()` to check for role-based access when the user is not an owner or subuser.

### `app/Policies/ServerPolicy.php`
- Modified `before()` to directly check role-based permissions using `hasPermissionTo()` after verifying the user can target the server's node.

## Root Cause

The `Node::roles()` relationship was incorrectly defined as `HasManyThrough` instead of `BelongsToMany`, causing `accessibleNodes()` queries to return incorrect results. Additionally, permission checks were relying on a complex policy chain that wasn't properly triggering Spatie's permission verification.

### DIFF
--- a/app/Models/Node.php
+++ b/app/Models/Node.php
@@ -4,7 +4,6 @@ namespace App\Models;
 
 use App\Contracts\Validatable;
 use App\Exceptions\Service\HasActiveServersException;
-use App\Models\NodeRole;
 use App\Repositories\Daemon\DaemonSystemRepository;
 use App\Traits\HasValidation;
 use Carbon\Carbon;


memesalot:

app/Models/Node.php
```diff
@@ -277,9 +276,9 @@ public function databaseHosts(): BelongsToMany
         return $this->belongsToMany(DatabaseHost::class);
     }
 
-    public function roles(): HasManyThrough
+    public function roles(): BelongsToMany
     {
-        return $this->hasManyThrough(Role::class, NodeRole::class, 'node_id', 'id', 'id', 'role_id');
+        return $this->belongsToMany(Role::class, 'node_role', 'node_id', 'role_id');
     }
```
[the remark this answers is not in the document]
memesalot: Fixed

rmartinoscar:
Hey thanks for your contribution !
I believe https://github.com/pelican-dev/panel/pull/1978 already took care of #1976 so what's more in this ?

Boy132 (changes requested):
The only change I agree with is changing `HasManyThrough` to `BelongsToMany`.
The rest of the changes just workaround policies. Which is not desirable.

memesalot:
> Hey thanks for your contribution ! I believe #1978 already took care of #1976 so what's more in this ?

Tested out 1978 - still had the same issue. This is what fixed it for me.

app/Models/User.php
```diff
@@ -278,6 +278,22 @@ public function accessibleServers(): Builder
...
+     */
+    public function canViewServers(): bool
+    {
+        if ($this->isRootAdmin()) {
+            return true;
+        }
+
+        try {
+            return $this->hasPermissionTo('viewList server');
+        } catch (\Spatie\Permission\Exceptions\PermissionDoesNotExist) {
```
lancepioch, line 292: Import please

app/Models/User.php
```diff
@@ -278,6 +278,22 @@ public function accessibleServers(): Builder
...
+     * Check if the user has permission to view servers via role permissions.
+     */
+    public function canViewServers(): bool
+    {
+        if ($this->isRootAdmin()) {
+            return true;
+        }
+
+        try {
+            return $this->hasPermissionTo('viewList server');
```
lancepioch, line 291: Don't we have enums?

app/Policies/ServerPolicy.php
```diff
@@ -40,6 +40,16 @@ public function before(User $user, string $ability, string|Server $server): ?boo
             return false;
         }
 
+        // Check if user has role-based permission for this specific ability
+        $permissionName = $ability . ' ' . $this->modelName;
+        try {
+            if ($user->hasPermissionTo($permissionName)) {
+                return true;
+            }
+        } catch (\Spatie\Permission\Exceptions\PermissionDoesNotExist) {
```
lancepioch, line 49: Import class please

app/Policies/ServerPolicy.php
```diff
@@ -40,6 +40,16 @@ public function before(User $user, string $ability, string|Server $server): ?boo
             return false;
         }
 
+        // Check if user has role-based permission for this specific ability
+        $permissionName = $ability . ' ' . $this->modelName;
```
lancepioch, line 44:
```suggestion
        $permissionName = "$ability $this->modelName";
```

Isn't there a better way to do this @Boy132 ?

Boy132:
I'm going to close this for now because this is not a desired solution.
Feel free to open an issue if you still experience problems or create a new PR. Thanks!

PR #2004, opened 2025-12-14
Add Eggs to Installer

notAreYouScared:
todo: Add stock eggs to pelican-eggs repos

Adds eggs to the installer to allow users to select the eggs they'd like to install when they set up the panel. Egg installs are then queued to prevent locking up the installer and crashing due to the php time limit of 30 seconds. 

chrome_wE78kRzQi0

### DIFF
--- a/app/Jobs/InstallEgg.php
+++ b/app/Jobs/InstallEgg.php
@@ -8,6 +8,7 @@ use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Log;
 use Throwable;
 
 class InstallEgg implements ShouldQueue


notAreYouScared:
@coderabbitai review

database/Seeders/DatabaseSeeder.php
```diff
@@ -1,19 +0,0 @@
...
-class DatabaseSeeder extends Seeder
-{
-    /**
-     * Run the database seeds.
-     */
-    public function run(): void
-    {
-        $this->call(EggSeeder::class);
-
-        Role::firstOrCreate(['name' => Role::ROOT_ADMIN]);
```
rmartinoscar, line 18: Pretty sure this is mandatory

rmartinoscar (dismissed):
Are you sure its safe to remove the ROOT_ADMIN seeding ?

There's a tone of egg that doesn't exist outside of the Seeders, 

Simply doesn't exists:
`bungeecord`, `forge`, `vanilla mc`, `mumble`, `teamspeak3`

Exist but are different
`rust`, `gmod`, `insurgency`,

database/Seeders/DatabaseSeeder.php
notAreYouScared: 100% thought it was only seeding eggs 😄 🤣

notAreYouScared: But I was also a root admin when I installed the panel....

Leon-JavaScript:
Will it after this update also be possible to delete the default eggs that are currently always getting recreated with every update?

notAreYouScared:
> Will it after this update also be possible to delete the default eggs that are currently always getting recreated with every update?

You've always been able to delete the stock eggs. They just won't be auto-installed anymore.

notAreYouScared:
@coderabbitai review

app/Jobs/InstallEgg.php
```diff
@@ -0,0 +1,42 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    public string $downloadUrl;
+
+    public int $timeout = 15;
+
+    public function __construct(string $downloadUrl)
+    {
+        $this->downloadUrl = $downloadUrl;
+    }
```
lancepioch:
```suggestion
    public int $timeout = 15;

    public function __construct(public string $downloadUrl) {}
```

Simplify

app/Jobs/InstallEgg.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function handle(EggImporterService $eggImporterService): void
+    {
+        try {
+            $eggImporterService->fromUrl($this->downloadUrl);
+        } catch (Throwable $e) {
+            Log::error('InstallEgg job failed.', [
+                'downloadUrl' => $this->downloadUrl,
+                'exception' => $e->getMessage(),
+            ]);
```
lancepioch:
```suggestion
            report($e);

```

app/Livewire/Installer/PanelInstaller.php
```diff
@@ -53,7 +56,7 @@ public function getTitle(): string
 
     public function getMaxContentWidth(): Width|string
     {
-        return Width::SevenExtraLarge;
+        return Width::ScreenTwoExtraLarge;
```
lancepioch, line 59: Should this be in this PR?

app/Livewire/Installer/PanelInstaller.php
```diff
@@ -220,4 +227,52 @@ public function createAdminUser(UserCreationService $userCreationService): User
...
         }
     }
+
+    public function installEggs(EggImporterService $eggImporterService): void
+    {
+        try {
+            $selectedEggs = array_get($this->data, 'eggs', []);
+            $queueDriver = config('queue.default');
+
+            if ($queueDriver !== 'sync') {
```
lancepioch: You shouldn't need this. Dispatch will automatically work sync if that's the driver when you call `dispatch`

lancepioch (dismissed):
Looks good! @rmartinoscar says that we need to add the repo eggs to the pelican-eggs org

app/Livewire/Installer/PanelInstaller.php
notAreYouScared: Yes

app/Jobs/InstallEgg.php
notAreYouScared: I tweaked it, reporting just the exception does not provide the download url that failed. So we have no way of knowing what failed.

app/Jobs/InstallEgg.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * @throws Throwable
+     */
+    public function handle(EggImporterService $eggImporterService): void
+    {
+        try {
+            $eggImporterService->fromUrl($this->downloadUrl);
+        } catch (Throwable $e) {
+            Log::error('Failed to install egg from URL: ' . $this->downloadUrl, ['exception' => $e]);
+        }
```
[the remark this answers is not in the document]
notAreYouScared: there are no more retries?

notAreYouScared:
> martin says that we need to add the repo eggs to the pelican-eggs org

Correct, and I asked the Egg people, and provided the egg files, in the Discord channel.... two weeks ago. It's on them now.

PR #2009, opened 2025-12-16
Fix: egg images are not loading

hallo123wert:


Boy132:
Could you also update the server model?
https://github.com/pelican-dev/panel/blob/8b5f33ee7191cf7c4c9ca8169496d4391e129e3f/app/Models/Server.php#L541

_I swear the icons **did** work in local dev environment..._

PR #2014, opened 2025-12-18
feat(activity-log): Add event-based icons to activity log entries

killerbite95:
## Summary
Previously, activity log icons only displayed the actor type (user, system, API). This change adds event-specific icons that match the panel's icon conventions used in SubuserPermission and other resources.

## Changes
- Add \EVENT_ICON_MAP\ constant with 18 event group mappings
- Icons include: backup, file, database, schedule, allocation, startup, settings, subuser, console, control, power, sftp, server, auth, user, api-key, reset-password, ssh
- Extract \getActorIcon()\ as fallback for unmapped events
- Parse event string to determine group (e.g., \server:backup.create\  \ackup\)

## Benefits
- Visual consistency with the rest of the panel
- Better UX when browsing activity logs - users can quickly identify the type of action
- Uses existing tabler icons already in use throughout the panel

## Before/After
| Before | After |
|--------|-------|
| All entries show actor icon (user/system/api) | Entries show event-specific icons (backup, database, file, etc.) |

### DIFF
--- a/app/Models/ActivityLog.php
+++ b/app/Models/ActivityLog.php
@@ -143,7 +143,66 @@ class ActivityLog extends Model implements HasIcon, HasLabel
         });
     }
 
+    /**
+     * Icon mapping for activity event groups.
+     * Consistent with icons used throughout the panel (SubuserPermission, Resources, etc.)
+     */
+    private const EVENT_ICON_MAP = [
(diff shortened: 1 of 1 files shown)


killerbite95:
Closing to recreate with proper branch name (fix/activity-log-icons instead of main)

PR #2015, opened 2025-12-18
Feat(activity-log): Add event-based icons to activity log entries

killerbite95:
Activity log icons now show the event type (backup, database, files, etc.) instead of just the actor type (user/system/api).

Adds an EVENT_ICON_MAP with mappings for common event groups, falling back to actor-based icons for unmapped events.

### DIFF
--- a/app/Models/ActivityLog.php
+++ b/app/Models/ActivityLog.php
@@ -143,7 +143,66 @@ class ActivityLog extends Model implements HasIcon, HasLabel
         });
     }
 
+    /**
+     * Icon mapping for activity event groups.
+     * Consistent with icons used throughout the panel (SubuserPermission, Resources, etc.)
+     */
+    private const EVENT_ICON_MAP = [
+        'backup' => 'tabler-file-zip',
+        'file' => 'tabler-files',
+        'database' => 'tabler-database',
+        'schedule' => 'tabler-clock',
+        'allocation' => 'tabler-network',
+        'startup' => 'tabler-player-play',
+        'settings' => 'tabler-settings',
+        'subuser' => 'tabler-users',
+        'console' => 'tabler-terminal-2',
+        'control' => 'tabler-terminal-2',
+        'power' => 'tabler-power',
+        'sftp' => 'tabler-folder-share',
+        'server' => 'tabler-server',
+        'auth' => 'tabler-shield',
+        'user' => 'tabler-user',
+        'api-key' => 'tabler-key',
+        'reset-password' => 'tabler-key',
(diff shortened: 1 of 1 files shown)


killerbite95:

Boy132:
Not sure how I'm feeling about this. With these changes there is no way to tell the actor type anymore?

app/Models/ActivityLog.php
```diff
@@ -143,7 +143,66 @@ protected static function boot(): void
...
+            : explode('.', $eventParts[0])[0];
+
+        // Return event-specific icon if available
+        if (isset(self::EVENT_ICON_MAP[$group])) {
+            return self::EVENT_ICON_MAP[$group];
+        }
+
+        // Fallback to actor-based icon
+        return $this->getActorIcon();
+    }
```
lancepioch, line 195: Simplify this by using the Laravel String library instead:

```suggestion
public function getIcon(): string
{
    $group = str($this->event)
        ->after(':')
        ->before('.')
        ->toString();

    return self::EVENT_ICON_MAP[$group] ?? $this->getActorIcon();
}
```

app/Models/ActivityLog.php
```diff
@@ -143,7 +143,66 @@ protected static function boot(): void
         });
     }
 
+    /**
+     * Icon mapping for activity event groups.
+     * Consistent with icons used throughout the panel (SubuserPermission, Resources, etc.)
+     */
+    private const EVENT_ICON_MAP = [
```
lancepioch, line 150: Please translate this into a new enum! Something like ActivityEventIcon

TYVM

lancepioch (changes requested):
Looks great, thank you, 2 quick change requests!

app/Models/ActivityLog.php
```diff
@@ -160,7 +219,7 @@ public function getLabel(): string
     {
         $properties = $this->wrapProperties();
 
-        return trans_choice('activity.'.str($this->event)->replace(':', '.'), array_key_exists('count', $properties) ? $properties['count'] : 1, $properties);
+        return trans_choice('activity.' . str($this->event)->replace(':', '.'), array_key_exists('count', $properties) ? $properties['count'] : 1, $properties);
```
rmartinoscar, line 222:
```suggestion
        return trans_choice('activity.' . str($this->event)->replace(':', '.'), array_get('count', $properties, 1), $properties);
```
use laravel's helper instead of ternary

notAreYouScared:
Closed due to inactivity.

PR #2021, opened 2025-12-20
fix: Remove plugins directory before linking

DjordyKoert:
The `plugins` directory already exists during when creating the Symlinks. This causes the `ln` command to create a symlink inside the directory instead of replacing it.

This currently causes issues in dockerized installs, throwing errors like `ZipArchive::extractTo(): Permission denied` when attempting to install any plugin through the web panel.

Current incorrect structure:
```
/var/www/html
- plugins
  - plugins -> /pelican-data/plugins (Symlink)
```

Removing the existing plugins directory before creating the symlink. Gives us the correct structure:
```
/var/www/html
- plugins -> /pelican-data/plugins (Symlink)
```


rmartinoscar:
Superseded by #2051

PR #2028, opened 2025-12-21
add id to plugin list

notAreYouScared:
Adds id to plugin list, so if you need to use cli to interact with plugins, you can see the id, instead of guessing if its my-awesome-plugin or myawesomeplugin or some other random string...


Boy132:
You don't have to guess? If you don't run a command with a plugin id it will ask for it and display them in a table:
image

notAreYouScared:
image
image

Also, who knew you could just call `php artisan p:plugin:uninstall` 👍

PR #2031, opened 2025-12-21
Replace `Artisan::call` in plugin service for better error handling

Boy132:


app/Services/Helpers/PluginService.php
```diff
@@ -214,10 +215,11 @@ public function runPluginMigrations(Plugin $plugin): void
...
         if (file_exists($migrations)) {
-            $success = Artisan::call('migrate', ['--realpath' => true, '--path' => $migrations, '--force' => true]) === 0;
-
-            if (!$success) {
-                throw new Exception("Could not run migrations for plugin '{$plugin->id}'");
+            try {
+                $migrator = $this->app->make(Migrator::class);
+                $migrator->run($migrations);
+            } catch (Exception $exception) {
+                throw new Exception("Could not run migrations': " . $exception->getMessage());
```
rmartinoscar, line 222: I would keep the plugin id in the message.

Boy132: Why?

rmartinoscar: Easier when stacktrace doesn't include it

Boy132: These exceptions are shown on the frontend, when installing a specfic plugin. We also don't include the plugin id in other messages?

It previously was only there because there was no other error data.

rmartinoscar: nvm then

lancepioch (approved):

PR #2033, opened 2025-12-21
fix: plugin migration rollback and cache clearing on uninstall

PalmarHealer:
Fixes two critical bugs in the plugin uninstallation process that caused incomplete migration rollback and cache-related errors.

1. The `rollbackPluginMigrations()` method used `migrate:rollback`, which only rolls back the **last batch** of migrations. If a plugin had migrations run across multiple batches (e.g., initial install in batch 50, update adds migration in batch 75), only the most recent batch was rolled back, leaving orphaned tables and columns in the database.

2. After uninstalling a plugin, users encountered "Table not found" errors (e.g., `SQLSTATE[42S02]: Base table or view not found`).

### DIFF
--- a/app/Services/Helpers/PluginService.php
+++ b/app/Services/Helpers/PluginService.php
@@ -235,7 +235,7 @@ class PluginService
         if (file_exists($migrations)) {
             try {
                 $migrator = $this->app->make(Migrator::class);
-                $migrator->reset([$migrations]);
+                $migrator->reset($migrations);
             } catch (Exception $exception) {
                 throw new Exception("Could not rollback migrations': " . $exception->getMessage());
             }


PalmarHealer:
Collides with #2031

app/Services/Helpers/PluginService.php
```diff
@@ -330,6 +335,13 @@ public function uninstallPlugin(Plugin $plugin, bool $deleteFiles = false): void
             $this->buildAssets();
 
             $this->manageComposerPackages(oldPackages: $pluginPackages);
+
+            // This throws an error when not called with qualifier
+            foreach (\Filament\Facades\Filament::getPanels() as $panel) {
+                $panel->clearCachedComponents();
+            }
+
+            Artisan::call('optimize:clear');
```
Boy132: The `optimize:clear` call shouldn't be needed.

Boy132:
@PalmarHealer Please merge main and resolve the conflicts.

Boy132 (approved):

Issue #2018, opened 2025-12-20
Settings are no longer saving on the Panel


felinusfish:
### Current Behavior

Currently, on canary (Commit 1d5ace3) using a Docker installation, there appears to be a regression when trying to save the settings, as it does not save your settings after you have changed them. I am not 100% sure if this is an issue with Beta 30 or potentially an issue with Docker installations, But I'm sure somebody will chime in sooner or later on whether or not the issue exists for them outside of my specifications.

### Expected Behavior

The settings should be saving, but they are not.

### Steps to Reproduce

1. Log into an account with permissions to update the settings.
2. Try to change any setting.
3. Save, and notice that nothing has been modified.

### Panel Version

canary (1d5ace3)

### Wings Version

N/A

### Error Logs

```bash
There are no error logs.
```

### Is there an existing issue for this?

- [x] I have searched the existing issues before opening this issue.
- [x] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [x] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

notAreYouScared:
No error logs whatsoever? What settings? Profile? Server? Panel? There's alot of settings... no notifications??

parkervcp:
Which settings are you changing?

felinusfish:
Any. All of them are not saving in "Settings"! There were no errors in the logs.

notAreYouScared:
So youre editing the options in the "Settings" tab... so panel settings... when you click save what happens? Does the page refresh? Does it popup in the top corner that it failed?

felinusfish:
The page either refreshes afterwards when it says "Settings saved!" without changing the settings, or it doesn't refresh and says "Settings saved!" and manually refreshing shows the original settings, not the modified settings.

notAreYouScared:
Verify the .env has correct permissions, as all those settings live in it, and if it cant write to it...  then yes it would still show old values.

felinusfish:
What would have changed about the .env permissions between now and two days ago?

notAreYouScared:
I dont have access to your system, nor setup. So I cannot answer that. 

Did you check the permissions? Did you run the permission commands in the install guide to make sure they're correct? Then try again?

parkervcp:
I tested this on my a copy of the panel and while the UI refreshed after I changed the name the env file updated.

PalmarHealer:
Having the issue aswell. I have beta30 in docker

[...]

PR #2045, opened 2025-12-23
Fix xargs delimiter in entrypoint script

### LINKED ISSUE #2037: Docker: Panel doesn't start with spaces/special chars in variable values in .env
### Current Behavior

When having a space or special character in a variable name (example APP_NAME or MAIL_FROM_NAME) in the .env file the docker container of panel doesn't start anymore, due to the way how the environment variables are loaded in the entrypoint.sh script.

### Expected Behavior

The panel should start normally.

### Steps to Reproduce

Have a space in an environment variable (for example APP_NAME) and start the panel, the panel will not start.

### Panel Version

v1.0.0-beta30

### Wings Version

nA

_No response_

### Docker Image

_No response_

### Error Logs

```bash
/entrypoint.sh: export: line 6: -: bad variable name
external vars exist.
```

harryyoud:
Fixes #2037

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -3,7 +3,7 @@
 if [ -f /var/www/html/.env ]; then
   echo "external vars exist."
   # load env vars from .env
-  export $(grep -v '^#' .env | xargs)
+  . /var/www/html/.env
 else
   echo "external vars don't exist."
   # webroot .env is symlinked to this path


rmartinoscar:
Thanks for your contribution !
Please sign the CLA so we can merge it.

harryyoud:

docker/entrypoint.sh
```diff
@@ -3,7 +3,7 @@
 if [ -f /var/www/html/.env ]; then
   echo "external vars exist."
   # load env vars from .env
-  export $(grep -v '^#' .env | xargs)
+  export $(grep -v '^#' .env | xargs -d '\n')
```
parkervcp: From a conversation in discord replacing this line with `. /var/www/html/.env` should load the env file properly and fixes an issue with the ui not updating when making changes in the panel settings page.

parkervcp (changes requested):
I haven't gotten to this due to the holiday.

docker/entrypoint.sh
parkervcp: This change should also fix #2018

lancepioch:
My concern is that it will also just run any commands in the .env too.

My current thoughts are that even though it's less hacky to do it this "better" way, it could get some less experienced people in trouble. I'd say keep it as it is and encourage people to use real Docker env variables instead.

@parkervcp you are welcome to veto this, end decision is up to you

parkervcp:
> My concern is that it will also just run any commands in the .env too.

@lancepioch The only people editing the env file should be trusted individuals.

lancepioch:
> > My concern is that it will also just run any commands in the .env too.
> 
> @lancepioch The only people editing the env file should be trusted individuals.

While that is true, our interface does currently let any admins edit the .env variables - so that line starts to get blurred

PR #2048, opened 2025-12-24
chore: Prevent users from caching Config

### LINKED ISSUE #2018: Settings are no longer saving on the Panel

rmartinoscar:
Closes #2018 
When config is cached every plugin config is null cause they use `env()`, I couldn't find anyway to circumvent this so i skipped config caching entirely.

### DIFF
--- a/app/Traits/EnvironmentWriterTrait.php
+++ b/app/Traits/EnvironmentWriterTrait.php
@@ -3,6 +3,7 @@
 namespace App\Traits;
 
 use Illuminate\Support\Env;
+use Illuminate\Support\Facades\Artisan;
 use RuntimeException;
 
 trait EnvironmentWriterTrait
@@ -17,5 +18,6 @@ trait EnvironmentWriterTrait
     public function writeToEnvironment(array $values = []): void
     {
         Env::writeVariables($values, base_path('.env'), true);
+        Artisan::call('config:clear');
     }
 }
--- a/app/Console/Commands/Overrides/ConfigCacheCommand.php
+++ b/app/Console/Commands/Overrides/ConfigCacheCommand.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace App\Console\Commands\Overrides;
+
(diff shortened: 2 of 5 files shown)


app/Traits/EnvironmentWriterTrait.php
```diff
@@ -17,5 +18,6 @@ trait EnvironmentWriterTrait
     public function writeToEnvironment(array $values = []): void
     {
         Env::writeVariables($values, base_path('.env'), true);
+        Artisan::call('config:clear');
```
Boy132, line 21: Why was this added here?

rmartinoscar: Every time you write to env you have to reload it to be taken into account so instead of deduplicating this line I thought might as well do it in the trait so its simpler for plugins too

Boy132: So even without config cache you need to clear the config cache after writing to the .env?

rmartinoscar: Its in case you were using it before since there's no migration like stuff for that
I guess we could technically use a db migration for such things and .env changes but that doesn't seem proper

lancepioch (approved):

PR #2055, opened 2025-12-28
Bungeecord: Fix Download

QuintenQVD0:
# Changes

When downloading the latest version, it now throws a 301 and curl was not allowed to follow it so you always got a corrupted jar file


notAreYouScared:
@QuintenQVD0 this update needs to be done in the egg repo. As the panel eggs will be removed in the next beta.

PR #2086, opened 2026-01-08
Improve file browser UI

hUwUtao:
Minor changes to address some UX issue:

- Refresh: When delete or rename, no longer do a full refresh that jump the viewport around
- Show all files: To show all files beyond the limit
- Sticky header: Allows friendly interaction on mobile or when long list wanted for bulk-actions

### DIFF
--- a/resources/views/filament/server/pages/list-files.blade.php
+++ b/resources/views/filament/server/pages/list-files.blade.php
@@ -1,4 +1,16 @@
 <x-filament-panels::page>
+    @once
+        <style>
+            .files-selection-merged .fi-ta-header-ctn {
+                position: sticky;
+                top: 0;
+                z-index: 1;
+                -webkit-backdrop-filter: blur(8px);
+                backdrop-filter: blur(8px);
+            }
+        </style>
+    @endonce
+
     <div
         x-data="
         {


app/Filament/Server/Resources/Files/Pages/ListFiles.php
```diff
@@ -94,9 +94,10 @@ public function table(Table $table): Table
         $server = Filament::getTenant();
 
         return $table
-            ->paginated([25, 50, 100, 150, 200])
+            ->paginated([25, 50, 100, 150, 200, 'all'])
```
rmartinoscar:
```suggestion
            ->paginated([25, 50, 100, 150, 200])
```
This will make the browser unresponsive if there's too many files.

app/Filament/Server/Resources/Files/Pages/ListFiles.php
```diff
@@ -94,9 +94,10 @@ public function table(Table $table): Table
...
             ->paginationMode(PaginationMode::Simple)
             ->defaultPaginationPageOption(50)
+            ->selectCurrentPageOnly(false)
```
rmartinoscar:
```suggestion
```
Please revert aswell.

resources/views/filament/server/pages/list-files.blade.php
```diff
@@ -1,4 +1,16 @@
 <x-filament-panels::page>
+    @once
+        <style>
+            .files-selection-merged .fi-ta-header-ctn {
+                position: sticky;
+                top: 0;
+                z-index: 1;
```
rmartinoscar, line 7:
```suggestion
```
Is the `z-index` useful ?

rmartinoscar (changes requested):
Thanks for your contribution just a couple nitpicks 👍

app/Filament/Server/Resources/Files/Pages/ListFiles.php
hUwUtao: i simply want this option because i actually have a complaint on this specificly, but it seems not via, i might want to seek other ux solutions

resources/views/filament/server/pages/list-files.blade.php
hUwUtao: without z-index
image

with z-index
image
image

it is because the container of the bar sat before the items in order

app/Filament/Server/Resources/Files/Pages/ListFiles.php
notAreYouScared: The only issue i can see is that if the root folder has so many rows that it locks up the browsers, you'd have to guess a folder that exists to load the table with less rows, to be able to change it back to another value

app/Filament/Server/Resources/Files/Pages/ListFiles.php
notAreYouScared: I agree with @rmartinoscar,  People wouldn't expect the default functionality to select every item on every page.

notAreYouScared:
@hUwUtao I would like to take another look once our requested changes have been applied.

hUwUtao:
i've removed some questionable ux functionalities, guess only the layout change is good enough.
although there are few way i can reorganize this but i can't keep up with the complexity of fillament. have a great day.

PR #2099, opened 2026-01-12
Make sure custom pages/relations don't override default pages/relations

Boy132:
Protect against plugins overriding default pages/relations.


JoanFo1456:
I would say that until the panel has proper ways to implement custom elements, I think the override is better than just closing every way...

Boy132:
> I would say that until the panel has proper ways to implement custom elements, I think the override is better than just closing every way...

No, like I said on discord this was never intentional. I would rather have no solution than a very bad one.

rmartinoscar (approved):

PR #2113, opened 2026-01-16
Replace icon strings with enum

Boy132:


notAreYouScared:
What is this supposed to do? Improve performance? Just easier than doing the string way?

Boy132:
> What is this supposed to do? Improve performance? Just easier than doing the string way?

Hardcoded strings are always bad. Also, Filament has an [enum for Heroicons](https://github.com/filamentphp/filament/blob/5.x/packages/support/src/Icons/Heroicon.php) for quite a while. So I thought this might be useful. :)

PR #2114, opened 2026-01-16
Update to filament v5, Livewire v4

notAreYouScared:
This needs to be tested, some livewire stuff _might_ be broken.


notAreYouScared:
I mean, i don't think anything is `hard` about this. It all appears to just work.

PR #2122, opened 2026-01-18
fix plugins in entrypoint

parkervcp:
I forgot to mkdir for plugins


HttpRafa:
@parkervcp Will this cause any issues when using the Docker image? I’m worried about bricking the panel by installing plugins.

Boy132 (approved):

PR #2143, opened 2026-01-27
Email Localization for Notifications

lancepioch:
Implement multi-language email support for the Pelican Panel application, enabling emails to be sent in the recipient's preferred language. This feature was identified during code review of PR #2041 and addresses the need for internationalized email communications across all notification types (password resets, server notifications, account management, etc.). The implementation will leverage Laravel's existing localization infrastructure and extend it to the email notification system.


lancepioch:
Was curious what this would look like, yucky lol

PR #2155, opened 2026-01-30
Fix the tabbing for username to password on login

lancepioch:
1. Type in username/email
2. Hit tab to go to password input
3. Notice it works now

### DIFF
--- a/app/Filament/Pages/Auth/Login.php
+++ b/app/Filament/Pages/Auth/Login.php
@@ -11,8 +11,6 @@ use Filament\Schemas\Components\Actions;
 use Filament\Schemas\Components\Component;
 use Filament\Schemas\Schema;
 use Filament\Support\Colors\Color;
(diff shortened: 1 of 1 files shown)


app/Filament/Pages/Auth/Login.php
```diff
@@ -66,6 +68,18 @@ protected function getLoginFormComponent(): Component
             ->extraInputAttributes(['tabindex' => 1]);
     }
 
+    protected function getPasswordFormComponent(): Component
+    {
+        return TextInput::make('password')
```
Boy132: Use `parent::getPasswordFormComponent()` again but store it in a variable and use phpdoc to make it a textinput.

```php
/** @var TextInput $component */
$component = parent::getPasswordFormComponent();

return $component->extraInputAttributes(['tabindex' => 2]);
```

Boy132 (changes requested):

PR #2202, opened 2026-02-08
container: rework container stack to frankenphp runtime

hUwUtao:
Hi

I just want to optimize further for the container. By now, there is no way the container ship effectively, so I change the container stack to frankenphp, that it technically plays nice with laravel.

Aside from the installer somewhat slow, I think the performance is pretty sleek

Cheers

### DIFF
--- a/docker/Caddyfile
+++ b/docker/Caddyfile
@@ -15,10 +15,7 @@
 
 	file_server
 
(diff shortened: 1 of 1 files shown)


lancepioch:
Hi, please open an GitHub issue so we can have a discussion around this.

For example you have `dunglas/frankenphp-dev` instead of `dunglas/frankenphp`

PR #2203, opened 2026-02-08
Change file browser sticky header again...

hUwUtao:
Seems like the related component has shifted after sometimes


JoanFo3123:
> Seems like the related component has shifted after sometimes

Hey! So there is a plugin for filament to do the same, I suggest you check it out, as perhaps you can find a better way

lancepioch:
@hUwUtao can you add a before/after screenshot? Thanks!

hUwUtao:
sure
image
image

PR #2204, opened 2026-02-08
Progressive Web App

JoanFo1456:
This PR uses (with Reeves consent) his plugin's code, merged on the panel codebase, along with a tiny change for notifications.
PWA was suggested some time ago, and now it's a reality. Thanks, Reeves, for the help.

### DIFF
--- a/app/Filament/Admin/Pages/Settings.php
+++ b/app/Filament/Admin/Pages/Settings.php
@@ -892,8 +892,8 @@ class Settings extends Page implements HasSchemas
                         ->onColor('success')
                         ->offColor('danger')
                         ->live()
-                        ->formatStateUsing(fn ($state): bool => (bool) $state)
-                        ->default((bool) $pwa->get('push_enabled', false)),
+                        ->formatStateUsing(fn ($state): bool => filter_var($state, FILTER_VALIDATE_BOOLEAN))
+                        ->default(filter_var($pwa->get('push_enabled', false), FILTER_VALIDATE_BOOLEAN)),
                     Toggle::make('pwa_push_send_on_database_notifications')
                         ->label(trans('pwa.fields.push_send_on_db.label'))
                         ->helperText(trans('pwa.fields.push_send_on_db.helper'))
@@ -903,8 +903,8 @@ class Settings extends Page implements HasSchemas
                         ->onColor('success')
                         ->offColor('danger')
                         ->visible(fn (Get $get) => $get('pwa_push_enabled'))
-                        ->formatStateUsing(fn ($state): bool => (bool) $state)
-                        ->default((bool) $pwa->get('push_send_on_database_notifications', true)),
+                        ->formatStateUsing(fn ($state): bool => filter_var($state, FILTER_VALIDATE_BOOLEAN))
+                        ->default(filter_var($pwa->get('push_send_on_database_notifications', true), FILTER_VALIDATE_BOOLEAN)),
                     Toggle::make('pwa_push_send_on_mail_notifications')
                         ->label(trans('pwa.fields.push_send_on_mail.label'))
                         ->helperText(trans('pwa.fields.push_send_on_mail.helper'))
@@ -914,8 +914,8 @@ class Settings extends Page implements HasSchemas
                         ->onColor('success')
                         ->offColor('danger')
                         ->visible(fn (Get $get) => $get('pwa_push_enabled'))
-                        ->formatStateUsing(fn ($state): bool => (bool) $state)
-                        ->default((bool) $pwa->get('push_send_on_mail_notifications', false)),
+                        ->formatStateUsing(fn ($state): bool => filter_var($state, FILTER_VALIDATE_BOOLEAN))
+                        ->default(filter_var($pwa->get('push_send_on_mail_notifications', false), FILTER_VALIDATE_BOOLEAN)),
                     TextInput::make('pwa_vapid_subject')
                         ->label(trans('pwa.fields.vapid_subject.label'))
                         ->helperText(trans('pwa.fields.vapid_subject.helper'))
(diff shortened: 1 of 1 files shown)


Boy132:
I think this should stay as plugin.

JoanFo1456:
> I think this should stay as plugin.

It was spoken some time ago on the discord that it should be on the panel, I guess something to toggle it would make it better? As I feel it’s really something all panels should have shipped.

app/Services/Pwa/PwaSettingsRepository.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function allWithDefaults(array $defaults): array
+    {
+        $settings = [];
+
+        foreach ($defaults as $key => $default) {
+            $settings[$key] = $this->get($key, $default);
+        }
+
+        return $settings;
```
lancepioch: ### Architecture
- [ ] Fix the race condition in `ensureVapidKeys()` — concurrent requests can each generate different VAPID key pairs, leaving mismatched keys. Use a DB lock or transaction with a check.

app/Filament/Admin/Pages/Settings.php
```diff
@@ -803,6 +811,214 @@ private function miscSettings(): array
         ];
     }
 
+    /**
+     * @return Component[]
+     */
+    private function pwaSettings(): array
+    {
+        $pwa = app(PwaSettingsRepository::class); // @phpstan-ignore myCustomRules.forbiddenGlobalFunctions
+        $pwa->ensureVapidKeys();
```
lancepioch: ### Architecture
- [ ] Stop calling `ensureVapidKeys()` on settings page render — this is a write operation (generates + persists keys) triggered as a side effect of building the form schema. Move it to an explicit action or defer to save time.
- [ ] Deduplicate test push logic — the VAPID-check → fetch-subscription → build-payload → send flow is copy-pasted between `Settings.php` and `PwaPushController.php`. Extract it into a shared method on `PwaPushService`.

app/Listeners/SendPwaPushOnDatabaseNotification.php
```diff
@@ -0,0 +1,148 @@
...
+
+namespace App\Listeners;
+
+use App\Jobs\SendPwaPush;
+use App\Models\PwaPushSubscription;
+use App\Services\Pwa\PwaSettingsRepository;
+use Illuminate\Notifications\Events\NotificationSent;
+use Illuminate\Notifications\Messages\MailMessage;
+
+class SendPwaPushOnDatabaseNotification
```
lancepioch: ### Correctness
- [ ] Rename `SendPwaPushOnDatabaseNotification` — the class also handles `mail` channel events, making the name misleading. Something like `SendPwaPushOnNotification` would be accurate.

app/Listeners/SendPwaPushOnDatabaseNotification.php
```diff
@@ -0,0 +1,148 @@
...
+            $mail = $event->notification->toMail($event->notifiable);
+            if ($mail instanceof MailMessage) {
+                return $this->normalizePayload($this->payloadFromMailMessage($mail));
+            }
+        }
+
+        return null;
+    }
+
+    /** @return array<string, mixed> */
```
lancepioch: ### Correctness
- [ ] Avoid calling `toMail()` a second time in the listener — the fallback in `buildPayload` re-invokes `toMail()` which may have side effects. Use the already-sent mail data or require `toPwaPush()` for meaningful push payloads.

app/Services/Pwa/PwaActions.php
```diff
@@ -0,0 +1,82 @@
...
+
+                SchemaActions::make([
+                    Action::make('test')
+                        ->label(trans('pwa.actions.test_push'))
+                        ->icon('heroicon-o-paper-airplane')
+                        ->color('warning')
+                        ->visible(fn () => (bool) app(PwaSettingsRepository::class)->get('push_enabled', false)) // @phpstan-ignore myCustomRules.forbiddenGlobalFunctions
+                        ->action(fn () => Notification::make()->title(trans('pwa.notifications.test_sent'))->success()->send())
+                        ->extraAttributes(['onclick' => <<<'JS'
+                            const btn = event.target;
```
lancepioch: ### Correctness
- [ ] Fix `event.target` vs `event.currentTarget` — clicking a child element (like the icon inside the button) will reference the wrong element. Use `event.currentTarget`.

database/migrations/2026_02_08_000000_create_pwa_settings_and_subscriptions.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        if (!Schema::hasTable('pwa_settings')) {
```
lancepioch: ### Correctness
- [ ] Remove `hasTable` guards from the migration — Laravel's migration system already tracks what has run. These guards mask schema drift and partial migration failures.

app/Http/Controllers/PwaPushController.php
```diff
@@ -0,0 +1,125 @@
...
+            ->where('endpoint_hash', hash('sha256', $request->string('endpoint')->toString()))
+            ->where('notifiable_type', $user->getMorphClass())
+            ->where('notifiable_id', $user->getKey())
+            ->delete();
+
+        return response()->json([
+            'message' => trans('pwa.notifications.unsubscribed'),
+        ]);
+    }
+
```
lancepioch: ### Security
- [ ] Escape values interpolated into the service worker JS — while `json_encode` is used for some replacements, audit all interpolated values to ensure no injection is possible through admin-controlled settings.

NerdsCorp:
Maybe make it a Vote on discord to see if people would actually use it? I made the plugin because i don't think it needs to be an APP cause it works well in the browser but there was some people asking about it. Either way it doesn't matter to me, if the decision is to not then ill continue to work on the plugin. Also no need to rush, we can always revisit it in the future. lol

lancepioch:
My honest opinion is that most people won't use this and the amount of work to keep it maintained is far too great, especially considering there are zero tests in this PR. Also I don't believe there's even an enable/disable switch either.

I think that a Plugin offers the perfect compromise, but thank you for the hard work regardless.

In the future, and this is to everyone, please open a discussion (on GitHub or Discord) with the Pelican team so that we evaluate whether we'd accept a new feature or not before any work is made on it.

**It will be rare that a pull request gets accepted and merged without any prior issue or discussion**

PR #2223, opened 2026-02-12
Allow custom ips to still be entered

### LINKED ISSUE #2217: Semi-regression, unable to add custom ip for allocations
### Current Behavior

When adding an allocation only host ip addresses can be added. With https://github.com/pelican-dev/wings/pull/49 merged, it makes it impossible to create allocations for macvlan use via the web ui.

### Expected Behavior

Return to previous behaviour where any IP could be added or add toggle whether to allow any IP address when creating an allocation. Not sure if it was this PR or one earlier than that https://github.com/pelican-dev/panel/pull/800 where it was changed from a text input to a select.

### Steps to Reproduce

Try to add an allocation with a non-host IP address.

### Panel Version

canary

### Wings Version

v1.0.0-beta23

_No response_

### Docker Image

_No response_

### Error Logs

```bash

```

lancepioch:
Fixes #2217

### DIFF
--- a/app/Filament/Admin/Resources/Nodes/RelationManagers/AllocationsRelationManager.php
+++ b/app/Filament/Admin/Resources/Nodes/RelationManagers/AllocationsRelationManager.php
@@ -91,7 +91,10 @@ class AllocationsRelationManager extends RelationManager
                     ->icon(TablerIcon::WorldPlus)
                     ->schema(fn () => [
                         Select::make('allocation_ip')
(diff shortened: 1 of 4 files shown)


QuintenQVD0:
Why this now? we added logic in the wings for it to return extra ip's?

lancepioch:
> Why this now? we added logic in the wings for it to return extra ip's?

Some ip addresses are bound after the fact.

PR #2251, opened 2026-02-18
Root Admin Assignment via WebUI

PaulaBras:
Added admins to assign "Root admin" Role to other users and prevent removing it on their user.
I've missed this feature, and it seems like it was intentionally this way. If not please add this to the beta/release.


notAreYouScared:
This was 100% intentional, and we see no reason for it to be changed. Youre more than welcome to create a role and assign it all permissions.

PR #2282, opened 2026-03-25
Add sortable feature to name column in ListServers

julianbenz:
Possibility to sort the server in the list after name

### DIFF
--- a/app/Filament/App/Resources/Servers/Pages/ListServers.php
+++ b/app/Filament/App/Resources/Servers/Pages/ListServers.php
@@ -81,7 +81,8 @@ class ListServers extends ListRecords
                 ->label(trans('server/dashboard.title'))
                 ->description(fn (Server $server) => $server->description)
                 ->grow()
-                ->searchable(),
+                ->searchable()
+                ->sortable(),
             TextColumn::make('allocation.address')
                 ->label('')
                 ->badge()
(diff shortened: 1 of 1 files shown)


julianbenz:

app/Filament/App/Resources/Servers/Pages/ListServers.php
```diff
@@ -122,6 +123,7 @@ public function table(Table $table): Table
             ->paginated($usingGrid ? [10, 20, 30, 40] : [10, 20, 50, 100])
             ->defaultPaginationPageOption($usingGrid ? 10 : 20)
             ->query(fn () => $baseQuery)
+            ->defaultSort('name', 'asc')
```
Boy132: I don't think the default sort should be by name.

Boy132 (changes requested):

app/Filament/App/Resources/Servers/Pages/ListServers.php
julianbenz: Okay, then I change the default one, but which collum should be the default one?

Boy132: The _default_ default one, `created_at`.
Simply removing the `->defaultSort` line will do.

julianbenz: okay

julianbenz: Done..

Boy132 (approved):

PR #1685, opened 2025-09-09
Fix mixed content issues when running behind HTTPS reverse proxy

PaulaBras:
TL;DR PR fixes the HTTPS reverse proxy issue. For your scenario (HTTPS outside, HTTP inside container)

## Problem

When deploying Pelican Panel behind an HTTPS reverse proxy (nginx, Apache, Cloudflare, etc.), the application generates HTTP asset URLs instead of HTTPS, causing browsers to block mixed content. This results in:

- **Mixed Content Errors**: `Mixed Content: Blocked loading mixed active content "http://example.com/css/app.css"`
- **JavaScript Runtime Errors**: `TypeError: can't access property "then", a.default.detectStore(...) is undefined`
- **Broken Filament Components**: Alpine.js expressions fail with `table is not defined` and `selectFormComponent is not defined`
- **Failed Livewire Hydration**: Components don't initialize properly due to blocked assets

## Root Cause

Laravel wasn't detecting the original HTTPS protocol from reverse proxy headers because:

1. **Missing TrustProxies middleware** - Laravel couldn't read `X-Forwarded-Proto` headers
2. **Insufficient HTTPS forcing** - Only worked when `APP_URL` explicitly started with `https://`
3. **No proxy header detection** - Production deployments behind proxies weren't handled

### 1. Added TrustProxies Middleware

Created `app/Http/Middleware/TrustProxies.php` that:
- Trusts all standard forwarded headers (`X-Forwarded-Proto`, `X-Forwarded-Host`, etc.)
- Supports configurable proxy trust via `TRUST_PROXIES` environment variable
- Registered in application bootstrap for all requests

### 2. Enhanced HTTPS Detection

Modified `AppServiceProvider.php` to force HTTPS scheme when:
- `APP_URL` starts with `https://` (existing behavior)
- **OR** running in production with `X-Forwarded-Proto: https` header (new)
- **OR** `X-Forwarded-SSL: on` header is present
- **OR** request is detected as secure

### 3. Secure Session Configuration

Added secure cookie settings to prevent session hijacking:
- `SESSION_SECURE_COOKIE` - ensures cookies only sent over HTTPS
- `SESSION_SAME_SITE_COOKIE` - CSRF protection with `lax` default

### 4. Updated Environment Defaults

Changed `.env.example` to be HTTPS-friendly by default:
```env
APP_URL=https://example.com
TRUST_PROXIES=*
SESSION_SECURE_COOKIE=true
```

### 5. Added Documentation

[...]

PR #2319, opened 2026-05-06
Fix trailing slash of untold torment and suffering

karaolidis:
When Pelican is deployed behind a reverse proxy (e.g., Traefik, nginx) that terminates TLS and forwards plain HTTP to Apache, navigating to the file editor triggers a mixed content block in the browser.

Why? `EditFiles::getUrl()` appends a trailing `/` to the URL. The `.htaccess` rewrite rule then strips it with a 301 Redirect. When Apache constructs the redirect URL, it uses the scheme of the incoming connection (plain `http://`), not the external-facing scheme. The browser sees a redirect from `https://` to `http://` and blocks it as mixed active content.

The fix simply removes the trailing slash concatenation. The redirect was always a no-op from the user's perspective (append slash, immediately redirect to remove it), and Laravel's front controller handles routing regardless of trailing slashes.

I have run out of hairs to pull out debugging this.

Possibly relevant to #1685.

### DIFF
--- a/app/Filament/Server/Resources/Files/Pages/EditFiles.php
+++ b/app/Filament/Server/Resources/Files/Pages/EditFiles.php
@@ -261,7 +261,7 @@ class EditFiles extends Page
 
     public static function getUrl(array $parameters = [], bool $isAbsolute = true, ?string $panel = null, ?Model $tenant = null, bool $shouldGuessMissingParameters = false, ?string $configuration = null): string
     {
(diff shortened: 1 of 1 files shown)


karaolidis:
> getUrl() override is now a no-op — consider removing it entirely.

I'll leave this one up to the maintainers.

edit: i.e. the decision, not the cleanup

QuintenQVD0:
It is needed else you can't edit .env files or .php files for example.

rmartinoscar (changes requested):
Hey thanks for your contribution but im afraid without that you won't be able to edit files that start with a dot or are processed by your webserver such as .html, .php etc...

PR #2321, opened 2026-05-06
Allow Caddy to bind privileged ports as www-data

Binero:
This adds the `cap_net_bind_service` file capability to the Caddy binary, letting it bind `:80`/`:443` as the unprivileged `www-data` user the image already drops to.

```dockerfile
RUN apk add --no-cache --virtual .setcap-deps libcap \
    && setcap cap_net_bind_service+ep /usr/sbin/caddy \
    && apk del .setcap-deps
```

This would allow us to listen on port 443 when using host networking. 

The reason to want host networking is that bridge mode obscures real client IPs and doesn't do public IPv6 without significant Docker daemon reconfiguration. The former additionally means that in access logs rather than the real user IP showing up, it will show Docker's internal virtual IP address, which is the main reason I want to pitch this PR. 

Another way to achieve the same is to run the entire container as root, but that would defeat the existing `USER www-data` directive, and basically boils down to deploying a nuke to squash a mosquito. 

People who are using bridge networking are unaffected by this change; they already have privilege over `:80`/`:443` by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,11 @@ RUN apk add --no-cache \
     # required for installing plugins. Pulled from https://github.com/pelican-dev/panel/pull/2034
     zip unzip 7zip bzip2-dev yarn git
 
+# Allow Caddy to bind privileged ports (80/443) as www-data, including under network_mode: host
+RUN apk add --no-cache --virtual .setcap-deps libcap \
+    && setcap cap_net_bind_service+ep /usr/sbin/caddy \
+    && apk del .setcap-deps
+
 # Copy composer binary for runtime plugin dependency management
 COPY --from=composer /usr/local/bin/composer /usr/local/bin/composer
 COPY --chown=root:www-data --chmod=770 --from=composerbuild /build .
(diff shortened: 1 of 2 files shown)


Binero:

rmartinoscar:
Hey thanks for your contribution, I don't think we should do it this way the current docker and log config to get the proper client ip header seems fine but we'll see what @parkervcp has to say about it, he is the one mainly managing docker stuff.

mristau:
in my caddy logs i have the correct client_ip, just remote_ip is the one my proxy is on, that one is always my docker network gateway

Binero:
I was surprised by your seemingly different results, but I figured it out, and the answer is somewhere in the middle. The `client_ip` does show up correctly for external connections. I tested it from localhost, using the localhost address, and ordinarily you would expect localhost to be the client IP, but in this case, it does incorrectly show the IP address of the Docker proxy instead. Connecting from the surprisingly outside doesn't seem to have this issue. 

So even though the main reason I gave why someone might not want to run over the bridge network might be more limited as I first understood it to be, there are many reasons why one might want to run over the host network (e.g. cleaner IPv6 support), and this small change would support those setups without requiring people to undo some of the security hardening the image already has built-in.

The addition to the Dockerfile seems to be self-contained enough to not pose any significant maintenance burden, while still opening the door for this common type of setup.

PR #2209, opened 2026-02-09
fix: composer content hash

Hythera:
There seems to be an issue with the composer hash for me when trying to build a nix derivation. Updating the composer content hash (`composer update --lock --no-install`) seems to resolve that issue.
This seems to have happened before (see #533).

Issue #533, opened 2024-08-07
composer.lock outdated


mark-boute:
### Current Behavior

When following the Get Started documentation and running `composer install` I get the following output:

`Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run composer update or composer update <package name>.`

This does not matter when building on Ubuntu, but I am trying to build a NixOS derivation using `php83.buildComposerProject`. That requires the .lock file to be up-to-date unless I set the `composerStrictValidation = true;` option, which is not recommended for published packages.

### Expected Behavior

I expect the composer.json and .lock files to be in sync.

### Steps to Reproduce

Follow the Getting Started guide

### Panel Version

v1.0.0-beta7

### Wings Version

n.a.

### Error Logs

```bash
pelican@boisserver:/var/www/pelican$ composer validate
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
# Lock file errors
- The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update` or `composer update <package name>`.
# General warnings
- No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues before opening this issue.
- [X] I have provided all relevant details, including the specific game and Docker images I am using if this issue is related to running a server.
- [X] I have checked in the Discord server and believe this is a bug with the software, and not a configuration issue with my specific system.

Silver-Golden:
@mark-boute by any chance did ye get it to work nicely on nixos?   
If so do ye have a config ye can share?

mark-boute:
> @mark-boute by any chance did ye get it to work nicely on nixos?   
> If so do ye have a config ye can share?

Semi-working. 
We managed to get it to build by changing a setting to not evaluate the lockfile, but decided to install the panel manually as it's essentially just a webpage.
We did get wings working. I think the repo is now private, so I'll check for relevant code real quick.

mark-boute:
```
{ inputs, pkgs, lib, config, ... }:

with lib; let
  cfg = config.modules.pelican-panel;
  dir = "/var/www/pelican";
in {
  options = {
    modules.pelican-panel = {
      enable = mkEnableOption "Pelican Panel";
    };
  };

  config = mkIf cfg.enable {
    environment.systemPackages = [
      pkgs.curl
      pkgs.gnutar
      pkgs.unzip
[...]

PR #2324, opened 2026-05-08
fix: composer content hash

Hythera:
Same reason as in #2209.

> There seems to be an issue with the composer hash for me when trying to build a nix derivation. Updating the composer content hash (`composer update --lock --no-install`) seems to resolve that issue.
This seems to have happened before (see #533).


Boy132 (approved):

notAreYouScared:
thanks for the PR, the hash will be fixed before the next beta. I always do a composer update to update Filament + packages before a release, Just didn't happen this time :/

PR #2362, opened 2026-05-27
Replace hardcoded `pelican` with config value

Boy132:


resources/views/filament/layouts/footer.blade.php
```diff
@@ -2,7 +2,7 @@
     {{ \Filament\Support\Facades\FilamentView::renderHook(\App\Enums\CustomRenderHooks::FooterStart->value) }}
 
     <a class="font-semibold" href="https://pelican.dev/docs/#core-team" target="_blank">
-        &copy; {{ date('Y') }} Pelican
+        &copy; {{ date('Y') }} {{ $appName }}
```
lancepioch, line 5: Revert this or feel free to change it to $appName powered by &copy; 2026 Pelican like above.

resources/views/filament/components/server-console.blade.php
```diff
@@ -4,6 +4,8 @@
         $userFont = (string) user()?->getCustomization(\App\Enums\CustomizationKey::ConsoleFont);
         $userFontSize = (int) user()?->getCustomization(\App\Enums\CustomizationKey::ConsoleFontSize);
         $userRows = (int) user()?->getCustomization(\App\Enums\CustomizationKey::ConsoleRows);
+
+        $terminalPrelude = str(config('app.name'))->slug()->lower()->toString();
```
lancepioch, line 8: I can still see people messing this up. I believe you'll also need to strip out certain special characters too. Otherwise revert this.

lancepioch (changes requested):

resources/views/filament/layouts/footer.blade.php
Boy132: If the app name isn't `Pelican` it will automatically add `Powered by Pelican` to it.

resources/views/filament/components/server-console.blade.php
Boy132: `slug` should remove all illegal characters.

lancepioch (approved):

PR #2371, opened 2026-06-03
Skip daemon RPC calls for servers on maintenance-mode nodes

hUwUtao:
[**Discussion: Node disruption causes UX delay**](https://github.com/pelican-dev/panel/discussions/2369)

**Purpose**: Practical way to handle Wings node down, yet. Due to the nature of "Maintenance mode" option, runtime consequences must be known by operator, which should make it useful. This option will disengage most of the query process from the dashboard to the node in order to let the Wings instance in a safe state.

### What changed:

- `retrieveStatus()`: return Missing immediately if node is under maintenance
- `retrieveResources()`: return empty if server status is not starting/running, which includes Missing (from maintenance guard) and all stopped states

### DIFF
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -457,6 +457,10 @@ class Server extends Model implements HasAvatar, Validatable
 
     public function retrieveStatus(): ContainerStatus
     {
+        if ($this->node->isUnderMaintenance()) {
+            return ContainerStatus::Missing;
+        }
+
         return cache()->remember("servers.$this->uuid.status", now()->addSeconds(15), function () {
             // @phpstan-ignore myCustomRules.forbiddenGlobalFunctions
             $details = app(DaemonServerRepository::class)->setServer($this)->getDetails();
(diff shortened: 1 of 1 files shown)


lancepioch (approved):

Boy132:
I don't know if this is desirable. When a node is under maintenance this would restrict what admins can do with servers.

hUwUtao:
> I don't know if this is desirable. When a node is under maintenance this would restrict what admins can do with servers.

As far as this do, this will only disrupt report of statistics and status.

For actual actions block, it would be introduced in `Server::isInConflictState()`

app/Models/Server.php
```diff
@@ -470,6 +474,10 @@ public function retrieveStatus(): ContainerStatus
      */
     public function retrieveResources(): array
     {
+        if (!$this->retrieveStatus()->isStartingOrRunning()) {
```
Boy132: Resources should still be retrieved when stopping.

Boy132 (changes requested):
You're right. This is probably fine.

Just one small nitpick.

app/Models/Server.php
hUwUtao: oh sure is

hUwUtao:
My fault, I rebased main onto this branch, should be fine now.

Boy132 (approved):

PR #2372, opened 2026-06-03
Fix sed command syntax in docker-publish.yml

hUwUtao:
to avoid conflict with release tag format: "rel/tag/something"


parkervcp (approved):
LGTM

PR #2373, opened 2026-06-03
Fix dark mode

hUwUtao:
Recently, dark mode left some component in light theme, even with correct class applied. Retested on both theme.

- @variant → @custom-variant to decouple from built-in media query variant
- Drop :where() so .dark .dark\:bg-gray-900 has specificity (0,0,2,0) beating bg-white at (0,0,1,0)
- Remove redundant bg-white from server-small-data-block
- Add dark mode classes to api-index page

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -26,4 +26,4 @@
 
 @source '../../vendor/gboquizosanchez/filament-log-viewer/resources/views/**/*.blade.php';
 
-@variant dark (&:where(.dark, .dark *));
+@custom-variant dark (.dark &);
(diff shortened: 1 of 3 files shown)


notAreYouScared:
I dont think this is the correct fix. The small data block has a dark:bg class thats clearly not working.

hUwUtao:
> I dont think this is the correct fix. The small data block has a dark:bg class thats clearly not working.

Clueless as I am, i tried to look at the style priority in chrome dev, trying to reorder class but no avail, though might be wrong in some aspect but it satisfied the way fillament behave

image

image

notAreYouScared:
The issue appears to only be on main, I have a suspicion that the laravel pr broke it. 

Do a composer update, build assets and see if its fixed.

notAreYouScared:
Well, I see it fixes it. Odd how it just broke.... I blame filament

notAreYouScared (approved):

PR #2378, opened 2026-06-06
docker: parse TRUSTED_PROXIES from .env

harryyoud:
if we don't lift this out, the CADDY_TRUSTED_PROXIES variable will be left empty, and the generated Caddyfile will have not have trusted_proxies set


mristau:
default way as in compose-full-stack.yml is to add TRUSTED_PROXIES into your compose file directly, it will be automatically added into .env and visible in the panel, just editing, you would need to add it yourself

lancepioch (approved):

PR #2379, opened 2026-06-07
add p:egg:normalize command

harryyoud:
used to upgrade and normalize egg files for programmatic upgrading of eggs

### DIFF
--- a/app/Console/Commands/Egg/NormalizeEggCommand.php
+++ b/app/Console/Commands/Egg/NormalizeEggCommand.php
@@ -1,0 +1,152 @@
+<?php
+
+namespace App\Console\Commands\Egg;
+
+use App\Enums\EggFormat;
+use App\Exceptions\Service\InvalidFileUploadException;
+use App\Models\Egg;
+use App\Services\Eggs\Sharing\EggExporterService;
+use App\Services\Eggs\Sharing\EggImporterService;
+use Carbon\Carbon;
+use Illuminate\Console\Command;
+use Symfony\Component\Yaml\Yaml;
+
+class NormalizeEggCommand extends Command
+{
+    protected $description = 'Upgrades egg json/yaml to latest format as yaml';
+
+    protected $signature = 'p:egg:normalize {file} {--delete-original}';
+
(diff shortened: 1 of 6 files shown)


harryyoud:
I disagree with CodeRabbitAI above regarding this comment:

======================

25-29: ⚡ Quick win

Consider checking file existence before reading.

While the code correctly handles the false return from file_get_contents, it will still emit a PHP warning if the file doesn't exist. Adding an explicit file_exists() or is_readable() check before line 25 would provide clearer error messaging and avoid warnings in logs.

======================

I think this would just introduce a TOCTOU, and it's easier to ask for forgiveness than permission in this scenario. We don't need elegant or graceful error handling in console as long as we are safe and bail out on error and leave a sensible error message.

Boy132:
I don't get why this command is needed.

harryyoud:
When the panel checks for updates to the egg, it compares the local egg as json to the remote egg as json. If the remote egg has an older version (e.g. PTDL_V2) or array items in a different order, it will appear out of date every single time.

Right now, the Minecraft eggs repo as an example has a large number of eggs in this way - and so when the check egg cron job is fixed, many of these eggs will always show that they need an update.

I thought to enable consistency, I created this command to enable automation of upgrading/normalizing egg files in the repos. This could potentially be used for CI on the egg repos to ensure the files are consistently formatted and don't constantly trigger an out of date state on the panel.

See https://github.com/pelican-eggs/minecraft/pull/133 as an example of this in action

lancepioch:
Thanks for the contribution! After re-reading it, I believe it would be good to have this. However, can you make the following adjustments first?

* [x] Add at least one fixture test for an old PTDL_v2 egg becoming PLCN_v3 YAML with upgraded variable/config paths.
* [x] Ensure `--delete-original` deletes only after a successful write.
* [x] Guard missing or non-string `meta`.`update_url`.
* [ ] Make normalization idempotent, especially around `exported_at` (eg. add test(s) proving running normalizer twice produces the same output).

lancepioch (approved):

PR #2388, opened 2026-06-11
add per-node wings app_name configuration

rxtted:
the panel never sends `app_name` to a node's wings configuration, so the daemon console prefix (`[<name> Daemon]:`) is stuck on whatever wings defaults to (`pelican`) unless you hand-edit the wings config directly. the terminal prompt prelude (`pelican@server`) is slugged off the global [`APP_NAME`,] so there's no wings or node level control of that either. a node's console identity is basically not configurable from the panel.

started this as a straightforward fix: just send app_name down in the node configuration payload so wings picks it up automatically. that alone would make the daemon prefix track the panel name without any manual config.yml editing.

but obivously things r never that simple lmao. i added create and edit fields for it, because there are reasonable cases where you'd want a node's wings app_name to differ from the panel's, things like per-node branding or multi-tenant setups. keeping it a free field rather than just mirroring `APP_NAME` seemed more useful.

once that was in place, the terminal prompt (`pelican@server`) was still permanently fixed on the app_name, so i wired the prompt prelude to pull from the same wings var. that gets both console surfaces configurable at the node level and decouples them from the panel's app_name entirely.

blank or whitespace-only input is stored as null, which falls back to `config('app.name')`.

to fit the field cleanly into the advanced settings section, i reworked that row into a small grid and re-spanned the deploy and maintenance toggles. it was the only way i could make the new field work cleanly

Screenshot_20260611_200420

Screenshot_20260611_200522


lancepioch:
These are a lot of changes for a problem that doesn't have any meaningful impact. Please feel free to create a new GitHub discussion about this, thank you again for the contribution.

PR #2389, opened 2026-06-12
Fix webhook payload serialization and null headers

lancepioch:


JoanFo1456:
Hey! This is fixed already with my pull request, I'll review your changes this weekend.

PR #2422, opened 2026-06-28
feat: support schedule cron input in user timezone (#1228)

### LINKED ISSUE #1228: Schedule cron input in user timezone
Currently the cron input for schedules is always UTC. Ideally it should be displayed to the user timezone on the frontend so users don't have manually calculate the correct timezone.

wu-zuan:
## Description
This PR addresses issue #1228 by allowing users to define and display server schedules using a custom timezone instead of forcing UTC globally on the frontend. 

## Key Changes
- **Database Schema**: Added a nullable `timezone` column to the `schedules` table.
- **Form UI**: Introduced a `timezone` dropdown (defaulting to the user's timezone) in the Filament schedule form.
- **Cron evaluation**: Updated the scheduler calculation helper (`Utilities::getScheduleNextRunDate`) to evaluate the next run date relative to the configured timezone, while storing the resulting `next_run_at` timestamp in UTC in the backend database.
- **API & Exporter**: Handled timezone field in API requests (`StoreScheduleRequest`/`ScheduleController`), fractal transformers (`ScheduleTransformer`), and import/export service layers.
- **Localization**: Updated key translations (`en`, `zh_TW`) for cron input timezone guidance.

Fixes #1228

### DIFF
--- a/app/Filament/Server/Resources/Schedules/Pages/CreateSchedule.php
+++ b/app/Filament/Server/Resources/Schedules/Pages/CreateSchedule.php
@@ -57,7 +57,8 @@ class CreateSchedule extends CreateRecord
                 $data['cron_hour'],
                 $data['cron_day_of_month'],
                 $data['cron_month'],
-                $data['cron_day_of_week']
+                $data['cron_day_of_week'],
+                $data['timezone'] ?? null
             );
         }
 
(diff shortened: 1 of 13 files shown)


Boy132:
Thanks but this is not a desired solution as it just works around the user timezone.

PR #2424, opened 2026-06-29
Add IO as a setting.

JoanFo1456:
I understand that hardcoded I/O is on purpose for what I can remember. But the other day I had a docker that did not like I/O and I had to edit core files to make it 0. Would be nice to have it as a feature actually and not a hardcoded 500.
On this instance, I am not sure where to place it, so I accept requests to move it.
IO explanation was copied from old io component.


lancepioch:
It is hidden on purpose because the majority of people don't/won't understand it and use it wrong.

It's only accessible through the api for that reasoning.

PR #2429, opened 2026-07-02
Fix activity log event filter query and add missing timestamp index

execsuroot:
This PR resolves issue where activity page timeouts (or loads for a long time) for servers with lots of activity.

Two small fixes in the activity log area:
- ActivityResource: event filter dropdown was pulling every row into memory and deduping/sorting in PHP. Switched to SELECT DISTINCT ... ORDER BY so the DB does the work. Same output, just faster on tables with a lot of activity.
- Added an index on activity_logs.timestamp — this column gets queried/sorted on constantly and had no index.


execsuroot:

lancepioch (approved):

PR #2431, opened 2026-07-04
fix(ui): align leading icons with text in table columns

JoshuaGerke:
Filament's .fi-ta-text-item applies line-clamp-(--line-clamp,none). Even without a clamp value set, this keeps `display: -webkit-box` with `-webkit-box-orient: vertical`, which stacks a leading icon above the label instead of inline (visible e.g. in the server File Manager list).

This adds a targeted override that restores a horizontal flex layout with a small gap whenever a `.fi-ta-text-item` contains a leading `.fi-icon`, without affecting other line-clamp behavior.

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -27,3 +27,21 @@
 @source '../../vendor/gboquizosanchez/filament-log-viewer/resources/views/**/*.blade.php';
 
 @custom-variant dark (.dark &);
+
+
+/*
+ * Fix: leading icons in table text columns (e.g. file/folder name in the
+ * File Manager) render stacked above the label instead of inline.
+ *
+ * Filament's .fi-ta-text-item applies line-clamp-(--line-clamp,none)
+ * (see vendor/filament/tables resources/css/columns/text.css). Even when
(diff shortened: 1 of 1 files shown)


Boy132:
What do you mean?
image

JoshuaGerke:
@Boy132 
brave_qXfs1kvrGA
it looks like this on our panel (newest update).

JoshuaGerke:
> What do you mean? image

1.0.0-beta35
image

Boy132:
You can see on the demo that the problem is not reproducible: https://demo.pelican.dev/server/ce6ee345/files

So it's most likely something with your panel installation. Or it was a dependency issue and got already resolved on main. The demo runs main.

vanlueckn:
Same issue on the demo:

image

I am using Helium Version 0.14.2.1 (Chromium 150.0.7871.46) (64-Bit), @JoshuaGerke is using Brave.

I also tested it in the latest Edge version.

Boy132:
Seems like it's an issue with chromium: https://github.com/filamentphp/filament/issues/20098

PR #2441, opened 2026-07-08
fix: resolve server-entry grid card resource freeze

### LINKED ISSUE #2427: Dashboard grid/card layout: server resource stats (CPU/RAM/uptime) freeze and never update
### Current Behavior

  On the client dashboard (App panel ListServers) in grid/card layout, each server card's CPU/RAM/uptime freezes at its first-rendered value and never updates. Switching the user's dashboard_layout to table fixes it — columns update every 15s as expected.

  Root cause (from inspection):
  - Grid cards render via ServerEntryColumn → @livewire('server-entry', ['server' => $server, 'lazy' => true], key($server->id)), and
  server-entry.blade.php has its own wire:poll.15s.
  - The parent ListServers table also polls (->poll('15s')). On each parent poll, the child server-entry is re-emitted as an empty
  placeholder (<div wire:name="server-entry" ...></div> with no rendered html in the Livewire response), resetting the lazy child
  before its own poll completes.
  - Net effect: the nested child never advances past its mount value. Table layout renders the stats as direct parent-table columns,
  so the parent poll refreshes them correctly.

### Expected Behavior

Card stats refresh on the 15s poll, same as table layout.

### Steps to Reproduce

  1. Set dashboard layout to Grid (cards).
  2. Open the dashboard for a running server.
  3. Observe CPU/RAM/uptime — they stay frozen at the initial value indefinitely (page reload shows a new frozen value).

### Panel Version

1.0.0-beta35

### Wings Version

v1.0.0-beta26

Minecraft (Paper)

### Docker Image

ghcr.io/pelican-eggs/yolks:java_25

### Error Logs

```bash

```

apple050620312:
Closes #2427 

#### What does this PR do?
This PR fixes an issue where the Server CPU, RAM, and Uptime statistics freeze on their initially loaded values when the dashboard is viewed in "Grid" layout (Cards). 

The root cause was a conflict between the Livewire `ListServers` table polling and the nested lazy-loaded `server-entry` component. Every 15 seconds, the parent table would trigger a refresh. Because the nested component was set to `lazy => true`, the parent refresh would reset the child component back to its empty placeholder state, interrupting its own internal `wire:poll.15s`. Consequently, the component never fully advanced past its original mount state.

This fix properly resolves the reactivity conflict by utilizing Livewire v3's `#[Reactive]` attribute to push parent data down to the children, rather than relying on the children to independently lazy load and poll.

#### Implementation Details
- Removed the `lazy => true` flag from the `@livewire` directive in `server-entry-column.blade.php`.
- Removed the redundant `wire:poll.15s` from `server-entry.blade.php` (as the parent table already polls every 15s).
- Added the `Livewire\Attributes\Reactive` attribute to the `public Server $server;` property in `App\Livewire\ServerEntry.php` to ensure the child component reacts to the `$server` prop updates driven by the parent component's poll.

#### How to Test
1. Access the Client Panel Dashboard.
2. Ensure your dashboard view layout is set to **Grid**.
3. Observe a running server's CPU, Memory, or Uptime progress bars.
4. Verify that the stats automatically refresh every 15 seconds without remaining permanently frozen.


resources/views/livewire/columns/server-entry-column.blade.php
```diff
@@ -4,5 +4,5 @@
 @endphp
 
 <div class="w-full">
-    @livewire('server-entry', ['server' => $server, 'lazy' => true], key($server->id))
```
Boy132, line 7: Removing lazy will drastically slow down the page load speed. So this is not feasible.

Boy132 (changes requested):

Boy132 (approved):

PR #2461, opened 2026-07-18
Fix docker healthcheck

lancepioch:
The `SKIP_CADDY` check in healthcheck.sh was invalid syntax, so the curl check against `/up` never actually ran and only php-fpm was being checked. Fixed the condition and bumped the check interval so a dead panel gets noticed in seconds instead of minutes.


parkervcp (approved):
LGTM

PR #2462, opened 2026-07-18
Enable opcache and tune PHP in the docker image

lancepioch:
The image was running with no opcache, default php.ini, and the stock fpm pool (max 5 workers). That means every request recompiles the whole framework. This turns opcache on, uses the production php.ini with a 256M memory limit, raises the pool to 15 workers, and caches blade views on boot.


parkervcp (approved):
LGTM

PR #2463, opened 2026-07-18
Make entrypoint .env parsing safer

lancepioch:
The entrypoint grepped .env without anchoring, so a comment line or another variable containing the same name could match and crash the container on boot. Now it only matches real assignments, container env wins over .env, generated keys use the normal `base64:` format, and there's a `SKIP_MIGRATIONS` flag for people running more than one replica.


parkervcp (approved):
LGTM

PR #2464, opened 2026-07-18
Show php-fpm and queue worker output in docker logs

lancepioch:
Only caddy's output made it to `docker logs` - php-fpm and the queue worker were writing to files under /tmp, so app errors and failed jobs were invisible. Route everything to stdout, and give the queue worker time to finish jobs on shutdown instead of killing them after 10 seconds.


parkervcp (approved):
LGTM

PR #2471, opened 2026-07-23
fix(i18n): translate hardcoded interface strings

kazaminosuke:
## Summary

This fixes a structural i18n gap where user-facing Filament and Livewire strings were passed directly to label(), tooltip(), Tab::make(), and related presentation methods instead of translation helpers.

- Translate the Node connectivity status tooltips, including WebSocket failure and no-server states.
- Translate webhook scope labels and use complete translated tab titles for Global and Server webhooks.
- Translate the admin user 2FA method tooltip values.
- Replace additional hardcoded management UI strings and common Captcha/OAuth settings labels with existing or new English translation keys.

## Root cause

Several components bypassed the trans() convention even though their strings are rendered directly in the user interface. This left non-English locales with English-only labels, tooltips, and tab titles.

## Validation

- php -l for all 24 changed PHP and translation files
- git diff --check


kazaminosuke:
CodeRabbit review follow-up: addressed the one actionable inline finding in c09ec3f05. Node connectivity tooltip text is now JSON-encoded before interpolation into the x-tooltip expression, and allowHTML is disabled because translation strings are plain text. Validation: php -l app/Livewire/NodeClientConnectivity.php and git diff --check passed. No other actionable inline or top-level review comments were present.

app/Livewire/NodeClientConnectivity.php
```diff
@@ -64,11 +64,11 @@ public function render(): View
...
-            'offlineIcon' => $this->makeIcon(TablerIcon::WorldX, 'danger', 'Node is not reachable from your browser'),
-            'onlineIcon' => $this->makeIcon(TablerIcon::WorldCheck, 'success', 'Node is reachable'),
-            'warningIcon' => $this->makeIcon(TablerIcon::WorldExclamation, 'warning', 'Node is reachable, but WebSocket failed. Check reverse proxy config.'),
-            'onlineNoWsIcon' => $this->makeIcon(TablerIcon::WorldCheck, 'success', 'Node is reachable (WebSocket not tested — no servers)'),
+            'loadingIcon' => $this->makeIcon(TablerIcon::WorldQuestion, 'warning', trans('admin/node.connectivity.checking')),
+            'offlineIcon' => $this->makeIcon(TablerIcon::WorldX, 'danger', trans('admin/node.connectivity.offline')),
+            'onlineIcon' => $this->makeIcon(TablerIcon::WorldCheck, 'success', trans('admin/node.connectivity.online')),
+            'warningIcon' => $this->makeIcon(TablerIcon::WorldExclamation, 'warning', trans('admin/node.connectivity.websocket_failed')),
+            'onlineNoWsIcon' => $this->makeIcon(TablerIcon::WorldCheck, 'success', trans('admin/node.connectivity.websocket_not_tested')),
```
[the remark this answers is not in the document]
kazaminosuke: Fixed in c09ec3f05. Tooltip content is now encoded with json_encode() and JSON_HEX_* flags before being embedded in the x-tooltip JavaScript expression. Since these translations are plain text, allowHTML is now false.

kazaminosuke:
@coderabbitai review

app/Filament/Admin/Resources/Nodes/Pages/EditNode.php
```diff
@@ -580,7 +580,7 @@ protected function getDefaultTabs(): array
                         ->columnSpanFull()
                         ->state(new HtmlString(trans('admin/node.instructions_help'))),
                     CodeEntry::make('config')
-                        ->label('/etc/pelican/config.yml')
+                        ->label(trans('admin/node.config_path'))
```
Boy132: That's a file path, it doesn't need translation.

Boy132 (changes requested):

app/Filament/Admin/Resources/Nodes/Pages/EditNode.php
kazaminosuke: Thank you for the clarification. You are right: this label is the literal configuration-file path, not localizable UI text. I reverted it to `/etc/pelican/config.yml` and removed the unnecessary translation key in commit 6bc2def76.

Boy132 (approved):

PR #2494, opened 2026-08-05
Auto-generate API docs descriptions from FormRequest rules with Scramble

Fritztechnic:
The generated OpenAPI docs for the Application and Client API were pretty bare: most operations had no description and request bodies just listed the field names without explaining what they are for.

This adds a Scramble extension that fills in the gaps automatically. For operations without much of a description already, it writes a short generic one based on the controller action (index/store/update/ destroy/etc). For request bodies it walks the FormRequest validation rules and adds a one-line explanation per field, pulled from a small lookup table of the field names we actually use (falls back to a generic guess for anything not in the table). The explanations describe what the field means, not the validation constraints, since those are already visible in the schema itself.

Registered the extension both in config/scramble.php (so it loads before Scramble reads its config) and via Scramble::registerExtension() in AppServiceProvider, belt and suspenders because of how Scramble boots its package config.


Fritztechnic:

lancepioch:
Hope you don't mind me pushing a few commits straight to your branch, it felt like less of a mess than leaving a dozen separate review comments that were all circling the same thing.

The big change is that the generated descriptions are gone, mostly because the extension appended a generic paragraph to any description shorter than 220 characters and we already have real ones written by hand on about 135 endpoints, so `GET /servers` ended up with "Return all the servers that currently exist on the Panel" followed by a second paragraph that didn't really say anything. The field descriptions had a similar problem in that they were keyed on the bare field name across the whole API, which meant `user` got described as the server owner even on the subuser endpoints where it means something else entirely.

The nice surprise is that Scramble already handles this natively, since a comment sitting above a rule becomes that field's description:

```php
/** Memory the server may use, in MiB. Use `0` for unlimited. */
'limits.memory' => $rules['memory'],
```

so I went through and did that everywhere, which has the side benefit that the wording lives right next to the rule and can't drift away from the endpoint it belongs to. A handful of requests take their rules straight off the model and have no array to comment on, and those use `#[BodyParameter]` instead, which just merges a description onto whatever Scramble already inferred.

Your Fractal extensions are easily the best part of this PR and they're staying, though I did add `meta.pagination` to the envelope since Fractal attaches the paginator itself and every list endpoint really does return that block, and I put in a test that generates both documents so a Scramble upgrade can't quietly break any of it without CI noticing.

I dropped `config/scramble.php` as well, because registering the extensions in `register()` happens before Scramble's own provider boots and works for every extension type, so the config file wasn't really buying us anything and the hardcoded `vendor/` path in it would break for anyone running a custom composer `vendor-dir`.

While I was in there I noticed that PATCH on users, nodes, database hosts and mounts wasn't documenting a request body at all, for the same reason the extension was silently failing on those requests, so I fixed them too.

Last thing, and this isn't a knock on the work at all, I reverted the docs page redesign purely because it's a separate concern from the description generation and will get a much better read as its own PR.

Fritztechnic:
Nno problem at all. To be honest, I'm really grateful, I wasn't sure if everything would work the way I'd imagined it would, so thanks for taking another look :)

The doc-comment-on-the-rule trick is really clever, didn't even know Scramble picked that up natively, makes way more sense like this. I should have reaf the docs a bit more :/
Also fair on the config/scramble.php thing, I genuinely didn't think about custom vendor-dir setups.
No objection on reverting the docs page redesign either, that was expansion of the scope from me getting distracted by dark mode instead of finishing the actual feature xD, so ill do another PR for it later.

Appreciate you taking the time to review this. I'll take a look at the last couple of CodeRabbit nitpicks to make him happy again.

Fritztechnic:
btw what do you think of a docs window like this? 
grafik

lancepioch:
Pushed two more commits, and a note on a third I pushed and then took back out.

`f10a6a0` fixes the `meta` block. `addMeta()` is variadic over arrays and merges each one into the meta array, so the keys that end up in the response are the keys of every *argument*, while `withMeta()` was reading the argument positions as the keys. That documented `meta` as a tuple holding the block instead of the block itself:

```
POST /account/api-keys   meta => {"type": "array", "prefixItems": [{"secret_token": ...}]}
GET  /servers/{server}/backups   meta => {"pagination": ..., "0": {"backup_count": ...}}
```

All 22 endpoints that call `addMeta()` were affected. The pagination-only listings happened to look right because `buildEnvelope()` adds that block itself, which is exactly why the existing tests didn't catch it, so there's an assertion on the meta block now too.

`1032c94` is just wording: the database name is prefixed with the server ID *and* has to be unique within that server, and saying the prefix is what makes it unique for the server ran the two together.

The commit I dropped is worth writing down somewhere even though it doesn't belong in this PR. `MountController::addEggs()`, `addNodes()` and `addServers()` all `attach()` onto the `mountables` pivot, which has a composite primary key on `(mount_id, mountable_id, mountable_type)`. So assigning an egg, node or server that's already assigned raises a unique constraint violation and the endpoint 500s instead of doing nothing:

```
Illuminate\Database\UniqueConstraintViolationException
SQLSTATE[23000]: UNIQUE constraint failed: mountables.mount_id, mountables.mountable_id, mountables.mountable_type
```

`syncWithoutDetaching()` makes all three idempotent, but that's a behaviour change and this PR is about documentation, so it should go in on its own where it can be reviewed as one. I'll pick it up separately.

That does leave the mount descriptions inconsistent here: eggs now correctly says assignments are additive, while nodes still says omitted entries are "unassigned" and servers says they're "detached", neither of which is what the controller does.

lancepioch (approved):

PR #2502, opened 2026-08-09
Fix: servers per-page setting resets after session expiry

xLaugh:
## Summary

Filament only caches the servers table's "records per page" selection in the PHP session. Once the session expires (Laravel's default `SESSION_LIFETIME` is 120 minutes, and Pelican doesn't override it), the setting silently reverts to the hardcoded default (10 in grid view / 20 in list view) instead of staying at whatever the user picked.

This PR persists the selection in the user's `customization` column (the same mechanism already used for `dashboard_layout`, `top_navigation`, etc.), so it survives session expiry instead of only lasting a couple of hours.

- Session-based caching (Filament's default behavior) is untouched and still takes priority when valid, for instant client-side feedback.
- When the session doesn't have a value (expired, or a new session from a different device), the table now falls back to the value stored in `customization.servers_per_page` instead of the hardcoded default.
- Scoped to the servers list (`ListServers`) only, per the contributing guide's "one targeted change per PR" — the same session-only limitation exists on a few other tables (egg list, activity log, file manager) but I left those out of this PR.

## Test plan

Tested manually end-to-end in a local Docker dev build (`Dockerfile.dev`), driving a real browser against it:

- [x] Fresh user, no stored preference → per-page defaults to 10 (unchanged behavior)
- [x] Changing per-page to 20 persists `customization.servers_per_page = 20` in the database
- [x] Server-side session files deleted (simulating expiry) while still authenticated (remember-me) → per-page still shows 20 on reload, instead of reverting to 10
- No automated tests included yet.

### DIFF
--- a/app/Enums/CustomizationKey.php
+++ b/app/Enums/CustomizationKey.php
@@ -14,6 +14,8 @@ enum CustomizationKey: string
     case ButtonStyle = 'button_style';
     case RedirectToAdmin = 'redirect_to_admin';
 
+    case ServersPerPage = 'servers_per_page';
+
     public function getDefaultValue(): string|int|bool
     {
         return match ($this) {
@@ -25,6 +27,8 @@ enum CustomizationKey: string
             self::DashboardLayout => 'grid',
             self::ButtonStyle => true,
             self::RedirectToAdmin => false,
+            // 0 means "unset", the table falls back to its contextual default (see ListServers).
+            self::ServersPerPage => 0,
         };
     }
 
--- a/app/Filament/App/Resources/Servers/Pages/ListServers.php
+++ b/app/Filament/App/Resources/Servers/Pages/ListServers.php
@@ -113,11 +113,23 @@ class ListServers extends ListRecords
         ];
     }
 
+    protected function usingGrid(): bool
+    {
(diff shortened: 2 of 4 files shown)


xLaugh:

Boy132:
I don't think we need an extra customization for that. The panel also stores other values in the session and that's fine.

Besides that, this issue is pretty niche because the default lifetime is _2 hours_.

xLaugh:
Fair point on consistency, but I'd frame this less as a timing edge case and more as a basic UX expectation: once someone deliberately changes this setting, they expect it to stick, that's the whole point of changing it.

As a concrete example, I have around 20 servers myself, so I bumped the per-page from 10 to 20 on purpose, to see them all on one page every time I load the dashboard. I don't want to have to re-select that every time my session happens to renew, I want it to just stay set, permanently, like any other preference. It's less about how rare a 2-hour session expiry is and more about the setting not actually behaving like a saved preference at all right now.

lancepioch:
I personally think this idea looks good to me, plus it can always be cached. I agree with the expectation that once a user sets this value, it should remain the same.

FYI, I have NOT run this branch, NOR have I tested it yet though.

lancepioch (approved):
I have pushed some small changes and manually tested and confirmed this is working.

PR #2511, opened 2026-08-12
Add allocation port visibility option

ByCoquito:
Adds a Show port option to server allocations. It is enabled by default.

When disabled, users will only see the allocation IP or alias without the port. The port is still used internally, so server functionality is not affected.


ByCoquito:

lancepioch:
I like this idea and have thought about it for a couple years. I have NOT reviewed your implementation as of right now but am just letting everyone know that I want to push this PR through to get merged after code review and manual testing.

lancepioch:
> [!NOTE]
> **This comment was generated by AI (Claude Code), not written by Lance.** It has not been hand-verified line by line by a human maintainer. Treat it as a starting point for discussion — if anything below is wrong or you disagree with it, say so and a human will weigh in. It is not a formal blocking review.

Thanks for the PR — the feature itself is a good idea. Hiding the port is a common ask for SRV-record and proxied setups (Velocity/BungeeCord and friends), and the scoping is sensible: display-only, default-on, admin views keep the real port, and the migration is clean and reversible so upgrades are a no-op.

Three change requests before this is mergeable.

### 1. Two allocation-release paths bypass the new `updating` hook

The PR establishes the invariant *"an allocation with no `server_id` must have `show_port = true`"* and enforces it in `Allocation::booted()`:

```php
static::updating(function (self $allocation) {
    if (is_null($allocation->server_id)) {
        $allocation->is_locked = false;
        $allocation->show_port = true;
    }
});
```

That hook only fires on model-level saves. Two of the main release paths use query-builder updates, which skip model events entirely:

- `app/Services/Servers/BuildModificationService.php:124` — `$allocations->update(['notes' => null, 'server_id' => null])`
- `app/Services/Servers/ServerDeletionService.php:104` — `$server->allocations()->update(['server_id' => null, 'notes' => null])`

Concrete failure: an admin hides the port on allocation X of server A, then removes X from A via **Edit Server → Networking** (or `PATCH /api/application/servers/{id}/build`). X returns to the free pool with `show_port = 0`. When server B later picks it up, `FindAssignableAllocationService::handle()` does `$allocation->update(['server_id' => $server->id])` — and by the time the hook runs, `$allocation->server_id` is already the *new* non-null id, so the reset doesn't fire there either. Server B's owner now sees `—` where the port should be, with no way to fix it themselves since the toggle is admin-only.

Note the comment sitting directly above the `BuildModificationService` call:

```php
// Remove any of the allocations we got that are currently assigned to this server on
// this node. Also set the notes to null, otherwise when re-allocated to a new server those
// notes will be carried over.
```

That's exactly the bug, reintroduced for `show_port`. `is_locked` has the same gap.

Minimal fix is adding `'show_port' => true, 'is_locked' => false` to both payloads. Cleaner would be a small shared helper on the model so the invariant lives in one place, since there's a third site with the same problem: `ServerTransferController.php:52` and `:84` (`Allocation::query()->whereIn('id', $allocations)->update(['server_id' => null])`) reset neither `notes`, `is_locked`, nor `show_port` on transfer success/failure.

### 2. Please revert the six activity-log call sites back to `address`

These six changed `$allocation->address` → `$allocation->display_address` in **activity log properties**:

- `app/Http/Controllers/Api/Client/Servers/NetworkAllocationController.php:66, 91, 117, 149`
- `app/Filament/Server/Resources/Allocations/AllocationResource.php:99, 122`

Activity records are an audit trail, not a display surface. With `show_port = false` on several allocations sharing one hostname, every resulting `server:allocation.*` entry records the identical string (`play.example.com`), and an admin reading the log can no longer tell which allocation was edited, deleted, or made primary. That's a permanent loss of audit fidelity in exchange for a cosmetic win, and the log isn't shown to the user you're hiding the port from anyway.

`display_address` is right everywhere else you used it — the dashboard entries, `ServerOverview`, the App server list, the allocations table. Just not in the log.

### 3. Scope the bulk-dissociate update

`app/Filament/Admin/Resources/Servers/RelationManagers/AllocationsRelationManager.php:122`:

```php
DissociateBulkAction::make()
    ->after(function () {
        Allocation::whereNull('server_id')->update([
            'notes' => null,
            'is_locked' => false,
            'show_port' => true,
        ]);
```

This is unscoped by node *and* by the selected records, so dissociating one allocation from one server wipes `notes` on **every** unassigned allocation across the entire panel. The unscoped query predates this PR, but since it's being touched here it's a good moment to target the dissociated ids — the single-record `after()` at line 107 already does it correctly.

---

### Two smaller notes (non-blocking)

- **`show_port` isn't honored everywhere.** `app/Filament/Server/Pages/Startup.php` exposes the preview-startup action, which substitutes `{{SERVER_PORT}}` — and most game eggs carry `-p {{SERVER_PORT}}` in the startup command, so the port is one click away on the same panel where it was just hidden. `AllocationTransformer` also still returns `port` to the client API. Both are arguably fine (third-party frontends genuinely need the port), but worth deciding deliberately, and any user-facing description should avoid implying the port is hidden *from the user* — it's a display preference, not concealment.
- **Two small cleanups:** `AllocationResource.php:59` uses a hardcoded `->placeholder('—')` where every sibling placeholder in that file uses `trans(...)`; and `Allocation.php:137` re-implements the existing `getAliasAttribute()` accessor as `$this->ip_alias ?? $this->ip` — using `$this->alias` keeps both branches of `display_address` in sync if alias resolution ever changes.

> [!NOTE]
> Again: **AI-generated review, not a human maintainer's verdict.** Please push back on anything that looks wrong.

lancepioch:
### Admin Toggle
image

### Ports Hidden
image

### Port Hidden
image

### Console w/o Port
image

lancepioch (approved):

PR #2515, opened 2026-08-14
Rename pelican-dev org references to pelican

lancepioch:
**Do not merge this yet.** Holding it as a draft until the org move is finished, since the new paths don't resolve before then. Staging it now so the switch is a merge rather than a scramble.

Most of this is links and image references, but there are a few things in here that are behaviour rather than cosmetics and are worth a look.

`SoftwareVersionService` hits `api.github.com/repos/pelican-dev/{panel,wings}/releases/latest` for the update and version checks, and if that ever stops returning anything the version banners just go blank without anything obviously erroring, which is the annoying kind of failure. Same category of thing: the egg `update_url`s in the test fixtures, the compose files pulling `ghcr.io/pelican-dev/panel:latest`, and the CLA workflow's `remote-organization-name`, which decides where signatures actually get written.

`composer.json` changes the package name to `pelican/panel`, which also shifts `composer.lock`'s content hash, so that's updated too. It's a one-line change in the lock and no dependency versions moved. This one needs the Packagist side renamed to match or the package stops auto-updating.

While I was in here I also fixed the two `ghcr.io/pelican-dev/yolks` references, in the bug report template and in `SettingsControllerTest`. That image never existed under our org, it's a pre-existing typo unrelated to anything else in this PR, and every other yolks reference in this repo already uses `ghcr.io/pelican-eggs/yolks`, so I made those two match. They deliberately do not become `ghcr.io/pelican/yolks`, because that wouldn't exist either.

Two things I deliberately left alone: the Crowdin links in `contributing.md` and `EditProfile.php` still say `crowdin.com/project/pelican-dev`, because Crowdin is a separate service with its own project slug and changing the link before renaming the project over there would just break a link that currently works. If we do rename the Crowdin project we can flip those in a follow-up.

Follow-up work that a PR can't cover: rename the Packagist package, republish the panel image under `ghcr.io/pelican/panel`, move the GitHub Sponsors profile so the FUNDING.yml handle resolves, and fix the repo description and homepage fields in settings.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,2 @@
-github: pelican-dev
+github: pelican
 custom: [https://hub.pelican.dev/donors]
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -56,7 +56,7 @@ body:
   attributes:
     label: Docker Image
     description: The specific Docker image you are using for the game(s) above.
-    placeholder: ghcr.io/pelican-dev/yolks:java_17
+    placeholder: ghcr.io/pelican-eggs/yolks:java_17
 
 - type: textarea
   id: panel-logs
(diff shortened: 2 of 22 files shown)


lancepioch:
@coderabbitai review

notAreYouScared (approved):

PR #2518, opened 2026-08-17
Add server cloning to the admin panel

arthurr0:
Adds a `Clone` action to servers in the admin panel. It opens the regular server creation wizard, pre-filled with the configuration of the selected server, so a copy only needs a new allocation before it can be created.

Requested in #1028.

### What is copied

Egg, docker image, startup command, egg variable values, resource limits, feature limits, container labels, mounts, owner, description and the install script/CPU pinning/swap settings.

### What is not copied

The allocations and the external id, since both have to be unique, and everything that is not part of the server configuration: files, databases, backups, schedules and subusers. A notification points this out when the pre-filled form is opened.

### Implementation

* `ServerCloneService` turns an existing server into the data used to fill the creation form.
* `CreateServer` fills the wizard from `?clone_from={server}`, but only if the user is allowed to view that server. Without the parameter the page behaves exactly like before.
* `CloneServerAction` is shown in the server list and in the header of the server edit page. It requires `view` on the server that is cloned and `create` on servers.

### Tests

* `tests/Integration/Services/Servers/ServerCloneServiceTest.php` covers the copied configuration.
* `tests/Filament/Admin/CloneServerTest.php` covers the pre-filled wizard, creating the clone from it and the action itself. (Note that CI currently only runs `tests/Unit` and `tests/Integration`, so this file is not executed there.)


arthurr0:

PR #2521, opened 2026-08-18
Replace the Fractal transformers with laravel-data wire objects

lancepioch:
Second layer of the API freeze stack, on top of #2520. This replaces the League Fractal transformer pipeline with laravel-data backed wire objects behind an envelope layer we own. The old transformers and Fractal extensions stay in place untouched so the next layer can delete them in one sweep.

### How to review this (65 files, +2444/-210)

The diff splits into three buckets that need very different levels of attention.

**1. The envelope layer, `app/Data/Api/` core (4 files, ~450 lines): read closely**

This is where the risk lives:

- `ApiResource.php` is the laravel-data base class for all wire objects
- `Envelope.php` renders the panel's `{object, attributes}` format with includes nested under `relationships`
- `IncludeContext.php` carries the request and include paths along with the authorize checks ported from the transformers
- `PanelResponse.php` is a drop-in for the injected Fractal wrapper: it reproduces League's include parsing, recursion trimming, the pagination block with its empty `links` object, and Fractalistic's `addMeta` union semantics

**2. Wire objects (30 files): spot-check**

One Data class per existing transformer, 15 application and 15 client. Each is a field-for-field port, and the deliberate transformer quirks are preserved exactly: the egg variable resource key mismatch, the plural `databases` include key, and the deprecated duplicate payload fields. Picking two or three and diffing them against their old transformer is the right sampling strategy; the snapshot suite below covers the rest.

**3. Controller call sites (31 files): mechanical, skim for pattern breaks**

Every call site changes the same way:

```diff
- return $this->fractal->collection($request->user()->sshKeys)
-     ->transformWith($this->getTransformer(UserSSHKeyTransformer::class))
+ return $this->response->collection($request->user()->sshKeys)
+     ->transformWith(UserSSHKeyData::class)
      ->toArray();
```

### Proof

The contract freeze suite from #2520 is the point of the stack: all 75 response snapshots pass with zero changes, and the pre-existing application and client Integration suites stay green.

### Behavior change (one, deliberate)

The egg `config` and `script` includes and the server `transfer` include were declared but never implemented, so requesting them used to 500. They are now silently ignored like any other unknown include.

### DIFF
--- a/app/Data/Api/ApiResource.php
+++ b/app/Data/Api/ApiResource.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace App\Data\Api;
+
+use Carbon\CarbonImmutable;
+use Carbon\CarbonInterface;
+use Spatie\LaravelData\Data;
+
+/**
+ * Base class for the API wire format objects that replaced the Fractal transformers.
+ * Concrete classes are plain property bags hydrated by a static fromModel method,
+ * which the envelope layer invokes through the container so extra typed parameters
+ * receive dependency injection. Includes are declared as closures so the envelope
+ * can resolve them lazily with the request context.
+ */
+abstract class ApiResource extends Data
+{
+    public const RESPONSE_TIMEZONE = 'UTC';
+
+    /**
+     * Include names a request may ask for on this resource. Only names listed here
+     * are honored from the query string; unknown names are silently ignored.
+     *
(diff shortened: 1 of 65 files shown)


notAreYouScared (approved):
yolo?

PR #2523, opened 2026-08-18
Remove Fractal

### LINKED ISSUE #1051: Replace Fractal with Laravel Data
This change ensure that Scramble docs can read the Request & Response (input/outputs), making easier for new developers to onboard over the current API implementations.

Replacing Fractal with Laravel Data comes with the following benefits:
- Improvements over current maintability for long term
- OpenAPI & Typescript Schema's (3rd party integrations)

This change will be backward compatible with Pterodactyl (1.X) for a mid-term period, to let developers adapt to the new ecosystem changes.

___

### Current plan for development:
1. Make our own Data class that extends the original. Add a `setFractal(bool $value = true)` and `bool $isFractal` property to it.
2. Override the generation of the array return to put it into the same shape as a Fractal collection and resources if $isFractal is true.
3. Replace all old array $params and returns with Data classes for the old api.
4. Optionally create a new V2 api that has none of the Fractal shenanigans.

lancepioch:
Final step of the API contract freeze stack (#2520 → #2521 → this). Both predecessors are merged: the wire format is pinned by the fixture suite and every endpoint renders through the laravel-data layer, so nothing uses Fractal anymore. This PR deletes it. **Net −2,424 lines**, and almost all of it is pure removal.

### What's removed

| Removal | Scope |
|---|---|
| All 31 transformer classes | `app/Transformers/**` |
| Custom Fractal wrapper + serializer | `Extensions/Spatie/Fractalistic/Fractal`, `Extensions/League/Fractal/Serializers/PanelSerializer` |
| Fractal config | `config/fractal.php` |
| Composer dependency | `spatie/laravel-fractal` and its `league/fractal` chain |
| Fractal-specific Scramble extensions | `FractalResponseTypeInfer`, `TransformerFactoryTypeInfer`, `TransformerModelBindingExtension` |

### What replaces it

One new class, `app/Extensions/Dedoc/Scramble/PanelResponseTypeInfer.php`, takes over API doc generation from the three deleted Scramble extensions. It follows the `PanelResponse` fluent chain and derives attribute schemas from the Data class properties, covering:

- `MapOutputName` renames
- nullable and `Optional` keys
- enum backing types
- the same envelope, pagination, and `addMeta` shapes as before

### Suggested review order

1. **`PanelResponseTypeInfer.php`**: the only file with new logic (373 lines).
2. **Runtime edits**: `ApplicationApiController`, `ClientApiController`, `BackupController`, `AppServiceProvider`, which just drop the Fractal wiring.
3. **Test helpers**: `assertJsonTransformedWith` and friends now build expected attributes through the Data classes, the same container path the envelope takes in production.
4. Everything else is deletion, safe to skim.

### Verification

- ✅ The #2520 fixture suite passes with **zero snapshot changes**, so the API wire format is byte-for-byte unchanged
- ✅ Full Integration and Unit suites are green
- ✅ `ApiDocumentationTest` passes against the new inference

### Known tradeoff

Array-shaped fields like `limits` now document as generic arrays rather than keyed objects, since the reflection doesn't parse array-shape docblocks. The old extension occasionally did better there; parsing those docblocks is the upgrade path if it matters. Runtime responses are unaffected, this is docs-only.


Boy132:
@coderabbitai full review

Boy132 (approved):

PR #2529, opened 2026-08-18
Add installer preflight and guarded update validation

LegacyAngel2K9:
## Summary

This adds a shared, structured first-boot and lifecycle validation flow for Pelican Panel.

- reuse one environment-check service in the web installer and CLI to validate PHP, required extensions, actual path writability, the encryption key, database connectivity, migrations, cache, administrator state, and queue processing
- verify a real queued job after migrations and before administrator creation, then mark the installation complete only after required first-boot work and run a final health check
- add `p:environment:preflight` and `p:environment:health` for explicit pre-install and post-install/update validation with actionable failure messages
- register the queue worker check in the existing health scheduler
- add guarded `p:maintenance:prepare-update` and `p:maintenance:finish-update` commands that check the target release's locked Composer platform requirements, capture a protected pre-update state snapshot, provide database-backup guidance, and point to exact rollback instructions when post-update validation fails

The update preparation command does not replace application files or run migrations. Non-SQLite databases receive backup guidance without exposing credentials; SQLite is copied into the private snapshot. Environment snapshots are stored below `storage/app/private`, with the `.env` copy restricted to mode `0600`.

## Validation

- `vendor/bin/pint` on every changed PHP file
- `vendor/bin/phpstan analyse --no-progress`
- 14 focused tests / 47 assertions covering installer ordering, queue failure behavior, CLI results, complete health checks, target compatibility, and update snapshots
- full local suite: 471 passed, 3 skipped; 4 pre-existing Windows-only failures in `NormalizeEggCommandTest` because its temporary path loses backslashes (`C:\Users\...` becomes `C:Users...`). No Egg normalizer files are changed in this PR.

## CLI examples

```bash
php artisan p:environment:preflight --with-database --with-queue
php artisan p:environment:health
php artisan p:maintenance:prepare-update --source=/path/to/extracted/release --target-version=v1.x.y
php artisan p:maintenance:finish-update
```


app/ValueObjects/UpdateSnapshot.php
lancepioch, line 1: Instead of `ValueObjects` can you please simply do `Data` and extend and utilize spatie/laravel-data that we already have required?

app/Services/Environment/InstallationHealthService.php
lancepioch, line 1: I'd say we should definitely consolidate this with the Spatie Health checks that we already use

app/Services/Environment/QueueWorkerProbeService.php
```diff
@@ -0,0 +1,77 @@
...
+
+                    return new EnvironmentCheckResult(
+                        'queue',
+                        trans('installer.health.queue.label'),
+                        EnvironmentCheckStatus::Passed,
+                        trans('installer.health.queue.passed', ['connection' => $connection]),
+                    );
+                }
+
+                usleep(100_000);
```
lancepioch, line 44: I'm not a huge fan of sleeps and if sleeps are ever required, then we should definitely use the Laravel Sleep library

app/Services/Environment/QueueWorkerProbeService.php
lancepioch, line 1: My guess is that this might cause issues, so we might need some sort of UI to interface with this for retries instead of only hard failing... not entirely sure yet.

app/Livewire/Installer/PanelInstaller.php
```diff
@@ -155,12 +166,48 @@ public function submit(UserCreationService $userCreationService): void
             // Install selected eggs
             $this->installEggs();
 
+            // Disable installer only after every required first-boot action completed.
+            $this->writeToEnvironment(['APP_INSTALLED' => 'true']);
```
lancepioch, line 170: We put it first originally because if any of the previous steps fail, then we're not sure if that would block the installation from happening; ex. the admin user already existing or the install eggs call not being idempotent

I think a test to satisfy if you call those multiple times it doesn't crash would suffice.

app/Services/Environment/InstallationHealthService.php
```diff
@@ -0,0 +1,343 @@
...
+        'gd',
+        'intl',
+        'json',
+        'mbstring',
+        'pdo',
+        'xml',
+        'zip',
+    ];
+
+    public const DATABASE_EXTENSIONS = [
```
lancepioch, line 32: Should this be an enum somewhere or is it already?

app/Services/Environment/InstallationHealthService.php
```diff
@@ -0,0 +1,343 @@
...
+use Illuminate\Database\Migrations\Migrator;
+use Illuminate\Encryption\Encrypter;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+use Throwable;
+
+class InstallationHealthService
+{
+    public const MIN_PHP_VERSION = '8.3.0';
```
lancepioch, line 18: correct but might live somewhere better

app/Console/Commands/Environment/AppSettingsCommand.php
```diff
@@ -2,16 +2,29 @@
...
 
     protected $signature = 'p:environment:setup';
 
-    public function handle(): void
+    public function handle(InstallationHealthService $health): int
     {
+        $results = $health->systemRequirements();
+        $this->displayEnvironmentChecks($results);
+
+        if ($health->hasFailures($results)) {
```
lancepioch, line 22: I think this is correct behaviour to block the install, but also we should probably add a flag (--) to bypass or skip this just in case.

app/Services/Maintenance/UpdateCompatibilityService.php
```diff
@@ -0,0 +1,55 @@
...
+        $composerLock = $source . DIRECTORY_SEPARATOR . 'composer.lock';
+
+        if (!is_file($composerJson) || !is_file($composerLock)) {
+            return $this->failure(trans('commands.update.compatibility_files_missing'));
+        }
+
+        try {
+            $result = Process::path($source)
+                ->timeout(120)
+                ->run(['composer', 'check-platform-reqs', '--lock', '--no-dev']);
```
lancepioch, line 24: I don't believe we can just simply assume that `composer` lives in the PATH. I would love to, but I don't think it's reasonable.

app/Services/Maintenance/UpdateSnapshotService.php
lancepioch, line 1: Need some sort of built in pruning because these snapshots will keep growing and also contain .env and database.sqlite files too.

app/Livewire/Installer/PanelInstaller.php
```diff
@@ -106,9 +115,9 @@ protected function getFormSchema(): array
     protected function getDefaultSteps(): array
     {
         return [
-            RequirementsStep::make(),
+            RequirementsStep::make($this->installationHealth),
             EnvironmentStep::make($this),
-            DatabaseStep::make($this),
+            DatabaseStep::make($this, $this->installationHealth),
```
lancepioch, line 120: I believe the service needs to be resolved (DI) inside the Step class instead. If it's a livewire component then it'd be the boot method.

tests/Filament/Installer/PanelInstallerTest.php
lancepioch, line 1: There is a lot of mocking and this is a phpunit test. New tests should be in pestphp!

lancepioch (changes requested):
Can we please split this into separate pull requests so we can merge them easier?

(a) installer requirements refactor + shared checks,
(b) the QueueCheck + scheduler registration (two lines, merge-ready today),
(c) the update snapshot/rollback tooling

LegacyAngel2K9:
Closing this umbrella PR now that the requested split is complete:

- #2531 — queue health monitoring and scheduler integration (merged)
- #2533 — installer preflight and post-install health checks
- #2545 — guarded update snapshots, compatibility validation, retention, and rollback guidance

The focused PRs preserve the original work while making each concern independently reviewable. Thank you for the direction on splitting it up.

PR #2534, opened 2026-08-19
Fix/wings server list latency

Mesharsky:
## Summary

- Use the configured connect and request timeouts for Wings requests.
- Load admin server conditions after the database-backed rows render.
- Defer the user server table until after the initial page render.
- Keep grid placeholders from requesting Wings resources.

## Notes

Admin condition checks load independently for visible rows. The user table still resolves its server data in a deferred Livewire request.

### DIFF
--- a/app/Filament/Admin/Resources/Servers/Pages/ListServers.php
+++ b/app/Filament/Admin/Resources/Servers/Pages/ListServers.php
@@ -16,6 +16,7 @@ use Filament\Actions\EditAction;
 use Filament\Resources\Pages\ListRecords;
 use Filament\Tables\Columns\SelectColumn;
 use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Columns\ViewColumn;
 use Filament\Tables\Grouping\Group;
 use Filament\Tables\Table;
 use Illuminate\Database\Eloquent\Builder;
@@ -37,12 +38,9 @@ class ListServers extends ListRecords
                 Group::make('egg.name')->getDescriptionFromRecordUsing(fn (Server $server) => str($server->egg->description)->limit(150)),
             ])
             ->columns([
-                TextColumn::make('condition')
+                ViewColumn::make('condition')
                     ->label(trans('admin/server.condition'))
-                    ->default('unknown')
-                    ->badge()
-                    ->icon(fn (Server $server) => $server->condition->getIcon())
-                    ->color(fn (Server $server) => $server->condition->getColor()),
+                    ->view('livewire.columns.server-condition-column'),
                 TextColumn::make('uuid')
                     ->hidden()
                     ->label(trans('admin/server.uuid'))
--- a/app/Filament/App/Resources/Servers/Pages/ListServers.php
+++ b/app/Filament/App/Resources/Servers/Pages/ListServers.php
@@ -136,6 +136,7 @@ class ListServers extends ListRecords
         $defaultPerPage = in_array($storedPerPage, $pageOptions, true) ? $storedPerPage : ($usingGrid ? 10 : 20);
 
         return $table
+            ->deferLoading(!$usingGrid)
             ->paginated($pageOptions)
             ->defaultPaginationPageOption($defaultPerPage)
             ->query(fn () => $baseQuery)
--- a/app/Livewire/ServerCondition.php
+++ b/app/Livewire/ServerCondition.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Livewire;
+
+use App\Models\Server;
+use Illuminate\View\View;
+use Livewire\Attributes\Locked;
+use Livewire\Component;
+
+class ServerCondition extends Component
+{
+    #[Locked]
+    public Server $server;
+
+    public function render(): View
+    {
+        return view('livewire.server-condition', [
+            'condition' => $this->server->condition,
+        ]);
+    }
+
+    public function placeholder(): View
+    {
+        return view('livewire.server-condition-placeholder');
+    }
+}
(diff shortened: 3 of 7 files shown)


Mesharsky:

app/Models/Node.php
```diff
@@ -408,8 +408,8 @@ public function statistics(): array
             try {
 
                 $data = Http::daemon($this)
-                    ->connectTimeout(1)
-                    ->timeout(1)
+                    ->connectTimeout(config('panel.guzzle.connect_timeout', 5))
+                    ->timeout(config('panel.guzzle.timeout', 15))
```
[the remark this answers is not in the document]
Mesharsky: This call is already outside the initial page render. In table mode, deferLoading(!$usingGrid) is enabled, so Filament first renders the unloaded shell and invokes loadTable through wire:init. Line 103 only runs in that follow-up request. In grid mode, tableColumns() is not used and each server entry is a lazy Livewire component with a Wings-free placeholder.

A cold statistics cache can still delay the deferred table request once per distinct node. That is a remaining time-to-data limitation, not an initial-render blocker. A dashboard-specific shorter timeout would reintroduce failures for high-latency nodes and would not address the separate per-server status and resource calls.

Multiple tests were ran before the commit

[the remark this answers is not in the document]
Mesharsky: What could be added in the future would be a pooled, deduplicated preload for the visible table page.
But still that PR fixes the main problem

Mesharsky:
> [Review Change Stack](https://app.coderabbit.ai/change-stack/pelican/panel/pull/2534?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)
> 📝 Walkthrough
> 
> 	
> 
> 	
> 
> 	
> 
> 	
> 
> 	
> 
> 	
> 
> **Merge Risk:** _🟡 Moderate_ · up to `0e395`
> 
> The server list can still block page rendering for up to 15 seconds per unavailable node, with multiple nodes potentially causing substantially longer loads. Merge readiness requires a bounded dashboard timeout or deferred/asynchronous condition check, or explicit owner acceptance.
> 🚥 Pre-merge checks | ✅ 5
> 		
> 		
> 		
> 		
> 		
> 		
> 
> Thanks for using [CodeRabbit](https://coderabbit.ai?utm_source=oss&utm_medium=github&utm_campaign=pelican/panel&utm_content=2534)! It's free for OSS, and your support helps us grow. If you like it, consider giving us a shout-out.
> ❤️ Share
> 
> 
> Comment `@coderabbitai help` to get the list of available commands.

This call is already outside the initial page render. In table mode, deferLoading(!$usingGrid) is enabled, so Filament first renders the unloaded shell and invokes loadTable through wire:init. Line 103 only runs in that follow-up request. In grid mode, tableColumns() is not used and each server entry is a lazy Livewire component with a Wings-free placeholder.

A cold statistics cache can still delay the deferred table request once per distinct node. That is a remaining time-to-data limitation, not an initial-render blocker. A dashboard-specific shorter timeout would reintroduce failures for high-latency nodes and would not address the separate per-server status and resource calls.

app/Models/Node.php
lancepioch: Remove the defaults from here.

```suggestion
                    ->connectTimeout(config('panel.guzzle.connect_timeout'))
                    ->timeout(config('panel.guzzle.timeout'))
```

app/Models/Node.php
```diff
@@ -430,7 +430,11 @@ public function ipAddresses(): array
             $ips = collect();
 
             try {
-                $addresses = Http::daemon($this)->connectTimeout(1)->timeout(1)->get('/api/system/ips')->json();
+                $addresses = Http::daemon($this)
+                    ->connectTimeout(config('panel.guzzle.connect_timeout', 5))
+                    ->timeout(config('panel.guzzle.timeout', 15))
```
lancepioch: Remove the defaults.

```suggestion
                    ->connectTimeout(config('panel.guzzle.connect_timeout'))
                    ->timeout(config('panel.guzzle.timeout'))
```

app/Models/Node.php
lancepioch, line 1: These timeouts are purposefully much smaller so that when a Node is unavailable the page doesn't hang for 5-15 seconds. Does the rest of the PR changes fix that? I'm only fine with these changing if it doesn't cause the page to hang.

lancepioch (changes requested):

app/Models/Node.php
Mesharsky: > These timeouts are purposefully much smaller so that when a Node is unavailable the page doesn't hang for 5-15 seconds. Does the rest of the PR changes fix that? I'm only fine with these changing if it doesn't cause the page to hang.

Yes. The rest of the PR prevents these requests from blocking the page render. Admin conditions and grid entries load through lazy Livewire components, while table mode uses Filament’s deferred loading, so the page shell renders immediately and remains usable while the table data loads.  
Further isolating or pooling the table’s Wings-dependent work would improve time-to-data, but that is a separate optimization and not a page-hang issue.

lancepioch:
@Mesharsky I just tested this on your branch with a fresh node and servers:

```php
// php artisan tinker
$node = Node::factory()->create(['fqdn' => '10.255.255.1']); // unroutable IP, every call eats the full connect timeout
Server::factory()->count(10)->withNode($node)->create(['owner_id' => 1, 'egg_id' => Egg::first()->id]);
```

This causes / (root) to take over 100 seconds to load in your branch.

I'll need this addressed or that part reverted for this PR, thanks!

Mesharsky:
> @Mesharsky I just tested this on your branch with a fresh node and servers:
> 
> ```
> // php artisan tinker
> $node = Node::factory()->create(['fqdn' => '10.255.255.1']); // unroutable IP, every call eats the full connect timeout
> Server::factory()->count(10)->withNode($node)->create(['owner_id' => 1, 'egg_id' => Egg::first()->id]);
> ```
> 
> This causes / (root) to take over 100 seconds to load in your branch.
> 
> I'll need this addressed or that part reverted for this PR, thanks!

I removed the configured timeout changes

The remaining changes only defer Wings-dependent rendering: admin conditions load lazily, list mode defers table records, and grid placeholders no longer perform a Wings resource request. Please check now

PR #2539, opened 2026-08-23
improve file-upload

wan0v:
Hey everyone,
a few days ago I installed Pelican and had to migrate my gameserver savefiles. Here are my suggestions to improve the file-upload experience a bit.

You now can:
- see a progress bar for each file
- abort an upload of a file
- minimize the file upload window and navigate to other pages

[file-upload.webm](https://github.com/user-attachments/assets/ac014fff-58b8-44a6-82f5-dec38a72a1ea)


wan0v:

lancepioch:
Thank you for putting this together. The per-file progress, cancellation, and ability to continue uploads while navigating are all genuinely useful improvements, and I’d like to find a path forward for them.

Before we can merge this, though, we’ll need to revisit the frontend structure. Right now, much of the upload state and orchestration lives directly inside a single 600+ line Blade view, including roughly 400 lines of JavaScript. That would make it difficult for us to test, maintain, and safely extend over time.

This is a concern with the implementation structure, not the feature itself. Before you spend more time refactoring it, please join us in Discord so we can discuss how best to separate the responsibilities and agree on a maintainable direction.

PR #2545, opened 2026-08-25
Add guarded update snapshots and rollback guidance

LegacyAngel2K9:
## Summary

- verify an extracted target release against its locked Composer platform requirements before any files change
- capture a protected pre-update snapshot containing `.env`, Composer state, metadata, and a WAL-safe SQLite backup or driver-specific database guidance
- prune old secret-bearing snapshots with a configurable retention limit
- run the existing Spatie health checks after updating and provide concrete rollback guidance on failure

## Context

This is the final focused PR split from #2529, covering the update snapshot and rollback tooling requested during review. It is independent of the installer preflight PR.

## Safety details

- Composer is resolved from `--composer`, a local `composer.phar`, or PATH; PATH availability is not assumed.
- Named database connections are resolved to their actual driver.
- SQLite snapshots use `VACUUM INTO`, including committed WAL data.
- Required writes, permissions, metadata, and rollback artifacts are validated before a snapshot is exposed.
- Snapshots are private (`0700` directories and `0600` files) and retention defaults to three.

## Validation

- `vendor/bin/pest tests/Integration/Maintenance/UpdateSafetyTest.php` — 6 passed, 19 assertions
- Pint — passed
- PHPStan — no errors


config/panel.php
```diff
@@ -87,4 +87,9 @@
         'dev_mode' => env('PANEL_PLUGIN_DEV_MODE', false),
         'max_import_size' => env('PANEL_PLUGIN_MAX_IMPORT_SIZE', 1024 * 1024 * 100),
     ],
+
+    'updates' => [
+        // Snapshots contain .env and possibly SQLite data, so retain only a small recovery window.
+        'retained_snapshots' => env('PANEL_UPDATE_RETAINED_SNAPSHOTS', 3),
```
lancepioch: Instead of a hard number, instead I think we should have a number of days retention and they are auto deleted after that.

PR #2553, opened 2026-09-02
Add user account suspension controls

ByCoquito:
This PR adds user account suspension controls.

Administrators can suspend a user while keeping their servers running, or suspend both the account and its owned servers. Suspended users are signed out and blocked from accessing the panel, API, OAuth, passkeys, WebSocket, and SFTP.

When lifting a suspension, administrators can optionally unsuspend only the servers affected by that suspension.


app/Data/Api/Application/UserData.php
```diff
@@ -26,6 +26,8 @@ public function __construct(
         public string $email,
         public ?string $language,
         public bool $root_admin,
+        public bool $suspended,
+        public ?string $suspended_at,
```
lancepioch: Looks like you only need the suspended_at date because it's nullable, if it's null then the user is not suspended

lancepioch:
Hi, thank you for your pull request. Please see the following for my requests:

* [x] Remove any changes to api outputs, they are currently frozen in preparation for 1.0 release (a follow-up pr to 1.0 is welcomed to add them back if necessary)
* [x] Remove the test fixtures accordingly, they are only there to help with the Fractal removal
* [x] Attach screenshots or screencast at minimum to demonstrate this working
* [x] The suspension messages/strings need to be translatable
* [x] Is it programmatically possible to not require the new column `auth_session_version`?
* [x] Because `suspended_at` is a nullable date, we don't need `suspended` as a boolean.
* [x] Let's remove `user_suspensions` and `user_suspension_servers` and keep this as a single column/toggle on the User model itself

Honestly at this point I'm not sure if this should live in the core or be a plugin instead.

ByCoquito:
Updated based on the review feedback.

The implementation now uses only the nullable `suspended_at` column, without additional suspension tables, a boolean flag, or `auth_session_version`. API output and fixture changes were removed, all suspension messages are translatable, and active sessions are invalidated using the existing remember token.

Account suspension no longer changes the state of the user's servers. Screenshots demonstrating the updated flow are attached below.

ByCoquito:
image
image
image
image

ByCoquito:
Fixed in `cd45e7315`.

Authenticated sessions that do not contain `auth_remember_token_hash` are now invalidated instead of silently adopting the new marker. This safely expires legacy sessions containing only `auth_session_version`, including sessions using non-database drivers.

A regression test using the `array` session driver was added.

redthirten:
Hey @ByCoquito, if you don't mind, could you help me understand a scenario where this addition might be useful? I'm sure I'm missing something, but I don't easily see a use-case where you would only want to suspend a user and not their servers. 

Only thing I can really think of at the moment is locking down a compromised account, but if you're going in to manually suspend an account due to this, you might as well just reset their password. For other issues, like billing or abuse, I don't see why you wouldn't want the servers suspended as well.

ByCoquito:
Hi @redthirten, Sure! The main idea is to separate account access from the operational state of its servers.

For example, if an account is compromised, under investigation, or involved in a temporary ownership dispute, an administrator may want to immediately prevent the user from accessing it without causing downtime for the players or customers using their servers. It also allows the service to remain available while the administration team investigates the situation or transfers the servers to another owner.
A password reset may not be sufficient because a suspension represents an explicit administrative restriction and prevents any new access until an administrator removes it.
I agree that for abuse or billing issues, suspending the servers would usually make sense as well. The purpose of this option is to let administrators decide based on the situation instead of forcing both actions to always be tied together.

Let me known if you have another question, greatings!

lancepioch:
Hey @ByCoquito, I greatly appreciate the work you put into this and how quickly you addressed all of the review feedback. After discussing it with the rest of the Pelican team, we don't think account suspension separate from server suspension fits in core: our main audience is self-hosters, and the folks who need this are generally better served by dedicated billing software. That said, we'd genuinely welcome this as a plugin.

Thank you again for the contribution, closing this for now.

PR #2562, opened 2026-09-09
Add an api_version field to plugin.json

lancepioch:
Part of the plugin API contract work.

Plugins can declare which plugin api version they target so the panel can refuse to load plugins written against a future incompatible api instead of failing unpredictably. The field is optional and defaults to 1. Plugins declaring a higher version than the panel supports are marked incompatible before loading using the same status flow as the panel_version check, plugins missing the field get a warning note under their status badge on the plugins page, and `p:plugin:make` scaffolds the field.

### Screenshot

plugins-api-version

### DIFF
--- a/app/Services/Helpers/PluginService.php
+++ b/app/Services/Helpers/PluginService.php
@@ -42,6 +42,13 @@ class PluginService
         $plugins = Plugin::orderBy('load_order')->get();
         foreach ($plugins as $plugin) {
             try {
+                // Filter out plugins that require a newer plugin api than this panel supports
+                if (!$plugin->isApiVersionSupported()) {
+                    $this->setStatus($plugin, PluginStatus::Incompatible, 'This Plugin requires plugin api version ' . $plugin->effectiveApiVersion() . ' but this Panel only supports up to version ' . Plugin::SUPPORTED_API_VERSION . '!');
+
+                    continue;
+                }
+
                 // Filter out plugins that are not compatible with the current panel version
                 if (!$plugin->isCompatible()) {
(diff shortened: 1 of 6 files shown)


Boy132:
Why do we need both a `panel_version` and a `api_version`? A specific panel  version contraint also implies a specific api version constraint.

lancepioch:
> Why do we need both a `panel_version` and a `api_version`? A specific panel version contraint also implies a specific api version constraint.

`panel_version` only works in one direction. It can say "this plugin needs panel 1.6 or newer". It cannot say "this plugin stops working after panel X", because the plugin author does not know which future panel version will break the plugin API.

Example: a plugin says panel_version: `^1.2`. Later, panel 1.9 changes the plugin API in a breaking way. The old check still passes (1.9 is newer than 1.2), so the panel loads the plugin and it crashes. With `api_version`, the panel sees the plugin was built for API 1, knows it now uses API 2, and can refuse to load it with a clear message instead.

Also, panel versions change all the time (features, bug fixes). The API version only changes when the plugin API actually breaks. Paper/Bukkit does the same thing: plugins declare api-version separately from the server version.

Boy132 (approved):

app/Services/Helpers/PluginService.php
```diff
@@ -42,6 +42,13 @@ public function loadPlugins(): void
         $plugins = Plugin::orderBy('load_order')->get();
         foreach ($plugins as $plugin) {
             try {
+                // Filter out plugins that require a newer plugin api than this panel supports
+                if (!$plugin->isApiVersionSupported()) {
```
Boy132, line 46: Thinking about it, this check should only be executed in production env. This way plugin developers in a local env can work on their plugin/fix it for the current api version.

Boy132 (changes requested):

Boy132 (approved):

PR #2563, opened 2026-09-09
Raise the support floor to PHP 8.4

lancepioch:
## Summary

- Raises the CI matrix to the 1.0 PHP floor: the SQLite and lint matrices drop PHP 8.3, and the mysql legs run 8.4 and 8.5 so the minimum PHP is exercised against a real engine.
- `composer.json` now requires `^8.4 || ^8.5` with `platform.php` 8.4, lock hash regenerated.
- Registers the `Feature` testsuite in phpunit.xml so `vendor/bin/pest --testsuite Feature` works locally the same way the CI step from #2583 runs it.
- The MariaDB matrix is unchanged (10.11 and 11.4). Nothing in the panel needs 11.4, and the MariaDB floor is a W7 decision.

## Test plan

- `vendor/bin/pest tests/Feature --parallel` passes on the merged branch (40 tests).
- CI on this PR shows the reduced PHP matrix.


Boy132:
Is there a specific reason to drop 8.3 support?

lancepioch:
@Boy132 

1. PHP 8.3 left active support in Dec 2025 and is security-only until Dec 2027. If 1.0 ships late 2026 with a 1.0.x support window, we'd be committed to supporting an EOL PHP for part of that window. 8.4 is supported through Dec 2028.

2. Semver: Once 1.0 lands with the stability commitment, the floor can only be raised at a major. The 1.0 boundary is the one free chance to drop 8.3; keeping it means carrying it until 2.0.

3. Practically, one less CI leg per matrix, dependencies increasingly require 8.4 (Filament/Laravel ecosystem is moving there quickly), and 8.4 features (property hooks, asymmetric visibility) become usable in the codebase.

.github/workflows/ci.yaml
```diff
@@ -159,7 +159,7 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.5]
-        database: ["mariadb:10.11", "mariadb:11.4"]
+        database: ["mariadb:11.4"]
```
rmartinoscar: What is the use of a matrix there and why do we drop mariadb 10 ?

rmartinoscar (approved):
Other then my question LGTM

.github/workflows/ci.yaml
lancepioch: Fair on both. Nothing in the panel needs 11.4: Laravel's MariaDB grammar only gates on 10.5.2 and 10.7.0, our compose-full-stack.yml ships 10.11, and it's the Debian 12 and Ubuntu 24.04 default, so it's what most installs run. The drop was following the W7 plan number, and that floor hasn't been decided yet, unlike PHP 8.3 where the composer constraint makes it a semver question for 1.0. Put 10.11 back in 8d13b202a, which also gives the matrix its second entry again. The MariaDB floor stays with W7 alongside the other engine numbers.

PR #2589, opened 2026-09-18
fix: add retry to daemon system info request to handle transient connection failures

KotyaraKryt:
## Problem
The panel occasionally shows "Error connecting to node!" for a node that is actually online and responding normally. This happens even when:
- Manual `curl` requests to the node's Wings API succeed instantly (<50ms)
- The node's DNS resolves consistently
- Wings is confirmed listening on the configured port
- 6 parallel manual curl requests to the same node all succeed without issue

The failure surfaces as a fast `ConnectionException` (`Couldn't connect to server ... after ~30ms`), not a timeout — suggesting a transient TCP-level issue (e.g. an occasional refused connection) rather than a systemic network or configuration problem.

## Fix
Add `->retry(2, 200)` to the `getSystemInformation()` request in `DaemonSystemRepository`, so a single transient connection failure doesn't immediately surface as a node-down error in the UI.

## Testing
Reproduced the issue on a live 2-node setup connected via a low-latency private network. After applying this patch, the node list no longer intermittently shows nodes as unreachable under normal operating conditions.

## Checklist

- [x] This pull request does one thing (unrelated changes are split into separate pull requests)
- [x] New translation strings are added to English only (other languages come from [Crowdin](https://crowdin.com/project/pelican-dev))
- [x] UI changes include before/after screenshots


KotyaraKryt:

lancepioch:
Please create an issue first, thank you.

